PR #3, opened 2026-06-03
Improve web chat alignment and models browser performance

VernSG:
## Summary

- Virtualize the models.dev browser list so large catalogs render smoothly.
- Defer model search filtering and avoid re-sorting the catalog on every filter change.
- Right-align user chat messages while keeping assistant and tool messages left-aligned.

## Details

The models.dev browser was rendering every filtered model row at once. This keeps the existing fetch behavior, but only renders the visible rows plus overscan, which reduces DOM work for large catalogs.

The chat UI also had a global alignment override that forced every message role to the left. Message alignment is now role-aware: user messages align right, while assistant and tool messages remain on the left.

## Testing

- `bun run --filter @tinyclaw/web build`
- Verified the models.dev picker loads and search updates correctly.
- Verified existing chat history shows user messages on the right and assistant messages on the left.

### DIFF
--- a/apps/web/src/components/ModelsBrowseList.tsx
+++ b/apps/web/src/components/ModelsBrowseList.tsx
@@ -1,4 +1,11 @@
-import { useMemo, useState } from "react";
+import {
+  useDeferredValue,
+  useEffect,
+  useLayoutEffect,
+  useMemo,
+  useRef,
+  useState,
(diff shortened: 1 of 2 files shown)


ahmadrosid (approved):
nice thanks

Issue #101, opened 2026-07-14
Bug: File preview and creation pipeline broken 


fajarhide:
# Summary

**Four related bugs in artifact handling and preview:** 
1. text files written directly via write_file cannot be previewed because they lack a .nakama-meta.json sidecar and are served as application/octet-stream
2. write_file fakes .docx files by writing HTML/XML and renaming them, producing broken Word documents
3. convertHtmlToMarkdown crashes on HTML tables because remark-gfm is missing from the pipeline
4. the artifact preview panel re-fetches content and loses the Copy/Download buttons when the same file is re-clicked due to config being overwritten instead of merged.

# Why

1. Text preview blocked: The frontend relies solely on sidecar metadata for MIME type. Files written directly via write_file have no sidecar, so the server defaults to application/octet-stream and the frontend rejects preview. Extension-based inference is never attempted.
2. Fake .docx files: write_file is text-only but does not reject .doc/.docx extensions. The agent works around this by writing HTML/XML with a .docx name, producing files that Word opens as raw text. Meanwhile, read_file and knowledge base cannot read real .docx (ZIP) files because they force UTF-8 decoding.
3. HTML table crash: The rehype-parse → rehype-remark → remark-stringify pipeline lacks remark-gfm, so table nodes are unknown and throw. Word conditional comments also leak through because HTML comments are not stripped.
4. Preview panel state loss: show() in chat-attachment-panel-context.tsx replaces the entire panel config instead of merging. openPanel calls show() without headerActions, so Copy/Download are added later via update(). On re-click, show() overwrites the config and drops the buttons, but update() does not re-run because dependencies are unchanged. The fetch effect also resets content to null on every open, forcing unnecessary re-requests.

PR #102, opened 2026-07-14
Improve document preview and generation based on content type

fajarhide:
## Problem

All four bugs share one root cause: files are judged by their name, not their content. This leads to text files being treated as binary, .docx files being faked as HTML, HTML tables crashing the converter, and preview panel state being rebuilt from scratch instead of cached.

## Fix Issue #101 

https://github.com/user-attachments/assets/431ce9dd-bc3f-4b64-a20b-3df2e7cf8160

## Changes

**Mime resolution** — one shared module (`artifact-mime.ts`) used by both server and web. Falls back to the extension when the sidecar is missing, so `.md` is served as `text/markdown`.

**Rendering matches the format** — Markdown renders through the same Streamdown pipeline as chat; code/data files (`json`, `yaml`, `ts`, `py`, `sql`, …) get syntax highlighting; unknown extensions are content-sniffed for UTF-8 rather than assumed binary. Real binaries (PNG, ZIP) are still correctly refused.

**Word documents are decided by their bytes** — `PK` magic → real `.docx` (mammoth); `D0CF11E0` → genuine legacy OLE `.doc`, refused with an actionable message; otherwise HTML/XML/text → converted. This makes the already-broken files above readable instead of dumping CSS at the reader. Applies to `read_file`, the knowledge base, and the artifact preview.

**New `write_docx` tool** — builds a real OOXML archive from Markdown (headings, bold/italic, lists, tables, code). `write_file` and `edit_file` now refuse `.doc`/`.docx` paths and point at it, so the fake cannot recur. Downloads stay byte-exact; only previews convert.

**Knowledge base accepts `.docx`** — previously PDF was supported but Word was not.

## Pre-existing bugs found while converting the real reports

- `convertHtmlToMarkdown` **threw on any `table` node** (`Cannot handle unknown node 'table'`). This broke **`web_fetch` on any page containing a table**, not just documents. Fixed with `remark-gfm`.
- Word conditional comments (`<!--[if gte mso 9]>`) leaked through as visible text.

## Preview caching

Reopening an already-loaded preview refetched every time and re-showed the spinner. Content is now fetched once per artifact and reopens instantly.

## Verification

Tested against the actual broken files, not mocks: all three now convert to clean Markdown with correct headings and **zero CSS leakage**, while downloads remain byte-identical. `write_docx` output carries the `PK` ZIP magic, is identified by `file(1)` as *"Microsoft Word 2007+"*, and round-trips back through the reader with headings, tables, and bold intact.

Suites: core 417, server 453, web 89, agent 44, client 11 — all passing. Web typecheck clean.

**Not verified:** the generated `.docx` has not been opened in Microsoft Word itself, and the preview UI was not driven in a browser.


ahmadrosid:
Looks great — I like this approach of judging files by content rather than the extension alone. The shared MIME helpers, `write_docx`, and the opt-in `render=markdown` path are a solid fix.

Thank you @fajarhide for the careful work and the thorough verification!

Issue #58, opened 2026-06-18
Implement strict role-based access control for all actions


ahmadrosid:
Tinyclaw needs strict RBAC so teams can safely use the platform beyond just owners and C-
  level users. The goal is to give staff limited, intentional access without exposing
  dangerous actions.

  Today, if access is too broad, it becomes unsafe for companies to adopt the platform across
  their internal teams.

  Why this matters:

  - Enables safe team adoption
  - Prevents accidental or unauthorized destructive actions
  - Makes the product suitable for real organizations with different user roles

  Acceptance Criteria:

  - Define clear roles such as Owner, Admin, Manager, Member, and Viewer
  - Restrict actions based on role permissions
  - Enforce permissions on both frontend and backend
  - Prevent staff from accessing sensitive admin-only functions
  - Ensure dangerous actions require explicit permission checks
  - Add auditability for permission-sensitive actions if possible

  Notes:

  - RBAC should be strict by default, not permissive
  - Avoid relying only on UI hiding for security
  - Permissions should be enforced server-side

PR #108, opened 2026-07-16
fix: enforce RBAC on automation & task mutations and profile binding (#107)

### LINKED ISSUE #107: Security: viewer role can create and run automations (authorization not enforced on state-changing routes)
## Summary

Several state-changing API routes enforce **organization context** but not **organization role**. As a result, a user with the lowest-privilege `viewer` role can create and execute automations. An automation runs an agent, so this turns a "read-only" viewer into someone who can perform destructive and outbound actions (delete/write files, send email, fetch arbitrary URLs) via the default profile's tools.

This relates directly to #58 (strict RBAC). The role primitives already exist, they are simply not applied on these routes.

## Impact (confirmed)

A `viewer` can create and run an automation bound to the org's **default** profile. That profile carries the built-in tools: `write_file`, `write_docx`, `delete_file`, `edit_file`, `read_file`, `search_files`, `knowledge_base_search`, `web_search`, `web_fetch`, `email`. So a role intended as read-only can:

- delete / overwrite / edit files in the profile workspace,
- send email (`email` tool),
- exfiltrate data to an attacker-controlled URL (`web_fetch`),
- incur unbounded LLM cost by scheduling/running automations.

This is enforced server-side only in intent, the API accepts the requests directly, so hiding buttons in the frontend does not mitigate it.

## Root cause

The role guards already exist in `apps/server/src/http/org-guards.ts`:

- `requireOrgAdmin`, `requireNotViewer`, `requirePlatformAdmin`

But some route files only call `requireActiveOrgIdFromContext`, which validates that an active org is present and returns its id, it performs **no role check**. The org-context middleware (`org-middleware.ts`) merely *populates* `orgRole`; it does not enforce it.

Guard coverage audit (mutating routes vs. role-guard calls per file):

| File | Mutating routes | Role guards |
|---|---|---|
| `routes/automations.ts` | 7 | **0** |
| `routes/tasks.ts` | 5 | **0** |
| `routes/user-context.ts` | 2 | **0** |

## Evidence

Verified locally against the in-memory adapter using the existing test harness (`setupFreshInstallSession(app, db, email, "viewer")`), no source changes:

**Viewer creates an automation**
```
POST /v1/automations   (session role = viewer)
-> HTTP 201
-> automationService.create called: 1 time(s)
```

**Viewer triggers an automation run**
```
POST /v1/automations/{id}/run   (session role = viewer)
-> HTTP 200
-> agent.runAutomation invoked: 1 time(s)
```

Both should be `403 Forbidden` for a viewer. A regression test asserting `403` fails on `main` today, which is the reproduction.

## Secondary / latent gap (not a confirmed exploit)

`AutomationService.resolveProfileId` accepts a caller-supplied `profileId` and only checks that the profile belongs to the caller's **org**, it does not re-check the caller's **role** against profile access. Super Bot (which carries the `bash` tool) is seeded into every org (`ensureOrgSuperBotProfiles`), so that org-membership check passes for it.

`GET /v1/profiles` filters Super Bot out for viewers, so a viewer cannot obtain that id through the normal API, and the id is a `nanoid` (not guessable). But the automation layer trusting `profileId` without a role check is a defense-in-depth weakness worth closing alongside the primary fix, if a privileged profile id leaks by any channel, a viewer could bind an automation to it.

## Suggested fix (small, uses existing primitives)

- Apply `requireNotViewerFromContext` (or `requireOrgAdmin` where appropriate) to the mutating routes in `automations.ts`, `tasks.ts`, and `user-context.ts`.
- In `AutomationService.resolveProfileId` (or the route), enforce role-based profile access — the same rule `GET /v1/profiles` already applies — so a caller cannot bind an automation to a profile they may not use.
- Add per-route regression tests asserting `403` for `viewer` on each mutating endpoint.

This is intentionally the narrow authorization-enforcement slice of #58, not the broader Owner/Manager/Member role redesign that issue also describes.

## Environment

- `main` @ commit `f69b1c7`
- Server: `apps/server` (Hono + `bun:sqlite` / in-memory adapter)

I've reproduced this locally and am happy to open a PR with the guard additions plus the regression tests.

fajarhide:
Fixes #107. Two authorization gaps in the same area, one commit each.

**1. A viewer can reach state-changing automation/task routes.**
These routes only checked org context (`requireActiveOrgIdFromContext`), never `orgRole`. So a `viewer`, which is meant to be read-only, could create, update, delete and run automations and tasks. Running one executes an agent, so a viewer could trigger file writes/deletes, send email, fetch arbitrary URLs, and burn LLM cost.

Fix: add the existing `requireNotViewerFromContext` guard to the mutating routes (automations: `POST /draft`, `POST /`, `PUT /:id`, `DELETE /:id`, `POST /:id/run`, `DELETE /:id/runs/:runId`; tasks: `POST /draft-prompt`, `POST /`, `PUT /:id`, `DELETE /:id`, `POST /:id/run`). It runs before any service call, so it's a 403 before any side effect.

Repro on `main`, viewer session via the existing test harness:
```
POST /v1/automations          -> 201, automationService.create called 1x
POST /v1/automations/:id/run  -> 200, agent.runAutomation called 1x
```
Both should be 403.

**2. A caller-supplied `profileId` wasn't role-checked.**
`resolveProfileId` (both services) only checked the profile belongs to the caller's org, not that their role may use it. Super Bot, the `bash`-capable profile, is admin-only per `canAccessSuperBotProfile` and is seeded into every org, so a non-admin passing a Super Bot `profileId` on create could bind to it.

Fix: enforce that same admin-only rule at the binding point, 403 reusing the `createSession` message. Enforcement is opt-in via an access context passed from the HTTP routes, so the agent-tool path is unaffected (it binds the session's own, already-checked profile and has no role context).

**Left alone on purpose:** read routes (`GET` list/detail/runs/messages), and per-user self-service (`/v1/user/context`, run `mark-read`), since those only write the caller's own data.

**Tests:** `rbac-mutations.test.ts` asserts 403 for a viewer on every guarded route and that no service/agent call leaks through, plus an admin allow case. `profile-access.test.ts` covers member binding to Super Bot rejected, admin allowed, and the no-access path left ungated. Full `apps/server` suite: 359 pass, 0 fail.

**Scope:** just the authz enforcement slice of #58, not the Owner/Manager/Member redesign. The third commit (telegram timeout) is unrelated and only there to get CI green, happy to split it out.

### DIFF
--- a/apps/platform/telegram/src/chat-handler.test.ts
+++ b/apps/platform/telegram/src/chat-handler.test.ts
@@ -1,5 +1,5 @@
 import path from "node:path";
-import { describe, expect, test, spyOn, afterEach } from "bun:test";
+import { describe, expect, test, spyOn, afterEach, setDefaultTimeout } from "bun:test";
 import { TelegramAuthStore } from "./auth-store";
 import { createChatHandler } from "./chat-handler";
 import { UNSUPPORTED_DOCUMENT_TYPES_REPLY, UNSUPPORTED_MEDIA_REPLY } from "./attachments";
@@ -13,6 +13,10 @@ import {
   withTempHome,
   writeTelegramConfigIni,
 } from "./test-helpers";
+
+// These handler tests run in ~0.2s locally but occasionally exceed the 5000ms
+// default under CI's concurrent all-workspace load. Give them more headroom.
+setDefaultTimeout(10_000);
 
 async function waitForCondition(
   condition: () => boolean,
--- a/apps/server/src/http/routes/automations.ts
+++ b/apps/server/src/http/routes/automations.ts
@@ -11,7 +11,10 @@ import type {
   UpdateAutomationRequest,
 } from "@nakama/core";
 import { errorResponse, getRequestAuth, json, parseChannel, readJson } from "../shared";
-import { requireActiveOrgIdFromContext } from "../org-guards";
+import {
+  requireActiveOrgIdFromContext,
+  requireNotViewerFromContext,
+} from "../org-guards";
 import type { HonoApp } from "../types";
 import type { ServerOptions } from "../context";
 
@@ -169,6 +172,7 @@ export function registerAutomationRoutes(app: HonoApp, options: ServerOptions): 
   }));
 
   app.post("/v1/automations/draft", async (c) => {
+    requireNotViewerFromContext(c);
     const body = await readJson<DraftAutomationRequest>(c.req.raw);
     const automation = await agent.draftAutomation(body.prompt, parseChannel(body.channel));
     return json<DraftAutomationResponse>({ automation });
@@ -182,6 +186,7 @@ export function registerAutomationRoutes(app: HonoApp, options: ServerOptions): 
   });
 
   app.post("/v1/automations", async (c) => {
+    requireNotViewerFromContext(c);
     const orgId = requireActiveOrgIdFromContext(c);
     const body = await readJson<CreateAutomationRequest>(c.req.raw);
     const automation = await automationService.create(orgId, body, body.profileId);
@@ -201,6 +206,7 @@ export function registerAutomationRoutes(app: HonoApp, options: ServerOptions): 
   });
 
   app.put("/v1/automations/:automationId", async (c) => {
+    requireNotViewerFromContext(c);
     const orgId = requireActiveOrgIdFromContext(c);
     const automationId = decodeURIComponent(c.req.param("automationId"));
     const body = await readJson<UpdateAutomationRequest>(c.req.raw);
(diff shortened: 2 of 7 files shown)


ahmadrosid:
Nice catch. Is this pull request ready for review, or are you still finalizing it?

fajarhide:
> Nice catch. Is this pull request ready for review, or are you still finalizing it?

Yeah it's ready for review, nothing pending on my side. CI is green now.

Two heads-ups:

1. The last commit (telegram timeout bump) isn't related to the RBAC fix. The red CI earlier was 8 chat-handler tests hitting the 5s default under CI load, they run in ~0.2s locally. I put it here just to get CI green. It's more of a band-aid than a real fix, so happy to drop it or move it to its own PR if you want this one clean.

2. Scope is intentionally small, just the authz enforcement part of #58, not the full Owner/Manager/Member redesign. I left read routes and the per-user stuff (`/v1/user/context`, run `mark-read`) ungated on purpose since those only touch the caller's own data. Let me know if you want those gated too.

ahmadrosid:
Yeah, I think it's fine. It looks good to me. I'll merge it now.

PR #154, opened 2026-08-04
feat(coding-agent): add pi.dev as supported coding harness

fadlee:
## Summary
- Add `pi.dev` (pi) as a new coding agent harness kind alongside codex, claude_code, and opencode
- Provider passthrough maps Nakama provider types to pi.dev's provider system with correct env vars and CLI flags (`--provider`, `--model`)
- Custom base URLs (proxies/gateways) generate a "nakama" provider entry using `openai-completions` API format; default base URLs override built-in provider settings
- Model ID prefixes (e.g. `cx/`) are preserved for `openai_compatible` providers to avoid 404s from proxies
- Readiness check errors now include `combinedOutput` (stdout/stderr summary) for better debugging

#### Test plan
- [x] `bun test` for affected test files (coding-agent-command, coding-agent-spawn-env, coding-agent, create-automation) — 31 pass, 0 fail
- [x] Manual test: `pi -p "Reply with OK and nothing else."` with provider passthrough
- [x] Verified readiness check reports `ok: true`, `authenticated: true`, `ready: true`
- [x] Confirmed `--provider`, `--model`, and `spawn.env` are correctly set from Nakama routing

### DIFF
--- a/apps/server/src/services/coding-agent-command.test.ts
+++ b/apps/server/src/services/coding-agent-command.test.ts
@@ -47,6 +47,29 @@ describe("buildCodingAgentCommandTemplate", () => {
     expect(template.command).toContain("'Refactor auth module'");
   });
 
+  test("builds pi.dev command and spawn env", async () => {
+    const template = await buildCodingAgentCommandTemplate(
+      {
+        kind: "pi",
+        name: "pi.dev",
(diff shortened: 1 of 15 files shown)


ahmadrosid:
nice

Issue #188, opened 2026-08-07
feat(discord): allow starting a new guild thread while another is active


ahmadrosid:
## Problem

Discord guild conversations currently map **one thread per user per parent channel** (`g:{channelId}:u:{userId}` → thread id in `chat-threads.json`).

That means:

- Mentioning the bot again in the same channel **reuses** the existing thread instead of opening a new one.
- While an agent run is in progress, a second mention waits on the same per-user+channel chat lock, then still lands in that same thread.
- `/new` inside a thread starts a **new chat session**, but stays in the same Discord thread.

Users cannot open a second parallel bot conversation thread in the same channel (e.g. start a fresh topic while another thread is still running or archived in the mapping).

## Desired behavior

Support starting a **new Discord thread** for a new conversation from the parent guild channel, without being forced back into the previously mapped thread.

### Proposed UX (pick one or combine)

1. **Fresh mention from parent channel always creates a new thread** when the previous mapped thread is still active / locked, *or*
2. **`/new` (or a dedicated command) from the parent channel** creates a fresh Discord thread and remaps the user there.
3. Optionally keep reusing the mapped thread for in-thread follow-ups only (current in-thread “no mention needed” behavior).

Also clarify concurrency:

- Can two agent runs for the same user exist in **two different threads** of the same parent channel at once?
- If yes, lock/session keys should be per **conversation/thread**, not only per `user+parent channel`.

## Current implementation pointers

- Thread reuse: `apps/platform/discord/src/chat-handler.ts` (`resolveOrCreateGuildThread`)
- Lookup key: `resolveThreadLookupKey` in `apps/platform/discord/src/guild-message.ts`
- Persist map: `apps/platform/discord/src/thread-store.ts` → `chat-threads.json`
- Chat lock is currently `user + parent channel`, which serializes all activity for that pair across threads

Related: #186 / #187 (route guild conversations into threads)

## Acceptance criteria

- [ ] From a parent guild channel, a user can start a **new** bot thread without being forced into the previous one
- [ ] In-thread follow-ups continue to work in each thread (session isolation per thread)
- [ ] Org selection continues to inherit from the **parent channel** (not the thread id)
- [ ] Documented behavior for concurrent runs (allowed vs queued) and how `/new` / mention interact with thread creation
- [ ] Tests covering: second mention / explicit “new thread” path creates a distinct thread; old thread sessions remain independent

PR #190, opened 2026-08-07
feat(discord): allow starting a new guild thread while another is active

ahmadrosid:
## Summary
- Parent-channel `@mention` / reply-to-bot always creates a **new** Discord thread instead of reusing the mapped one; prior threads stay owned so in-thread follow-ups keep working.
- Chat locks are now per conversation/thread (`g:parent:t:threadId`), so the same user can run agents in two threads of one parent channel concurrently; same-thread messages still queue.
- `/new` in a parent channel creates a fresh Discord thread and session; `/new` inside a thread still only resets the Nakama session.
- `chat-threads.json` stays backwards compatible (legacy string values); multiple owned threads serialize as arrays.

Tracks #188 (does not close the issue).

## Test plan
- [x] `bun test apps/platform/discord/src/` (47 pass)
- [ ] In a guild channel: `@bot first topic` → thread A
- [ ] While A is running (or after): `@bot second topic` → distinct thread B
- [ ] Message in thread A still gets a reply with A’s session
- [ ] `/new` in the parent channel opens another thread and remaps ownership
- [ ] `/new` inside a thread stays in that Discord thread
- [ ] Org selection still comes from the parent channel

### DIFF
--- a/apps/platform/discord/src/chat-handler.test.ts
+++ b/apps/platform/discord/src/chat-handler.test.ts
@@ -480,10 +480,14 @@ describe("createChatHandler guild thread routing", () => {
     });
   });
 
-  test("second mention in the same channel reuses the existing thread", async () => {
+  test("second mention in the same channel creates a distinct thread", async () => {
     await withTempHome(async (homeDir) => {
-      const { handleMessage, threadStore } = await createPairedHandler(homeDir, {
-        onSendStream: async () => "Again",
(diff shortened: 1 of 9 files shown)


ahmadrosid:
Closing as a duplicate of #191, which matches the agreed #188 UX: parent-channel mentions always create a new Discord thread, and `/new` is left unchanged.

PR #195, opened 2026-08-07
Opt-in error reporting

fajarhide:
## Summary

Opt-in crash reporting for self-hosted installs. Today an uncaught exception or unhandled rejection in the server or a worker leaves nothing behind, and no request carries an id, so “it broke” has nothing to reproduce from.

This adds:

- One `reportError` path with request-scoped context (request id, hashed org/user/session ids, bounded breadcrumb kinds)
- Process handlers on server, CLI, and platform workers (WhatsApp keeps its survive-after-crash handlers, and now installs the delivery sink)
- Consent-gated delivery to a Sentry-compatible ingest
- Scrubbing before anything leaves the machine
- `nakama report --show` so the last stored report and the exact ingest event are inspectable

**Needs your call:** reports go to a Sentry-compatible ingest, and the default DSN points at a GlitchTip project from the original author. Point it at one you control and drop theirs, or keep it.

## What ships (after review cuts + hardening)

Core lives in a single module: `packages/core/src/crash-report.ts`. Consent, pending hold, scrubbing, and Sentry delivery are all there — no split package exports.

| Surface | Behavior |
| --- | --- |
| Consent | Tri-state in `~/.nakama/crash-reports.ini` (`unset` / `granted` / `denied`). Default `unset` sends nothing. |
| Prompt | Asked on the first crash in the CLI TTY, not at install. States that a scrubbed message+stack are sent; points at `nakama report --show`. |
| Overrides | `DO_NOT_TRACK=1` denies. `NAKAMA_CRASH_REPORTS` / `NAKAMA_CRASH_REPORT_DSN` override file/env. Empty DSN env disables delivery. |
| Pending | Bounded, fingerprint-deduped hold file. Validated + re-scrubbed on read. Failed deliveries stay pending. |
| Invariant | Worker start failure that previously only `console.warn`’d is reported at warning level. |

## What goes over the wire

| Sent | Never sent |
| --- | --- |
| Scrubbed error message + stack, error type | Chat content / prompts / tool args |
| Nakama, Bun, OS versions | Org names, emails, hostname |
| Breadcrumb kinds + timings | File contents, API keys, URI passwords, Basic auth, Slack webhooks |
| Hashed org / user / session ids (when wired) | Those ids in the clear |

## Intentionally removed during review

- `crash_issue` tool and `triage-crash-report` skill
- Agent tool-loop invariant hooks
- Orphaned automation-run boot sweep
- Split `crash-report-{config,pending,scrub,sentry}` modules (folded into one file)
- Breadcrumb *data* allowlist path (breadcrumbs are kind-only now)

## Review hardening (follow-up)

- Consent prompt matches the real outbound payload
- Flush no longer wipes pending when ingest fails
- Pending/last reports are type-checked and re-scrubbed before send/show
- Extra scrub patterns for URL userinfo, HTTP Basic, Slack webhooks
- WhatsApp installs the crash sink without changing Baileys exit semantics
- Uncaught path awaits report persistence before `process.exit`

## Also in this PR

- `@nakama/db` test script wiring + longer timeout for the sqlite reopen test on loaded CI runners

## Open calls before merge

- [ ] Default GlitchTip DSN ownership (keep, replace, or remove)
- [ ] Whether non-WhatsApp workers should hard-exit on `unhandledRejection` or survive like WhatsApp

## Test plan

- [ ] `bun test packages/core/src/crash-report*.test.ts apps/cli/src/crash-report-command.test.ts apps/server/src/http/crash-context.test.ts`
- [ ] Crash with consent unset → report held pending; CLI prompt appears; yes flushes, no clears
- [ ] Failed ingest leaves pending intact
- [ ] `nakama report --show` prints consent, endpoint, last report, and ingest payload
- [ ] `DO_NOT_TRACK=1` and `NAKAMA_CRASH_REPORT_DSN=` both prevent delivery


fajarhide:
Closing this in favour of smaller pieces. 2000 lines across capture, consent, delivery and a CLI command was too much to ask for in one review, and the history made it worse: nine of the sixteen commits deleted code the same PR had just added, so reading it commit by commit did not work either.

Split into:

- #233, unrelated to crash reporting: `@nakama/db` had no test script, so its 55 tests never ran in CI. Two files.
- #234, the capture half. Request context, `reportError`, scrubbing, process handlers. Nothing leaves the machine: no sink, no consent file, no endpoint in the diff. ~940 lines.
- Consent, the hold file, Sentry delivery and `nakama report --show` come after #234 lands, on `fajarhide:feat/crash-delivery`. That is also where the DSN ownership question belongs, so it no longer blocks the half that makes no network calls.

Splitting turned up a real bug worth naming: with consent granted, `reportError` fired the sink without awaiting it, and the uncaught handler calls `process.exit(1)` as soon as `reportError` resolves. Delivery lost that race on every actual crash, and the granted path never fell back to the hold file, so those reports were dropped entirely. Fixed on the delivery branch, with a test that fails on the old behavior: the send is awaited, and a failed send is held for the next flush instead of discarded.

PR #212, opened 2026-08-09
feat(web): add list/grid toggle on the Files page

ahmadrosid:
## Summary
- Files (`/files`) can switch between list and grid layouts; list stays the default and the choice persists in `localStorage` under `files-view-mode`.
- Grid cards reuse the same filtered artifact set as list view, with image thumbnails via `ArtifactAttachmentPreview`, type/size/date metadata, and the existing share/download/delete actions.
- Search, type filter, refresh, and delete/share behavior are unchanged.

## Test plan
- [x] Open `/files` with artifacts present — list view is default
- [x] Toggle to grid — cards show thumbnails/icons, type, size, date, and actions
- [x] Search and type filter update both views identically
- [x] Reload — grid preference is restored from `files-view-mode`
- [x] Download, share, and delete still work from grid cards
- [x] Narrow the viewport — grid collapses to fewer columns without overflow
- [x] `bun test apps/web/src/lib/files-page.shared.test.ts`
- [x] `bun run --cwd apps/web build`


ahmadrosid:
## Before / after

### Before
Files page before

### After — artifacts
Files page artifacts

### After — knowledge base
Files page knowledge base

PR #233, opened 2026-08-10
test(db): run the @nakama/db suite in CI

fajarhide:
`bun run test` fans out to every workspace with a test script. `@nakama/db` did not have one, so its 55 tests have never run in CI.

Turning it on surfaced one failure, fixed here. The reopen test runs migrations twice against a real sqlite file. Under a second locally, past bun's 5s default on a runner building nine workspaces at once, and the ENOENT that followed was `afterEach` removing the temp dir underneath the timed-out body rather than a second fault. It gets 30s. It is the only test in the workspace over 100ms.

```
$ bun run test
@nakama/db test:  55 pass
@nakama/db test:  0 fail
```

Split out of #195 so that PR is only about crash reporting.

### DIFF
--- a/packages/db/package.json
+++ b/packages/db/package.json
@@ -6,7 +6,8 @@
     ".": "./src/index.ts"
   },
   "scripts": {
-    "build": "bun build ./src/index.ts --outdir dist --target bun"
+    "build": "bun build ./src/index.ts --outdir dist --target bun",
+    "test": "bun test"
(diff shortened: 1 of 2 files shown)


fajarhide:
Dropping this. The db test script was only turned on as a side effect of the crash-reporting work, and that is not a good enough reason to change how CI fans out here.

PR #236, opened 2026-08-10
fix(server): release the session turn when the client cancels a stream

### LINKED ISSUE #216: Discord /stop leaves session turn locked — follow-ups get 409
## Summary

Discord `/stop` reports `Stopped.` to the user, but the **server session turn stays active**. Follow-up messages in the same thread fail with:

> A response is already in progress for this session.

Observed in Discord thread: https://discord.com/channels/1525955095433576620/1536009821231386725

## Repro

1. In a Discord bot thread, ask for a long-running agent turn (e.g. coding-agent / bash / ffmpeg work).
2. While it is still running, send `/stop`.
3. Bot replies `Stopping…` then `Stopped.`
4. Send another message in the same thread.

**Expected:** Agent accepts the new message (or clearly says it is still cancelling server-side work, then recovers).

**Actual:** Bot replies `A response is already in progress for this session.` immediately after `/stop`.

## Root cause

`/stop` only aborts the Discord worker’s local `AbortSignal` and cancels the SSE reader. It does **not** cancel the in-flight server turn.

- Discord: `stopActiveStream()` → client stream abort → Discord UX says stopped (`apps/platform/discord/src/active-stream.ts`, `chat-handler.ts`)
- Server: `POST /v1/sessions/:sessionId/messages` holds a turn in `sessionTurnRegistry.beginTurn()` until `endTurn()` (`apps/server/src/http/routes/sessions.ts`, `session-turn-registry.ts`)
- `streamMessage()` awaits `session.sendStream(...)` without wiring client disconnect/abort into cancellation; `endTurn()` only runs when that await finishes (`apps/server/src/http/shared.ts`)

So for long tool runs, Discord looks idle while the session remains locked → concurrent `beginTurn` returns `started: false` → **409** with that error string.

## Suggested fix

1. Propagate request/client abort into the agent stream / tool loop so `/stop` (and SSE disconnect) actually cancels server work.
2. Call `sessionTurnRegistry.endTurn(...)` promptly on abort so the session can accept new messages.
3. Cover with a regression test: abort mid-stream → next `messages` call succeeds (not 409).

## Workaround

`/new` (or `/clear`) to bind a fresh session id, or wait until the background turn finishes / hits stream timeout.

fajarhide:
`/stop` aborted only the Discord worker's own fetch. The server kept the turn in `sessionTurnRegistry` until `session.sendStream` settled, so the follow-up got 409 and the work behind the turn kept running.

Four parts. Each one was needed because the previous one left the reported symptom in place, and each was found by instrumenting a real session rather than by reading the code.

**Release the turn.** `streamMessage` derives a turn signal from the request abort and the response stream `cancel()`, and passes it into `sendStream`. The tool loop checks it before each provider call and before pushing an assistant message, and the signal reaches tools through `ToolContext`, so `bash` SIGTERMs its shell.

**Abort the provider request.** `generateReply` awaits the whole response, so an abort arriving while the model was thinking was not seen until that request finished by itself, and a plain chat turn stayed locked. `GenerateChatInput` now carries the signal and every provider hands it to its HTTP client: fetch init for openai, openai-compatible, cerebras, fireworks and the responses API, request options for the anthropic and openrouter SDKs, `config.abortSignal` for gemini. ollama and opencode-go inherit it by delegating.

**Retry the send that lands mid-abort.** A browser session still hit 409s. Instrumentation showed the follow-up POST arriving 29 to 44ms *before* the abort, three times out of three, colliding with a turn that was already dying. `sendStream` retries that specific 409 for up to 3s and gives up at once if the caller's own signal aborts or the error is anything else. It lives in `packages/client` because Discord, the web app and the CLI share the path.

**Keep Stop reachable.** The web still failed, and the console showed why: `stopStreaming` was never called at all. `showStop` was `canStop && !hasContent`, so typing the next message replaced Stop with Queue. A user who types while the agent is answering cannot cancel; the turn holds the session and every send returns 409. Across 98 rejected sends the server logged no abort and no endTurn.

## Verified

Cancelling a live turn on a running instance, then sending a follow-up:

```
before:  +208ms 409   +1017ms 409   +16117ms 409
after:   +3144ms HTTP 200 accepted
```

Ten regression tests, each checked against the unfixed code first. `bun run test` exits 0 across all eight workspaces.

The 29-44ms race needs two TCP connections with a proxy between them, so it does not reproduce from a script on loopback; its tests drive an injected fetch, and the timing evidence is the instrumented log.

Closes #216


ahmadrosid:
Amazing I'm merging this thank you 🙏

PR #240, opened 2026-08-11
fix(web): order MCP tool parameters deterministically

### LINKED ISSUE #215: test: parseMcpToolParameters order assertion fails on main
## Summary

`bun test` in `apps/web` fails on `src/lib/mcp-tool-schema.test.ts`:

```
(fail) parseMcpToolParameters > extracts parameter metadata from JSON schema objects
```

Pre-existing on `main` — unrelated to any open PR (verified: file untouched in #212).

## Repro

```
cd apps/web && bun test src/lib/mcp-tool-schema.test.ts
```

## Failure

`expect(received).toEqual(expected)` — array order mismatch:

- Expected first param: `path` (required: true)
- Received first param: `encoding` (required: false)

## Root cause

`parseMcpToolParameters` builds results with `Object.entries(properties)`, which iterates keys in insertion order. The test schema declares `encoding` before `path`, so the function returns `[encoding, path]`, while the test expects `[path, encoding]`. The order of parameters is not specified by the MCP schema spec, so this is a fragile assertion — either the implementation or the test needs to pin the order.

## Options

1. Sort parameters in `parseMcpToolParameters` (e.g. required-first, then by name) so output order is deterministic and documented.
2. Make the test order-agnostic (compare as a set / sort both sides).

Prefer option 1: a stable, documented ordering (required params first, then alphabetical) is more useful for callers rendering the tool schema.

fajarhide:
`parseMcpToolParameters` built its result with `Object.entries(properties)`, so the order came from however the MCP server serialised its schema. The test declares `encoding` before `path` and expects `path` first, so it has been red on main.

Pinned the order in the function rather than loosening the assertion, which is the option the issue prefers: required first, then alphabetical. `McpToolList` is the only caller and renders this list, so required params now sort to the top of it.

Before, on current `main`:

```
$ cd apps/web && bun test src/lib/mcp-tool-schema.test.ts
(fail) parseMcpToolParameters > extracts parameter metadata from JSON schema objects
 1 pass, 1 fail
```

After:

```
$ cd apps/web && bun test
 264 pass, 0 fail
Ran 264 tests across 47 files. [15.85s]
```

The added test builds its fixture from an array, not an object literal, because the formatter sorts literal keys and would otherwise make both halves of the order-independence check identical. Dropping the `.sort` turns both tests red.

Closes #215

### DIFF
--- a/apps/web/src/lib/mcp-tool-schema.test.ts
+++ b/apps/web/src/lib/mcp-tool-schema.test.ts
@@ -39,4 +39,27 @@ describe("parseMcpToolParameters", () => {
       },
     ]);
   });
+
+  it("orders required parameters first, then alphabetically, whatever the schema order", () => {
+    // Built from an array rather than an object literal on purpose: the
+    // formatter sorts literal keys alphabetically, which would make both
+    // fixtures identical and leave this test asserting nothing.
+    const schemaDeclaring = (declarationOrder: string[]) => ({
+      properties: Object.fromEntries(
+        declarationOrder.map((name) => [name, { type: "string" }])
+      ),
+      required: ["target", "cursor"],
+      type: "object",
+    });
+    const names = (declarationOrder: string[]) =>
+      parseMcpToolParameters(schemaDeclaring(declarationOrder)).map(
+        (parameter) => parameter.name
+      );
+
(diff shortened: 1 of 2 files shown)


ahmadrosid:
LGTM

PR #241, opened 2026-08-11
fix(server): stop a hung provider holding a session for 30 minutes

### LINKED ISSUE #237: Hung provider holds a session for 30 minutes, and the timeout timer is never cleared
A turn that never produces a reply holds its session for up to 30 minutes.

`STREAM_TIMEOUT_MS` comes from `resolveChatStreamTimeoutMs()`, and `DEFAULT_CHAT_STREAM_TIMEOUT_MS` is `1_800_000` (`packages/core/src/chat-stream-timeout.ts:1`). It is the only thing that ends a hung turn: `streamMessage` races `session.sendStream` against that timer, and `endTurn` runs in the `finally` (`apps/server/src/http/shared.ts:584-605`). Until it fires, `beginTurn` keeps returning `started: false` and every message in that session gets 409 (`apps/server/src/http/routes/sessions.ts:500-506`).

With #236 in, a client that cancels unwinds the turn in milliseconds, so this only bites when nothing cancels: a provider that accepts the connection and then goes quiet while the caller waits, or a non-interactive caller with no abort path. The workaround is to start a new session, which loses the thread.

Two things worth separating:

1. 30 minutes is a long time to hold a session for a provider that has sent nothing. The keepalive is 4s (`shared.ts:555`), so the server knows early that only pings are flowing.
2. The timer is never cleared. `new Promise<never>(...)` schedules a `setTimeout` and there is no `clearTimeout` anywhere in the file, so every completed turn leaves a pending 30-minute timer behind. Long-running servers accumulate one per turn.

Reproducing the lock does not need 30 minutes: set `NAKAMA_CHAT_STREAM_TIMEOUT_MS=60000` (the clamp minimum), point a profile at a base URL that accepts the request and never responds, send a message, then send a second one. The second returns 409 for the full 60s.

Worth deciding: a shorter default, a separate and shorter no-first-token deadline, or a way to clear a stuck turn without abandoning the session. The `clearTimeout` is a fix regardless.

fajarhide:
A provider that accepts the connection and then goes quiet held its session for the full 30 minute stream timeout. Nothing else could reach that turn, so every following message got a 409 and the only way out was to start a new session and lose the thread.

Three separate things were wrong in `streamMessage`, all in the same few lines.

**The deadline was the only thing that could end a silent turn, and it was 30 minutes.** That budget exists for a healthy turn sitting in a long tool loop, which really can take that long. It is far too generous for a provider that has sent nothing at all. This adds a second, much shorter deadline covering only the provider's first output. Any chunk, thinking delta or tool call satisfies it. The 4s keepalive deliberately does not, because it is the server's own ping and proves nothing about the provider. Default 120s, `NAKAMA_CHAT_FIRST_TOKEN_TIMEOUT_MS` to change it, `0` to switch it off.

**The `setTimeout` was never cleared.** Every turn scheduled one and left it pending for the rest of the window, so a long-running server holds a live timer per turn.

**A timeout never aborted `sendStream`.** The turn ended but the provider request stayed open against a closed stream. It aborts now, after the reject so the race still reports the timeout rather than the abort.

## Demo

`bun .local-poc/hung-provider-demo.ts`, deadlines scaled to 8s/1s so it finishes quickly. Production is 1800s/120s, same ratio.

```
stream deadline 8000ms, first-token deadline 1000ms

silent provider, first-token deadline OFF (behaviour before this PR)
  session held        8021ms of a 8000ms budget
  2nd message 409'd   true
  user is told        {"error":"Chat timed out after 8s waiting for the provider. ..."}

silent provider, first-token deadline ON
  session held        1007ms of a 8000ms budget
  2nd message 409'd   true
  user is told        {"error":"The provider accepted the request but sent nothing for 1s. ..."}

provider streamed then sat in a tool run, first-token deadline ON
  session held        8004ms of a 8000ms budget
  2nd message 409'd   true
  user is told        {"error":"Chat timed out after 8s waiting for the provider. ..."}
```

The third row is the one that matters for regressions: a turn that produced output keeps the full budget, so a long tool run is not collateral damage.

## Seen in the UI

Driven with Playwright against a throwaway instance (separate `NAKAMA_CONFIG_DIR`, black-hole provider, stream deadline pinned to its 60s floor so the run is watchable). One message, nothing queued behind it, measured from send to the composer coming back:

| | user waited | what the composer said |
| --- | --- | --- |
| first-token deadline off, which is `main` | 60.4s | `Chat timed out after 60s waiting for the provider.` |
| first-token deadline 5s | 5.6s | `The provider accepted the request but sent nothing for 5s.` |

60s there is the clamp floor, not the default. On the shipped default the first row is 30 minutes.

Worth knowing if you only ever test through the web client: it never shows the 409. A second message sent during a hung turn sits in the composer marked **Queued** and goes out when the turn ends, so the symptom a web user reports is a chat that has silently stopped accepting input, not an error. The 409 is what Discord, the CLI and direct API callers get.

## The issue's own repro, at the HTTP layer

#237 states it as "send a message, then send a second one. The second returns 409 for the full 60s". Run against a live server with the stream deadline pinned to its 60s floor, polling `POST /v1/sessions/:id/messages?stream=true` every 500ms until it stops returning 409:

| | second message stayed 409 | turn active, single message | error delivered |
| --- | --- | --- | --- |
| `main` | 63.8s | 60.2s | `Chat timed out after 60s waiting for the provider.` |
| this branch | 8.5s | 5.0s | `The provider accepted the request but sent nothing for 5s.` |

Both runs asserted the 409 before measuring, so the harness is not silently passing.

## Reproduce on a real instance

A black hole that accepts the connection and never answers:

```
bun -e 'Bun.serve({port:9099, fetch: () => new Promise(() => {})})'
```

Point a profile's provider Base URL at `http://localhost:9099/v1`, send a message, then send a second one. On `main` the second returns 409 until the stream timeout expires. With this branch it clears after the first-token deadline. Shorten both with `NAKAMA_CHAT_STREAM_TIMEOUT_MS=60000 NAKAMA_CHAT_FIRST_TOKEN_TIMEOUT_MS=5000` so you are not waiting 30 minutes to see it.

## Tests

```
$ cd apps/server && bun test src/http/routes/sessions.stream.test.ts
 8 pass, 0 fail

$ bun run test      # all workspaces
 exit 0
```

Each new test was checked against a broken build: disarming the first-token deadline reddens only the silent-provider test, never latching the sawProviderOutput flag reddens only the long-tool-run test, dropping `clearTimeout` reddens only the leak test, dropping the abort reddens only the original timeout test.

## Calls I would like your opinion on

**I did not shorten the 30 minute default.** With the first-token deadline in place, that budget only applies to turns the provider is actively feeding, and cutting it would kill legitimate long tool runs. Happy to lower it too if you would rather have both.

**I did not build a way to force-release a stuck turn**, which is the third option in the issue. The obvious version, calling `sessionTurnRegistry.endTurn` from a route, is the same mistake #236 rejected: the agent's `history` array is shared, so dropping the lock while the turn is still inside `executeToolCalls` lets the next turn append its user message and then `rollbackFailedSend` pops it back off. Doing it safely means the registry holds each turn's `AbortController` so a route can abort cooperatively, which is a new surface I did not want to invent unasked. Say the word and I will open it as its own issue.

Is 120s the right default? It is a guess at the worst honest first-token latency, and you have better data on which models people actually run here.

Closes #237. Both halves of the issue are covered: the hold is now bounded by the first-token deadline, and the timer is cleared. The force-release option it listed is the one alternative I did not take, and it belongs in its own issue rather than keeping this one open, since a turn that streams once and then hangs is a different failure from a provider that never speaks.

### DIFF
--- a/apps/server/src/http/routes/sessions.stream.test.ts
+++ b/apps/server/src/http/routes/sessions.stream.test.ts
@@ -33,6 +33,26 @@ function createCancellableSession(): {
   return { sawSignal: () => signal, session };
 }
 
+/** Emits once and then stalls: a healthy turn sitting in a long tool run. */
+function createChattyThenStalledSession(): AgentChatSession {
+  return {
+    getContextUsage: () => null,
+    sendStream: (
+      _input: unknown,
+      handlers: { onChunk: (delta: string) => void },
+      options?: { signal?: AbortSignal }
+    ) =>
+      new Promise<string>((_resolve, reject) => {
+        handlers.onChunk("working");
+        options?.signal?.addEventListener(
+          "abort",
+          () => reject(options.signal?.reason),
+          { once: true }
+        );
+      }),
+  } as unknown as AgentChatSession;
+}
+
 async function waitForTurnToEnd(sessionId: string): Promise<void> {
   for (let attempt = 0; attempt < 50; attempt += 1) {
     if (!sessionTurnRegistry.isActive(sessionId)) {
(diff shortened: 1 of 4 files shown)


ahmadrosid:
nice

PR #242, opened 2026-08-12
fix(core): cap web_fetch content so one call cannot flood the context

fajarhide:
`web_fetch` is the only tool that can put an unbounded amount of text into the model context. `bash` caps at 32,000 chars, `read_file` and `search_files` cap and report `truncated`, and #176 capped Composio results at 16,000. `web_fetch` has `MAX_BODY_BYTES` at 1 MB, but that bounds the transfer and throws above it, so anything under 1 MB goes in whole.

## Reproduce

Same command on both branches:

```sh
bun -e '
import { webFetchTool } from "./packages/core/src/tools/web-fetch";
const o = await webFetchTool.run(
  { raw: true, url: "https://backend.composio.dev/api/v3.1/openapi.json" }, {}
);
console.log(`body ${o.bytes} -> content ${o.content.length}, truncated ${o.truncated}`);
'
```

```
main    body 915095 -> content 914971, truncated undefined
branch  body 915095 -> content 16000,  truncated true
```

## What it fixes

That 914,971 character string is pushed into `history` as a `role: "tool"` message (`chat.ts:585`, `:611`) and re-sent on every later turn of the session until compaction runs. So it is not one expensive call, it is one call that stays expensive.

## The change

This follows `truncateComposioToolResult` in `apps/server/src/services/composio-tool-bridge.ts` rather than inventing anything: same 16,000, same `\n...[truncated]` marker, and the marker length subtracted from the budget so a capped result is exactly 16,000 characters and never one more. `truncated` joins the existing `bytes` on `WebFetchOutput`.

The cap runs after markdown conversion, so it bounds what the model reads rather than markup it never sees, and it applies to `raw: true` since that path skips conversion.

## What it costs

Checked against live URLs, not stubs. One ordinary page in my sample does get truncated:

| url | body | content | truncated |
|---|---|---|---|
| `backend.composio.dev/.../openapi.json` | 915,095 | 16,000 | true |
| `nakama/coding-agent.html` | 123,725 | 16,000 | true |

That page converts to 16,591 characters and loses 606 of them. The dropped tail is the page's own anchor-link nav, which VitePress emits at the end:

```
[How the pieces fit together](#how-the-pieces-fit-together)
[Supported coding agents](#supported-coding-agents)
[Setup](#setup)
```

Worth knowing rather than hiding: 16,000 is tight enough that a long documentation page can lose its last section, not just its nav. 32,000 would clear that and still catch the megabyte case. I went with 16,000 to match #176; happy to raise it if you would rather documentation pages always arrive whole.

Four tests cover it and three fail if `MAX_CONTENT_CHARS` is raised out of the way. `bun run test` is green across all eight workspaces.

### DIFF
--- a/packages/core/src/tools/web-fetch.test.ts
+++ b/packages/core/src/tools/web-fetch.test.ts
@@ -295,3 +295,71 @@ describe("convertHtmlToMarkdown", () => {
     expect(md).toContain("Self-hosted AI agents");
   });
 });
+
+describe("web_fetch content cap", () => {
+  // Same shape as truncateComposioToolResult: the marker is inside the budget,
+  // so a capped result is exactly CAP characters and never one more.
+  const CAP = 16_000;
+  const MARKER = "\n...[truncated]";
+
+  test("leaves a small body whole", async () => {
+    stubFetch(async () => htmlResponse("<h1>Short</h1><p>Body.</p>"));
+
+    const out = await webFetchTool.run({ url: "https://example.com" }, CTX);
+
+    expect(out.truncated).toBe(false);
+    expect(out.content).toContain("# Short");
+    expect(out.content).not.toContain(MARKER);
+  });
+
+  test("caps a long body and reports the original size", async () => {
+    // Long enough that the markdown is still over the cap after conversion.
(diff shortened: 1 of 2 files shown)


ahmadrosid (approved):
Thanks this is very useful.

PR #244, opened 2026-08-12
test: run the web, db and client suites in CI

### LINKED ISSUE #243: CI skips three workspaces, so 342 tests never run
`bun run test` is `bun run --workspaces --if-present test`, so a workspace without a `test` script is skipped silently. Three of them have no such script:

| workspace | test files | tests | run locally today |
|---|---|---|---|
| `apps/web` | 47 | 264 | pass, 5.3s |
| `packages/db` | 12 | 55 | pass, 1.0s |
| `packages/client` | 3 | 23 | pass, 3.6s |

342 tests that CI has never run.

#215 is what this costs. That test lives in `apps/web/src/lib/mcp-tool-schema.test.ts`, it was failing on `main`, and every PR in that window went green anyway because the workspace it sits in is skipped. It took a human noticing to file it, and #240 to fix it.

All three suites pass right now, so turning them on does not turn CI red. It adds about ten seconds.

The fix is `"test": "bun test"` in three `package.json` files, plus one timeout: `packages/db`'s reopen test runs migrations twice against a real sqlite file and needs more than bun's 5s default on a runner building nine workspaces at once.

Worth saying plainly: #233 proposed the `packages/db` half of this and I closed it myself, because it had only been turned on as a side effect of other work and that was not a good enough reason to change how CI fans out. #215 is the reason that was missing. Happy to send the PR if you want it.

fajarhide:
Closes #243.

Three `"test": "bun test"` lines and one timeout.

`bun run test` is `bun run --workspaces --if-present test`, so a workspace with no `test` script is skipped without a word. `apps/web`, `packages/db` and `packages/client` had none, which kept 62 test files and 342 tests out of every CI run since they were written.

That is what let #215 sit red on `main`. Its test is `apps/web/src/lib/mcp-tool-schema.test.ts`, so no PR in that window ever ran it, and it took someone noticing by hand to file it.

Before and after, same machine:

```
before   8 workspaces, 1,656 tests, 0 fail
after   11 workspaces, 1,998 tests, 0 fail
```

Nothing turns red, because all three suites already pass. `apps/web` 264, `packages/db` 55, `packages/client` 23. Roughly ten seconds on top.

The one non-obvious line is the 30s on `packages/db`'s reopen test. It runs migrations twice against a real sqlite file, 632ms here, and has gone past bun's 5s default on a runner building nine workspaces at once. The ENOENT that follows such a timeout is `afterEach` removing the temp dir underneath the timed-out body, not a second fault. The comment says so in place.

For the record, #233 proposed the `packages/db` third of this and I closed it myself: at the time the script had only been switched on as a side effect of other work, which was not a reason to change how CI fans out. #215 is the reason that was missing.

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -6,7 +6,8 @@
     "build": "tsc --noEmit && vite build",
     "dev": "bun run scripts/dev.ts",
     "preview": "vite preview",
-    "start": "vite preview"
+    "start": "vite preview",
+    "test": "bun test"
   },
   "dependencies": {
     "@base-ui/react": "^1.5.0",
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -6,7 +6,8 @@
     ".": "./src/index.ts"
   },
   "scripts": {
-    "build": "bun build ./src/index.ts --outdir dist --target bun"
+    "build": "bun build ./src/index.ts --outdir dist --target bun",
+    "test": "bun test"
   },
   "dependencies": {
     "@nakama/core": "workspace:*"
--- a/packages/db/package.json
+++ b/packages/db/package.json
@@ -6,7 +6,8 @@
     ".": "./src/index.ts"
   },
   "scripts": {
-    "build": "bun build ./src/index.ts --outdir dist --target bun"
+    "build": "bun build ./src/index.ts --outdir dist --target bun",
+    "test": "bun test"
   },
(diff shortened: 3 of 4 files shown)


ahmadrosid (approved):
Woahh nice, I see some time the test are run but the docker build are failing. Turns out some test are never run in CI.

PR #274, opened 2026-08-15
feat: skill curator — archive unused profile skills

ahmadrosid:
## Summary
- Deterministic skill curator: stale at 30 days, archive (move, never delete) at 90 days for agent/human profile skills
- Org-admin enable (default off), Run now, dry-run; 7-day scheduled tick with first-run seed
- Bundled skills and profiles with enabled automations are skipped; write approval does not block archive

## Test plan
- [x] Enable curator on an org, confirm first scheduled tick seeds and does not archive
- [x] Dry-run then live run: unused 90d skill moves to skills/.archive/ and leaves the catalog
- [x] Bundled skill unused 90d stays assigned
- [x] Profile with an enabled automation: skill is skipped
- [x] Write approval on: archive still applies, no skill proposal
- [x] Cross-org curator run returns 404

## Validation
Unit tests passed locally: archive primitive, curator service, org routes, schedule seed/due/disabled, and worker tick. Re-ran curator service + org-curator HTTP tests after the restore-miss follow-up.


fajarhide:
@ahmadrosid I spent the review on the part that touches disk, since that is where this one can hurt.

`archiveSkillDirectory` holds up: name validation, the bundled-name guard, containment checks on both the source and the destination, the `.archive` assertion, a timestamp suffix on collision, and a restore that refuses to overwrite a live skill. `discover.ts` skipping `SKILL_ARCHIVE_DIR_NAME` closes the loop that would otherwise rediscover what the curator just moved. `archiveAssignedSkill` moves first, then touches the database, and compensates by moving back.

Three things.

**Archiving is one way for the operator.** `restoreArchivedSkillDirectory` exists only as the rollback inside that catch. Nothing in the API or the dashboard can bring a skill back, so "move, never delete" is true on disk and not true in the product: getting it back means shell access to `skills/.archive/`. An archived list with a restore action, or a docs line naming the path, would close that.

**A failed restore goes silent.** In the catch, if the restore itself throws, the skill stays in `.archive` while the row and the assignment are back at the live path, and the run counts it as a skip. Nothing names which skill ended up in that state. `run.json` and `REPORT.md` are already written per run, so logging the skill id and the archived path there would give an operator the one thing they need to fix it by hand.

**Freshness reads `lastUsedAt`, which only `recordMatches` and `recordPatch` write.** `recordCatalogViews` writes `viewedAt` instead, so a skill the model keeps reading in the catalog but never matches counts as unused and archives at 90 days. If that is deliberate, worth saying so in `self-improving-skills.mdx`, since it is the case an operator will report as a bug.

The test plan in the body is still unticked, and this has been open since 15 August. Worth walking it on a throwaway instance before merge, if only for the "profile with an enabled automation is skipped" and "cross-org run returns 404" lines.

ahmadrosid:
Restore stays disk-only in this PR — path and steps are in `self-improving-skills.mdx`. A restore miss now writes skill id + archived path to `run.json` / `REPORT.md`. Unused is match/patch only; catalog views do not reset the clock (same page).

Re-ran curator service + org-curator HTTP tests just now: enabled-automation skip and cross-org 404 both pass.

PR #276, opened 2026-08-15
fix(web): raise the text floor to 11px and drop unreadable muted text

fajarhide:
Two styling defects, both measured against a running instance.

## Font sizes

Sixty font sizes were written as one-off arbitrary values, and there was no scale step below `text-xs` to reach for. This adds one, `--text-2xs` at 11px, and replaces all of them.

| value | uses | before | after |
| --- | --- | --- | --- |
| `text-[11px]` | 27 | 11px | 11px |
| `text-[0.7rem]` | 2 | 11.2px | 11px |
| `text-[10px]` | 16 | 10px | 11px |
| `text-[0.65rem]` | 3 | 10.4px | 11px |
| `text-[0.6rem]` | 9 | 9.6px | 11px |
| `text-[9px]` | 2 | 9px | 11px |
| **arbitrary sizes left** | **59** | **59** | **0** |

Twenty-nine were already 11px or 11.2px and do not move. The rest grow by at most 2px. The three tightest containers were checked by hand: the 18px notification badge sets `leading-none`, and the 16px avatar fallback holds a single character.

## Contrast

Every opacity modifier on muted text fails WCAG AA. Measured by compositing the real token over each surface in the browser and reading the pixels back, so no oklch is parsed by hand:

| muted-foreground over | /100 | /70 | /60 | /50 |
| --- | --- | --- | --- | --- |
| light, background | 4.74 | 2.71 | 2.30 | 1.96 |
| light, card | 4.74 | 2.71 | 2.30 | 1.96 |
| light, muted | 4.35 | 2.58 | 2.21 | 1.90 |
| dark, background | 7.72 | 4.26 | 3.41 | 2.71 |
| dark, card | 6.91 | 4.04 | 3.30 | 2.71 |
| dark, muted | 6.41 | 3.85 | 3.20 | 2.63 |

The floor is 4.5:1, so all 16 uses of `text-muted-foreground/50`, `/60` and `/70` fail in both themes. They now use plain `text-muted-foreground`.

`text-foreground/70` is deliberately untouched: it measures 7.40:1 at worst.

## What this does not fix

Two failures are real, pre-existing, and want their own change:

- Primary buttons are 3.19:1. White text on the amber `--primary` at 12.8px, on `Save`, `Create tool`, `Create automation`, `Download`, `New chat`. Eight of the nine remaining failures found below. It is the brand accent, so moving it is not a drive-by.
- Plain `text-muted-foreground` over `bg-muted` is 4.35:1 in light, just under the floor. Stock shadcn token; changing it moves every muted string in the app.

## Before and after

| before | after |
| --- | --- |
|  |  |

`/profiles` in light. The change is deliberately subtle: 3741 pixels differ, 0.081% of the frame, and the smallest rendered font is 11px on both sides. A layout shift would move far more than that, so this is colour only.

A DOM sweep of every rendered text element on `/settings`, `/system`, `/profiles`, `/automations` and `/history`, in both themes, counts 9 failures before and 9 after. That is not a null result to hide: the sixteen classes this PR changes are not rendered on those five routes, because they live in chat, tasks, the skill picker, org memory and the model browse lists, which need seeded data or an open dialog. The nine that remain are the two cases listed above.

`bun run test`: 11 workspaces, 2060 pass, 0 fail. `bun run build` (tsc plus vite) clean.


ahmadrosid:
nice

PR #279, opened 2026-08-15
feat(web): Cmd+K palette to jump to any page the sidebar offers

### LINKED ISSUE #278: Cmd+K command palette for navigation: cmdk and CommandDialog are already here, nothing opens them globally
`cmdk` is a dependency and `CommandDialog` already exists in `components/ui/command.tsx`, but both are only used for pickers inside other dialogs: the toolkit, MCP server, skill, provider and timezone selects. There is no global shortcut. Grepping the web app for `metaKey` or `ctrlKey` finds exactly one handler, in `chat/AgentQuestionnairePanel.tsx`.

Meanwhile the console has eight nav items across three groups, plus standalone pages like notifications, the skill detail and tool playground routes, and a sidebar that collapses to icons.

Proposal, deliberately small: Cmd+K or Ctrl+K opens `CommandDialog` listing the destinations the current user can actually reach, and Enter navigates there. It reuses `NAV_GROUPS`, `canAccessSystemPage`, `canAccessIntegrationsPage` and `PLATFORM_ADMIN_PAGE_IDS`, the same gates `Layout` already applies, so the palette can never offer a page the sidebar hides.

Out of scope for a first cut: actions rather than destinations, searching sessions, skills or artifacts, recents, and per-page sub-commands. Navigation only, so it stays reviewable.

Happy to send the PR if this is wanted.

fajarhide:
Closes #278.

`cmdk` and `CommandDialog` were already here, used only for pickers inside other dialogs. Nothing opened them globally, and the web app had exactly one keyboard handler.

Cmd+K or Ctrl+K now opens a palette listing every nav destination the current user can reach, plus notifications, which is reachable but absent from the sidebar. Enter navigates, Escape closes.

## Before and after

Driven against a running instance with CDP, dispatching a real Meta+K:

| | before | after |
| --- | --- | --- |
| Cmd+K on `/chat` | nothing, dialog absent | dialog opens, 4 groups, 9 items |
| type `settings` | n/a | filters to Settings and System |
| Enter | n/a | navigates to `/settings`, dialog closes |
| Escape | n/a | closes, path unchanged |

## The destination list is not a second copy

The role gating that lived inline in `Layout` moves to `visibleNavGroups` in `lib/navigation`, and both the sidebar and the palette read it, so the palette cannot offer a page the sidebar hides. That function gets the tests: a platform admin sees all eight, an org admin gets System but not the platform-admin pages, a member loses System, a viewer also loses Integrations, and groups left empty are dropped.

The org-admin case is worth a look in review. `soul` is in `PLATFORM_ADMIN_PAGE_IDS` yet reachable by an org admin, because the `canAccessSystemPage` branch runs before the platform-admin check. That is existing behaviour, moved as-is and now pinned by a test rather than changed.

## Trap worth naming

`CommandDialog` drops its children straight into `DialogContent` without a cmdk root, so the parts render with no context and the dialog comes up empty. The `Command` wrapper has to come from the caller. That cost a debugging round here.

Out of scope, as in the issue: actions rather than destinations, searching sessions, skills or artifacts, recents, and per-page sub-commands.

`bun run test`: 11 workspaces, 2063 pass, 0 fail. `bun run build` clean.


ahmadrosid:
i like this

PR #280, opened 2026-08-15
feat(skills): install a public GitHub SKILL.md onto a profile

### LINKED ISSUE #262: feat: install skill from GitHub
## Summary

Paste a GitHub URL to a `SKILL.md` (or a folder that contains one). Server fetches the file, validates frontmatter, and creates + assigns the skill with the existing create path.

Not a marketplace (#76). No catalog, preview API, version pin, or update loop.

## Problem

Skills are created by typing into **Create skill** or via `skill_manage`. A public GitHub skill cannot be installed without copy-paste.

## Design (keep it small)

`POST /v1/skills/install`

```ts
interface InstallSkillRequest {
  url: string;
  profileId: string;
}
```

Returns `SkillResponse` (201). Platform admin only.

Flow:

1. Accept a GitHub `blob` / `raw` URL to `SKILL.md`, or a `tree` URL whose folder has `SKILL.md`
2. Fetch that file from `raw.githubusercontent.com` (allowlist only — no open fetch)
3. Parse with `parseSkillMarkdown`
4. Write + assign via `createAndAssignRawSkillToProfile`

Reuse create. Do not add a new skill store, source SHA, or install registry.

## UI

On the profile Skills section, next to **Create skill**: **Install from GitHub** — one URL field, submit, skill appears assigned.

## Out of scope

- Supporting files / `tool.js` (SKILL.md only)
- Global library install (profile only)
- Preview endpoint
- Write-approval proposals
- Private repos / `GITHUB_TOKEN`
- Re-install / update from upstream
- Non-GitHub hosts

## Tests

- Valid public `SKILL.md` URL creates and assigns the skill
- Missing / invalid frontmatter fails; no skill row
- Non-GitHub URL → 400
- Non-admin → 403

## Test plan

- [ ] Install a public GitHub `SKILL.md` onto a profile; it shows up assigned
- [ ] Bad URL or bad markdown shows an error and writes nothing

akmalsyrf:
## Summary

Closes #262.

Platform admins can paste a public GitHub URL to a `SKILL.md` (or a folder that contains one) and install it onto the active profile in one step — no copy-paste into **Create skill**.

This is intentionally small: fetch one markdown file from an allowlisted host, validate frontmatter with the existing parser, then reuse `createAndAssignRawSkillToProfile`. No marketplace, no supporting files, no private repos, no update-from-upstream loop.

### Why

Today the only ways to get a public skill into a profile are typing it into **Create skill** or asking an agent with `skill_manage`. Public GitHub skills still require manual copy-paste. Operators already share skills as repo folders; Nakama should accept that URL directly.

### API

`POST /v1/skills/install`

```ts
interface InstallSkillRequest {
  url: string;
  profileId: string;
}
```

- **Auth:** platform admin + active org (`X-Org-Id`)
- **Response:** `SkillResponse` with **201**
- Registered **before** `/v1/skills/:skillId` so `"install"` is not captured as an id

### Flow

1. Accept a GitHub `blob` / `raw` URL to `SKILL.md`, a `tree` URL to a skill folder, or a `raw.githubusercontent.com` URL
2. Normalize to `https://raw.githubusercontent.com/{owner}/{repo}/{ref}/…/SKILL.md`
3. Fetch **only** from `raw.githubusercontent.com` (allowlist; redirects rejected; size + timeout capped)
4. `parseSkillMarkdown` — reject missing/invalid frontmatter before writing
5. `createAndAssignRawSkillToProfile` — same write + assign path as agent `skill_manage` create

### URL examples that work

| Input | Resolved raw URL |
| --- | --- |
| `https://github.com/org/repo/blob/main/skills/foo/SKILL.md` | `…/main/skills/foo/SKILL.md` |
| `https://github.com/org/repo/tree/main/skills/foo` | appends `SKILL.md` |
| `https://github.com/org/repo/raw/main/skills/foo/SKILL.md` | same raw target |
| `https://raw.githubusercontent.com/org/repo/main/skills/foo/SKILL.md` | used as-is (after normalize) |

Non-GitHub hosts and blob paths that are not `SKILL.md` return **400**.

### UI

On **System → Profiles → Skills**, next to **Create skill**:

- **Install from GitHub** opens a dialog with a single URL field
- Submit installs and assigns; the skill appears on the profile list
- Errors (bad URL, bad markdown, fetch failure) surface in the dialog

### Client / contract

- `InstallSkillRequest` in `@nakama/core`
- `client.installSkill(request)` in `@nakama/client`
- `useInstallSkillMutation` invalidates skills + profile queries

### Out of scope (per #262)

- Supporting files / `tool.js`
- Global library install (profile only)
- Preview endpoint
- Write-approval proposals
- Private repos / `GITHUB_TOKEN`
- Re-install / update from upstream
- Non-GitHub hosts

## Test plan

- [x] `bun test packages/core/src/skills/github-skill-url.test.ts`
- [x] `bun test apps/server/src/http/routes/skills-install.test.ts`
  - [x] Valid public `SKILL.md` URL → **201**, skill created and assigned
  - [x] Missing / invalid frontmatter → **400**, no skill row written
  - [x] Non-GitHub URL → **400**
  - [x] Non-platform-admin → **403**
- [x] Manual: open a profile as platform admin → **Install from GitHub** → paste a public `SKILL.md` blob URL → skill shows as assigned
- [x] Manual: paste a bad URL or markdown-without-frontmatter → error shown, list unchanged
- [x] Manual: org member / viewer cannot call install (API 403; UI is platform-admin gated with other skill mutations)

## Implementation map

| Area | Files |
| --- | --- |
| URL normalize | `packages/core/src/skills/github-skill-url.ts` |
| Allowlisted fetch | `packages/core/src/skills/github-skill-fetch.ts` |
| Service | `SkillsService.installSkillFromGitHub` |
| Route + OpenAPI | `apps/server/src/http/routes/skills.ts` |
| Tests | `skills-install.test.ts`, `github-skill-url.test.ts` |
| UI | `SkillInstallDialog.tsx` + profile Skills section |

### DIFF
--- a/apps/server/src/http/routes/skills-install.test.ts
+++ b/apps/server/src/http/routes/skills-install.test.ts
@@ -229,4 +229,103 @@ describe("POST /v1/skills/install", () => {
 
     expect(response.status).toBe(403);
   });
+
+  test("incomplete body returns 400, not 500", async () => {
+    const { app, databaseAdapter } = createApp();
+    const adminSession = await setupFreshInstallSession(
+      app,
+      databaseAdapter,
+      "admin-malformed@org.com"
+    );
+    const orgId = adminSession.orgId!;
+
+    const response = await app.fetch(
+      new Request(`${BASE}/v1/skills/install`, {
+        body: JSON.stringify({ profileId: "p" }),
+        headers: adminSession.headers(
+          {
+            "Content-Type": "application/json",
+            "X-CSRF-Token": adminSession.csrfToken,
+          },
+          orgId
+        ),
+        method: "POST",
+      })
+    );
+
+    expect(response.status).toBe(400);
+    const body = (await response.json()) as { error: string };
+    expect(body.error).toMatch(/url is required/i);
+  });
+
+  test("installing the same skill onto a second profile returns 409", async () => {
+    globalThis.fetch = mock(
+      async () => new Response(VALID_SKILL, { status: 200 })
+    ) as typeof fetch;
+
+    const { app, databaseAdapter } = createApp();
+    const adminSession = await setupFreshInstallSession(
+      app,
+      databaseAdapter,
+      "admin-dup@org.com"
+    );
+    const orgId = adminSession.orgId!;
+    const profiles = await databaseAdapter.listProfilesForOrg(orgId);
+    const firstProfileId = profiles[0]!.id;
+
+    const first = await app.fetch(
+      new Request(`${BASE}/v1/skills/install`, {
+        body: JSON.stringify({
+          profileId: firstProfileId,
+          url: "https://github.com/acme/skills/blob/main/weather/SKILL.md",
+        }),
+        headers: adminSession.headers(
+          {
+            "Content-Type": "application/json",
+            "X-CSRF-Token": adminSession.csrfToken,
+          },
+          orgId
+        ),
+        method: "POST",
+      })
+    );
+    expect(first.status).toBe(201);
+
+    const now = new Date().toISOString();
+    const secondProfileId = "profile_second";
+    await databaseAdapter.upsertProfile({
+      createdAt: now,
+      id: secondProfileId,
+      isDefault: false,
+      isSuper: false,
+      model: null,
+      name: "Second",
+      orgId,
+      skillsPostTurnReview: false,
+      skillsWriteApproval: false,
+      systemPrompt: "",
+      updatedAt: now,
+    });
+
+    const second = await app.fetch(
+      new Request(`${BASE}/v1/skills/install`, {
+        body: JSON.stringify({
+          profileId: secondProfileId,
+          url: "https://github.com/acme/skills/blob/main/weather/SKILL.md",
+        }),
+        headers: adminSession.headers(
+          {
+            "Content-Type": "application/json",
+            "X-CSRF-Token": adminSession.csrfToken,
+          },
+          orgId
+        ),
+        method: "POST",
+      })
+    );
+
+    expect(second.status).toBe(409);
+    const body = (await second.json()) as { error: string };
+    expect(body.error).toMatch(/already exists|cannot be attached/i);
+  });
 });
--- a/apps/server/src/http/routes/skills-install.test.ts
+++ b/apps/server/src/http/routes/skills-install.test.ts
@@ -1,0 +1,232 @@
+import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
+import { createInMemoryDatabaseAdapter } from "@nakama/db";
+import { SkillsService } from "../../services/skills-service";
+import { setupTestConfigDir } from "../../test-config-dir";
+import { createMinimalHonoApp } from "../test-app-helpers";
+import {
+  loginUserSession,
+  setupFreshInstallSession,
+} from "../test-session-helpers";
+
+setupTestConfigDir("nakama-skills-install-routes-test-");
+
+const VALID_SKILL = `---
+name: github-weather
+description: Get weather from a public GitHub skill.
+---
+
+Call the weather tool.
+`;
+
+const INVALID_SKILL = `# No frontmatter
+
+Just a body.
+`;
+
+function createApp() {
+  const databaseAdapter = createInMemoryDatabaseAdapter();
+  const skillsService = new SkillsService(databaseAdapter);
+  return {
+    ...createMinimalHonoApp({
+      agent: {
+        installSkillFromGitHub: (orgId: string, request: unknown) =>
+          skillsService.installSkillFromGitHub(
+            orgId,
+            request as { profileId: string; url: string }
+          ),
+        listProfiles: async () => ({ profiles: [] }),
+      },
+      databaseAdapter,
+    }),
+    skillsService,
+  };
+}
+
+const BASE = "http://localhost:4310";
+
+describe("POST /v1/skills/install", () => {
+  const originalFetch = globalThis.fetch;
+
+  beforeEach(() => {
+    globalThis.fetch = originalFetch;
+  });
+
+  afterEach(() => {
+    globalThis.fetch = originalFetch;
+  });
+
+  test("platform admin installs a valid public SKILL.md and assigns it", async () => {
+    globalThis.fetch = mock(async (input: RequestInfo | URL) => {
+      const url = String(input);
+      expect(url).toBe(
+        "https://raw.githubusercontent.com/acme/skills/main/weather/SKILL.md"
+      );
+      return new Response(VALID_SKILL, { status: 200 });
+    }) as typeof fetch;
+
+    const { app, databaseAdapter } = createApp();
+    const adminSession = await setupFreshInstallSession(
+      app,
+      databaseAdapter,
+      "admin@org.com"
+    );
+    const orgId = adminSession.orgId!;
+    const profiles = await databaseAdapter.listProfilesForOrg(orgId);
+    const profileId = profiles[0]!.id;
+
+    const response = await app.fetch(
+      new Request(`${BASE}/v1/skills/install`, {
+        body: JSON.stringify({
+          profileId,
+          url: "https://github.com/acme/skills/blob/main/weather/SKILL.md",
+        }),
+        headers: adminSession.headers(
+          {
+            "Content-Type": "application/json",
+            "X-CSRF-Token": adminSession.csrfToken,
+          },
+          orgId
+        ),
+        method: "POST",
+      })
+    );
+
+    expect(response.status).toBe(201);
+    const body = (await response.json()) as {
+      skill: { name: string; id: string };
+    };
+    expect(body.skill.name).toBe("github-weather");
+
+    const assigned = await databaseAdapter.listSkillsForProfile(profileId);
+    expect(assigned.some((skill) => skill.id === body.skill.id)).toBe(true);
+  });
+
+  test("invalid frontmatter returns 400 and writes no skill", async () => {
+    globalThis.fetch = mock(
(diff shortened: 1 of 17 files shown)


akmalsyrf:
Screenshot 2026-08-15 at 18 02 35

akmalsyrf:
Screenshot 2026-08-15 at 18 04 03

akmalsyrf:
Screenshot 2026-08-15 at 18 05 40

akmalsyrf:
Screenshot 2026-08-15 at 18 34 16

fajarhide:
Full disclosure: I built the same feature before I noticed this PR. That is on me for checking upstream branches instead of open PRs, which cannot see a branch sitting in another fork. Yours is the better version and mine stays unpushed. The `getProfileForOrg` check, the OpenAPI registration, `withDisabledFetchIdle`, the `/raw/` and `www.` handling and the route test are all things mine either missed or skipped, and that org check closes an authorization hole my version had.

Since I had the same code fresh in my head I gave this a careful pass. Four findings, all reproduced against `3e086123`.

## 1. An encoded path can send the fetch to a different repo

`splitPath` runs `decodeURIComponent` on every segment and rejoins without re-encoding, so `%2F` becomes a real separator and `..` survives into the URL that gets fetched.

```
input       https://github.com/trusted-org/skills/blob/main/..%2F..%2F..%2Fattacker%2Fevil%2Fmain%2FSKILL.md
resolved    https://raw.githubusercontent.com/trusted-org/skills/main/../../../attacker/evil/main/SKILL.md
normalized  https://raw.githubusercontent.com/attacker/evil/main/SKILL.md
```

The host check still passes because the host never changes. I first wrote this off as harmless for that reason, which was wrong: the boundary users actually rely on here is the owner and repo, not the host. An admin who vets the pasted link by its visible `trusted-org/skills` installs agent-executable instructions from `attacker/evil`.

`%23` is a milder version of the same thing: `we%23ird/SKILL.md` turns into a fragment, so the fetch quietly targets a different path.

Rejecting any segment that contains an encoded separator, or keeping the raw segments and only stripping the `blob|tree|raw` element, would close it.

## 2. Installing the same skill onto a second profile returns 500

Normal flow, not an edge case. Every conflict from `createAndAssignRawSkillToProfile` is a plain `Error`, so `app.onError` maps it to 500 and the agent-facing wording reaches an admin dialog.

Install any public skill onto one profile, then the same URL onto another profile in the same org:

```
profile 1 -> HTTP 201
profile 2 -> HTTP 500 {"error":"Skill \"greeter-demo\" already exists at a different source path and cannot be attached to this profile."}
```



Re-installing after the upstream file changed, and a SKILL.md named after a bundled skill, land the same way. Wrapping that call and re-throwing as 400 or 409 covers all three.

## 3. A body missing either field returns 500 and leaks the internal error

`.trim()` runs before any presence check, and the route schema is `z.object({}).passthrough()`, so nothing rejects a partial body first.

```
{}                               -> 500 {"error":"undefined is not an object (evaluating 'request.profileId.trim')"}
{"profileId":"p"}                -> 500 {"error":"undefined is not an object (evaluating 'request.url.trim')"}
{"url":".../blob/main/SKILL.md"} -> 500 {"error":"undefined is not an object (evaluating 'request.profileId.trim')"}
```

A `TypeError` is not a `NakamaApiError`, so the `is required.` 400s just below are unreachable for a missing field. `request.profileId?.trim() ?? ""` restores them and matches `agent-service.ts:1880`, `mcp-service.ts:98` and `org-service.ts:578`.

Runnable repro, fails on this branch with `Expected: 400, Received: 500`:

```ts
// apps/server/src/http/routes/skills-install-malformed.test.ts
import { createInMemoryDatabaseAdapter } from "@nakama/db";
import { SkillsService } from "../../services/skills-service";
import { setupTestConfigDir } from "../../test-config-dir";
import { createMinimalHonoApp } from "../test-app-helpers";
import { setupFreshInstallSession } from "../test-session-helpers";

setupTestConfigDir("nakama-skills-install-malformed-");

describe("POST /v1/skills/install with an incomplete body", () => {
  test("returns 400, not 500", async () => {
    const skillsService = new SkillsService(databaseAdapter);
    const { app } = createMinimalHonoApp({
      agent: {
        installSkillFromGitHub: (orgId: string, request: unknown) =>
          skillsService.installSkillFromGitHub(
            orgId,
            request as { profileId: string; url: string }
          ),
        listProfiles: async () => ({ profiles: [] }),
      },
      databaseAdapter,
    });
    const session = await setupFreshInstallSession(
      app,
      databaseAdapter,
      "admin@org.com"
    );

    const response = await app.fetch(
      new Request("http://localhost:4310/v1/skills/install", {
        body: JSON.stringify({ profileId: "p" }),
        headers: session.headers(
          {
            "Content-Type": "application/json",
            "X-CSRF-Token": session.csrfToken,
          },
          session.orgId!
        ),
        method: "POST",
      })
    );

    expect(response.status).toBe(400);
  }, 20_000);
});
```

## 4. Minor: the size cap is checked after the body is buffered

`await response.arrayBuffer()` allocates the whole response before `MAX_SKILL_BYTES` is compared, and raw.githubusercontent.com will serve far more than 512KB. Checking `Content-Length` first, or reading the stream and aborting past the cap, avoids paying for the bytes you are about to reject.

## What held up

I threw 14 adversarial URLs at the resolver: lookalike host, `@` userinfo trick, `file:`, loopback, literal traversal, all refused. Credentials, ports and `http:` are accepted but harmless since you rebuild the URL from segments. `redirect: "error"` plus the abort signal covers the redirect and stall cases. `bun test packages/core apps/server` is green on this branch at 1344 pass, and your own tests pass 10 of 10.

One thing that is not yours: fetching raw.githubusercontent.com from a long-lived bun server timed out on me a few times, then worked after a restart. I saw the same on my own implementation, so it looks environmental rather than anything in this diff.

akmalsyrf:
Thanks for the careful pass — all four findings were valid and reproduced.

Addressed in `472779da`:

1. **Encoded path escape** — reject segments with encoded separators (`%2F`, `%5C`, `%23`, `%3F`), reject unsafe decoded segments (`.`, `..`, etc.), and rebuild the raw URL with `encodeURIComponent` per segment.
2. **Same skill → second profile** — wrap `createAndAssignRawSkillToProfile` and map conflicts (already exists / assigned / bundled) to **409**; other install failures to **400**.
3. **Incomplete body** — `request.profileId?.trim() ?? ""` / same for `url`, so missing fields hit the existing **400** paths instead of a `TypeError` **500**.
4. **Size cap** — check `Content-Length` first, then stream-read and abort once the body exceeds 512KB.

Regression coverage added for the encoded-separator URLs, incomplete body → 400, duplicate-profile install → 409, and both Content-Length / streamed oversize cases.

akmalsyrf:
## Product question: same skill on a second profile

While reviewing the install path after the security hardening, one behavior looks more like a product choice than a bugfix.

`installSkillFromGitHub` always goes through `createAndAssignRawSkillToProfile`. If skill name `foo` already exists under profile A’s skills directory, installing the same GitHub URL onto profile B returns **409** (`already exists at a different source path and cannot be attached`).

That is correct for the current write path, but it is surprising for the dashboard action “Install from GitHub” on another profile. Today you can still attach the existing skill to profile B via **Add skills** / `assignSkill`, so install and assign disagree.

Possible product directions:

1. **Keep 409** — skills are unique by name in the org; second profile must use **Add skills** (or rename). Document that in the UI error.
2. **Assign existing** — if the name already exists, just `assignSkillToProfile` (no second copy). Profile B then depends on profile A’s on-disk skill dir.
3. **Copy into the target profile** — write a fresh `SKILL.md` under profile B’s skills dir and create/assign that record, so installs are independent per profile.

Which should install do?

Related, smaller metadata question: this path currently records `createdBy: "agent"` because it reuses the agent write helper. Dashboard install is a human/admin action — should we mark it `"human"` instead?

fajarhide:
Hi @akmalsyrf 

Re-checked the four findings against `472779da`, all closed. `bun test packages/core/src/skills apps/server/src/http/routes/skills-install.test.ts` gives 65 pass, 0 fail.

On the product question: the 409 is not scoped to an org. Same URL, two orgs:

```
org_a -> 201
org_b -> 409 Skill "github-weather" already exists at a different source path...
```

`getSkillByName` has no org filter (`packages/db/src/types.ts:667`), so the first org to install `code-review` takes that name from every other org on the instance.

That rules out option 2, which would point profile B at another tenant's directory. Option 3 is what the schema already allows: `migrate.ts:124` drops `skills_name_unique` and makes `source_path` unique, so two copies of one name at different paths are legal, and only the name guard in `createAndAssignRawSkillToProfile` refuses them.

I would ship this on option 1 and do option 3 separately, since the fix sits in the shared agent write path rather than in install.

`createdBy`: "human". The skill page renders this as an "Agent" / "Human" / "Bundled" label, and a dashboard install currently reads "Agent". `upsertDiscoveredSkill` already takes an override, so it is one optional arg.

@ahmadrosid the 1-vs-3 call is yours.

ahmadrosid:
Lets just do options 1, we'll keep it simple.

akmalsyrf:
Following up on the product call and review notes:

- **Option 1 (keep 409)** — kept as decided (@ahmadrosid). Same-name install onto another profile/org stays a conflict; use **Add skills** to assign an existing skill.
- **`createdBy: "human"`** — done in `4d283411`. Dashboard GitHub install now records human authorship (skill page shows Human, not Agent). Agent `skill_manage` create path still defaults to `agent`.

Verified: `bun test apps/server/src/http/routes/skills-install.test.ts` — 6 pass, including `createdBy === "human"` on a successful install.

ahmadrosid:
Nice thanks.

PR #285, opened 2026-08-16
fix(web): raise light-mode primary and muted-foreground to AA

### LINKED ISSUE #277: Two token-level contrast failures: primary buttons at 3.19:1, muted-on-muted at 4.35:1 in light
Two contrast failures that live in the tokens rather than in any one component, so they want a colour decision rather than a patch. Both measured by compositing the resolved token over each surface in the browser and reading the pixels back, so no oklch is parsed by hand.

**Primary buttons are 3.19:1.** White text on `--primary` at 12.8px, on `Save`, `Create tool`, `Create automation`, `Download`, `Choose backup file` and `New chat`. A sweep of every rendered text element on `/settings`, `/system`, `/profiles`, `/automations` and `/history` finds these are 8 of the 9 remaining sub-AA elements in light mode. Dark mode is clean. AA wants 4.5:1 at that size.

**`text-muted-foreground` over `bg-muted` is 4.35:1 in light**, the only surface pairing that misses:

| muted-foreground over | light | dark |
| --- | --- | --- |
| background | 4.74 | 7.72 |
| card | 4.74 | 6.91 |
| sidebar | 4.54 | 7.36 |
| muted | 4.35 | 6.41 |

No PR for either. The first moves the brand accent, the second moves every muted string in the app, and neither is a drive-by.

#276 covers the part that needed no token change: the sixteen `text-muted-foreground/50`, `/60` and `/70` uses, which measured 1.90 to 4.26 and now use the plain token.

fajarhide:
White text on `--primary` measures 3.19:1 in light mode, and `text-muted-foreground` over `bg-muted` measures 4.35:1. AA wants 4.5:1 at the sizes both run at.

`--primary` goes from amber-600 `#d97706` to amber-700 `#b45309`, and `--ring`, `--sidebar-primary` and `--sidebar-ring` follow so the accent stays one colour. Prose links use `--primary` as text on white and were failing at the same 3.19:1, so they clear AA too. `--muted-foreground` goes from `oklch(0.556 0 0)` to `oklch(0.54 0 0)`. Every one of those lives in `:root`, so dark mode is untouched.

That left one element still short: the `N tools` subtitle on the selected profile row, muted text over a `bg-primary/10` tint, 4.37:1. The second commit drops that tint to `/5` in light and keeps `/10` in dark, which is what `StatusPage.tsx:598` already does.

| light mode | before | after |
| --- | --- | --- |
| white text on `--primary` | 3.19 | 5.02 |
| `--primary` as link text on `--background` | 3.19 | 5.02 |
| muted-foreground on `--background` and `--card` | 4.74 | 5.04 |
| muted-foreground on `--sidebar` | 4.54 | 4.83 |
| muted-foreground on `--muted` | 4.35 | 4.62 |
| `N tools` on the selected profile row | 4.25 | 4.68 |

Sweeping every rendered text element on 8 pages: light mode had **11 sub-AA before, 1 after the token change, 0 now**. Dark mode measured **0 at every step**. `--sidebar-accent-foreground` on `--sidebar-accent`, which the issue did not cover, is 10.37 in light and 15.74 in dark.

Measured in headless Chrome by compositing each element's ancestor background stack onto a canvas and reading the pixels back, theme forced through `localStorage` before load. Same instance, same 1440x900 viewport for every capture.

| before | after |
| --- | --- |
|  |  |
|  |  |
|  |  |

The selected-row tint is visibly paler as a result. Happy to keep `/10` and darken the subtitle to `text-foreground/70` instead (5.37:1) if the stronger tint matters more.

`bun run test` passes across all eight workspaces.

Closes #277


ahmadrosid:
love it

PR #291, opened 2026-08-17
fix(db): restore install-wide tool and MCP name uniqueness

### LINKED ISSUE #289: fix: tenant migration silently dropped tool and MCP name uniqueness
## Summary

`migrateTenantOrgScope` drops `tools_name_unique` and `mcp_servers_name_unique` and replaces them with `(org_id, name)` indexes, but no write path sets `org_id` on either table. SQLite treats every `(NULL, name)` pair as distinct, so both replacements are inert and the tables lost the uniqueness they used to have.

**Goal: tool and MCP server names are unique again at the database level, matching what `schema.sql` already declares.**

Found while fixing #287, which is the same defect on `skills`. This one is latent, not active: no current code path produces a duplicate. Filing it because the schema and the migration now disagree, and the next feature that renames or imports a record inherits the gap.

## The contradiction

`schema.sql` still declares both indexes, and `migrateMcpTables` recreates one of them on every boot:

| Where | What it does |
|---|---|
| `packages/db/sql/schema.sql:30` | `CREATE UNIQUE INDEX tools_name_unique ON tools (name)` |
| `packages/db/sql/schema.sql:187` | `CREATE UNIQUE INDEX mcp_servers_name_unique ON mcp_servers (name)` |
| `packages/db/src/migrate.ts:100` | recreates `mcp_servers_name_unique` |
| `packages/db/src/migrate.ts:645,648` | drops both |
| `packages/db/src/migrate.ts:646,649` | creates `(org_id, name)` in their place |

`migrateTenantOrgScope` runs last, so the drop wins.

## Reproduction

```ts
import { Database } from "bun:sqlite";
import { migrateDatabase } from "./packages/db/src/migrate";

const db = new Database(":memory:");
migrateDatabase(db);

for (const table of ["tools", "mcp_servers"]) {
  console.log(table, "->", db
    .query("SELECT name FROM sqlite_master WHERE type='index' AND tbl_name=? AND name NOT LIKE 'sqlite_%'")
    .all(table));
}

const ins = db.prepare(
  "INSERT INTO tools (id, name, description, handler_type, handler_config, created_at, updated_at) VALUES (?, 'bash', '', 'bash', '{}', '2026-08-17', '2026-08-17')"
);
ins.run("tool_1");
ins.run("tool_2"); // no error
console.log(db.query("SELECT count(*) c FROM tools").get());
```

```
tools -> tools_org_name_unique
mcp_servers -> mcp_servers_org_name_unique
duplicate tool name ACCEPTED, rows = { c: 2 }
duplicate mcp name ACCEPTED, rows = { c: 2 }
```

## Why it is latent, not active

Both create paths check the name first (`profile-service.ts:416`, `mcp-service.ts:63`), `updateServer` checks a rename (`mcp-service.ts:105`), and tools have no rename endpoint at all. So today the application layer is the only thing holding the invariant, and it happens to hold it.

It stops holding the moment anything writes these tables outside those two guards: a rename endpoint for tools, a bulk import, a restore that replays rows, or a second concurrent create.

## Proposed direction

Restore the guarantee where these objects actually live. `POST /v1/tools` and `POST /v1/mcp/servers` are `requirePlatformAdminFromContext`, and neither `createTool` nor `createServer` takes an `orgId`, so both are install-wide objects today:

```sql
CREATE UNIQUE INDEX IF NOT EXISTS tools_global_name_unique
  ON tools (name) WHERE org_id IS NULL;
CREATE UNIQUE INDEX IF NOT EXISTS mcp_servers_global_name_unique
  ON mcp_servers (name) WHERE org_id IS NULL;
```

That is two statements in `migrateTenantOrgScope`, and it keeps the `(org_id, name)` indexes working for the day these tables do become tenant-scoped. Wrap them so a database that already carries duplicates cannot fail the boot migration, the same way #288 handles the skills backfill.

## Not proposed

Making tools and MCP servers org-scoped. That means threading an org through both services, both route groups and the write paths, and nothing today asks for it: the routes are platform-admin by design. It is a product decision, not a fix for this defect. If you want it, it deserves its own issue and this one closes as part of it.

## Acceptance criteria

- [ ] Two tools cannot share a name at the database level
- [ ] Two MCP servers cannot share a name at the database level
- [ ] `schema.sql` and the migration agree on what is unique
- [ ] An existing database carrying duplicates still boots
- [ ] A migration test covers all four

## Environment

Bun 1.3.14, macOS, sqlite adapter, `main` at `41279809`.

fajarhide:
Closes #289, taking the direction you picked.

```sql
CREATE UNIQUE INDEX IF NOT EXISTS tools_global_name_unique
  ON tools (name) WHERE org_id IS NULL;
CREATE UNIQUE INDEX IF NOT EXISTS mcp_servers_global_name_unique
  ON mcp_servers (name) WHERE org_id IS NULL;
```

The `(org_id, name)` indexes stay exactly as they are, so nothing changes the day these tables do get an owning org. Both statements are created inside a `try`: a database that already collected duplicates keeps booting, and the create paths hold the invariant from there.

## Before

```
$ bun test packages/db/src/migrate.test.ts -t "install-wide"
(fail) install-wide name uniqueness > rejects a second tool or MCP server with the same name
 1 pass
 1 fail
```

Two rows named `bash` were accepted, because `tools_org_name_unique` compares NULL to NULL.

## After

```
$ bun test packages/db/src/migrate.test.ts
 16 pass
 0 fail
```

Two tests: duplicate tool and MCP names are rejected, and a database holding duplicates still migrates without throwing. The first fails if the call is removed.

`bun run test` is green across all 11 workspaces.

Note on ordering: #288 touches the same function for `skills`. Whichever lands first, the other rebases cleanly, they add different statements.


fajarhide:
@ahmadrosid this is the first option from #289, two partial indexes on `WHERE org_id IS NULL`. The `(org_id, name)` indexes are untouched, so nothing has to change the day these tables get an owning org. Both statements are wrapped so an install that already collected duplicates still boots.

PR #300, opened 2026-08-19
feat(agent): wait out repeated in-flight tool polls in the tool loop

ahmadrosid:
## Summary

An in-flight MCP job (Digest Instagram ingest in session `webKKzAzLjA8gSwm9QsCf`) used to cost a full LLM turn per `job_status` poll — 41 rounds for 13 progress states. After the **third** same `name`+args call in a turn, the tool loop now retries that tool with backoff until the payload **changes**, the snapshot is terminal, the turn is aborted, or 2 minutes elapse.

This is not Digest-specific. Any sequential tool (MCP, builtin, custom JS) that returns an in-flight `status`/`state`/`progress` snapshot can latch. Create/send acks (`{ok: true}`, `{job_id}` only) and `{error}` results are never retried, so a repeated `email_send` cannot loop-send.

The model still sees each changed snapshot (OCR 5% → 20% → 27%…), not a single wait-until-succeeded result. Kickoff tools are unchanged.

## Test plan

- [x] `bun test packages/agent` — 102 pass, including poll-wait unit tests and a chat-loop integration that collapses the third status poll
- [ ] Re-run a Digest `instagram_save_url` flow and confirm the UI shows a handful of status cards, not dozens, while OCR is running

### DIFF
--- a/packages/agent/src/chat-tool-loop.test.ts
+++ b/packages/agent/src/chat-tool-loop.test.ts
@@ -443,4 +443,87 @@ describe("agent chat tool loop", () => {
 
     expect(systems[0]).toContain("[pending] Ship");
   });
+
+  test("collapses repeated in-flight status polls after the third call", async () => {
+    let runs = 0;
+    const statusTool: ToolDefinition = {
+      description: "Poll a job",
+      name: "job_status",
+      parameters: {
+        properties: { job_id: { type: "number" } },
+        required: ["job_id"],
+        type: "object",
+      },
+      run() {
+        runs += 1;
+        if (runs < 4) {
+          return Promise.resolve({
+            progress: { percent: 5 },
+            status: "running",
+          });
+        }
+        return Promise.resolve({
+          progress: { percent: 20 },
+          status: "running",
+        });
+      },
+    };
+
+    const statusCall = (id: string) => ({
+      arguments: { job_id: 17 },
+      id,
+      name: "job_status",
+    });
+
+    const provider = createMockProvider([
+      {
+        assistantMessage: {
+          content: "",
+          role: "assistant",
+          toolCalls: [statusCall("call_1")],
+        },
+        content: "",
+        toolCalls: [statusCall("call_1")],
+      },
+      {
+        assistantMessage: {
+          content: "",
+          role: "assistant",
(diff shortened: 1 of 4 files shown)


fajarhide:
@ahmadrosid the per-turn state, the MCP envelope unwrapping and the stable serialization all look right, and the abort path is covered. One thing worries me: what qualifies as pollable.

`canPollWaitOnResult` latches on any record whose `status` / `state` / `job_status` / `task_status` is a string that is **not** in `TERMINAL_STATUSES`. That is a default-open rule, so it catches more than job snapshots. Two cases I ran against this branch:

```ts
canPollWaitOnResult({ message_id: "m1", status: "sent" })   // true
canPollWaitOnResult({ message_id: "m1", status: "queued" }) // true
canPollWaitOnResult({ id: 13, state: "open" })              // true
```

The send case is the one that bites. `never retries a mutating ack even on the third identical call` proves it for `{ ok: true, message_id }`, but a tool that answers `{ status: "sent", message_id }` takes the other branch. On this branch, with a fake clock and `timeoutMs: 20`, the third identical `email_send` ran the tool 4 more times. Under the real 120s budget with 500ms doubling to 5s, that is roughly 25 re-sends before the timeout returns.

`{ state: "open" }` is the quieter version: a read tool now blocks the turn for two minutes waiting for an issue to change state.

Two ways to close it. An explicit opt-in on the tool definition (`pollable: true`, or a name convention like `*_status`) makes the rule default-closed, which is what a mechanism that re-invokes arbitrary tools wants. Cheaper: invert the status test to a known in-flight set (`running`, `in_progress`, `processing`, `pending`, `started`, `working`) plus the `progress.percent < 100` branch you already have. That keeps the Digest case and drops the rest.

Smaller: during the wait the stream carries nothing but the 4s keepalive. With `NAKAMA_CHAT_STREAM_TIMEOUT_MS` at its 60s minimum, a 120s wait outlives the whole stream budget. Capping `timeoutMs` against the resolved stream timeout, or emitting a progress event per poll, would keep those two in step.

ahmadrosid:
Closing for now — we don't need the tool-poll wait behavior yet. Can reopen later if we pick this up again.

PR #304, opened 2026-08-19
fix(server): scope by-id session routes to the caller's org

fajarhide:
The `sessions` table carries no org column. The by-id handlers passed the raw
`sessionId` to `AgentService`, which resolved the org from the session record
itself, so org membership was checked but session ownership never was. An
authenticated member of one org could name another org's session id and read
its history, post a turn into it (running under that org's profile, tools and
credentials), compact it, branch it, or delete it.

`AgentService` now resolves scope through `profiles.org_id`, the way
`getTaskChatMessages` already did, and `orgId` is a required argument on
`getSessionMessages`, `getSessionTodos`, `getSessionQuestionnaire`,
`resolveSession`, `branchSession`, `clearSession`, `purgeSession` and
`compactSession`, so a new call site cannot drop it. The check runs before the
in-memory session cache, which returned a built session without a database
read.

`apps/server/src/http/routes/sessions.org-scope.test.ts` covers messages,
status, stream, send, delete, purge, compact and branch across two orgs.

Before (without the fix): 8 of 9 cases fail, cross-org read returns 200 and
branch returns 201.
After: 9 of 9 pass, and `bun run test` is green across all 11 workspaces.


ahmadrosid:
Already landed on main (076a8da6 + 6610891b) — HTTP by-id session routes are org-scoped and `sessions.org-scope.test.ts` covers the cross-org cases.

Closing this PR because the branch conflicts and the history is 27 merge/sync commits. Thanks for the write-up and the test shape.

Follow-up (separate PR from current main, if we want it): make `orgId` required on the AgentService by-id methods so a new call site cannot skip the check.

ahmadrosid:
Follow-up is up: https://github.com/ahmadrosid/nakama/pull/321

That PR makes `orgId` required on the AgentService by-id methods so a new call site cannot skip the check, passes the task org into `resolveSession`, and removes the unused cache-only `deleteSession`.

PR #310, opened 2026-08-19
perf(web): lazy-load route pages

### LINKED ISSUE #309: Reduce web first-load JavaScript with route-level code splitting
## Problem

`apps/web/src/App.tsx` statically imports every page, so lightweight entry points such as login, setup, and public artifact shares download the chat rendering stack, Mermaid, Cytoscape, settings, and admin pages before those features are visited.

Baseline measured at `041942ad` with:

```sh
cd apps/web
bun x tsc --noEmit
bun x vite build --manifest
```

The single application entry was **3,166.09 kB minified / 904.71 kB gzip**.

## Proposed change

Use `React.lazy` at the existing route boundary and a `Suspense` fallback using the existing full-screen spinner. Keep providers, guards, and the shared layout eager.

This should stay focused on route-level code splitting. Deep `hugeicons-react` imports and other barrel-import rewrites are intentionally excluded: isolated production probes produced the same 4,055-byte gzip output as root imports, so they do not reduce shipped bytes here. `manualChunks` is also excluded because it would rearrange, rather than defer, eagerly reachable modules.

## Local experiment

| Payload | Before gzip | After gzip | Reduction |
| --- | ---: | ---: | ---: |
| Initial application entry | 904.71 kB | 238.05 kB | 73.7% |
| Login route total | ~899 kB | 238.59 kB | 73.5% |
| Setup route total | ~899 kB | 273.95 kB | 69.5% |
| Profiles route total | ~899 kB | 286.06 kB | 68.2% |
| Chat route total | ~899 kB | 767.26 kB | 14.7% |
| Automations route total | ~899 kB | 793.13 kB | 11.8% |

Route totals include the application entry plus each route's recursively imported static JS chunks. Aggregate generated JS increases by 14,091 bytes (0.095%) from chunk wrappers, so this is specifically a first-load improvement rather than an aggregate artifact-size reduction.

## Acceptance criteria

- Each page is emitted as a dynamic entry in Vite's manifest.
- Login, setup, public-share, and authenticated route behavior is unchanged.
- A consistent loading state is shown while a route chunk loads.
- Web type checking, production build, tests, and React Doctor continue to pass without new findings.

mamenesia:
## Summary

- lazy-load every web page at its existing React Router boundary
- keep the authenticated layout visible while its outlet loads; retain the full-screen fallback for `/login`, `/setup`, and `/s/:token`
- show a reload prompt when a stale deployment chunk cannot be loaded
- keep providers, guards, and the shared layout eager
- use one typed helper for named page exports

Closes #309.

## Bundle results

Measured from production builds at `041942ad` using `bun x vite build --manifest`:

| Payload | Before | After |
| --- | ---: | ---: |
| Initial entry, minified | 3,166.09 kB | 768.21 kB |
| Initial entry, gzip | 904.71 kB | 238.14 kB |
| Login route total, gzip | ~899 kB | 239.31 kB |
| Setup route total, gzip | ~899 kB | 274.70 kB |
| Profiles route total, gzip | ~899 kB | 286.90 kB |
| Chat route total, gzip | ~899 kB | 770.76 kB |

The final initial entry is 75.7% smaller minified and 73.7% smaller gzip.

Deep barrel/icon import rewrites are excluded because isolated production probes produced identical 4,055-byte gzip output for `hugeicons-react` root and deep imports.

### First chat render

I compared the current lazy behavior with a post-mount `import()` warm-up over five cold-profile runs under the same synthetic Chrome profile (1.6 Mbps, 100 ms latency, 4x CPU slowdown):

| Variant | Median first chat render |
| --- | ---: |
| Lazy on navigation | 34.79 s |
| Warm after mount | 34.81 s |

The 0.02 s difference is noise, so the final code keeps Chat lazy rather than downloading its graph for authenticated visits that do not open Chat.

## Verification

- `bun x tsc --noEmit`
- `bun x vite build --manifest`
- `bun run --filter @nakama/web test` - 275 passed
- `bun x ultracite check apps/web/src/App.tsx apps/web/src/components/Layout.tsx apps/web/src/components/RouteBoundary.tsx`
- `bun run doctor` - 90/100, with the same seven existing findings outside the changed files
- production Chrome smoke tests with the built Chat chunk deliberately unavailable: authenticated `/chat` kept the layout and showed the reload prompt; public `/login` showed the full-screen reload prompt
- all 14 dynamic page chunks are present in the final production build

## Local environment notes

The repository-wide Ultracite command reports the Windows checkout's existing CRLF files. The full test command also reaches unrelated Windows path/locking failures in core Telegram fixtures and the SQLite reopen test. The changed files, web test suite, type check, production build, and browser smoke tests pass.

### DIFF
--- a/apps/web/src/App.tsx
+++ b/apps/web/src/App.tsx
@@ -1,80 +1,65 @@
 import { QueryClientProvider } from "@tanstack/react-query";
-import { lazy, Suspense, useEffect } from "react";
+import { type ComponentType, lazy, useEffect } from "react";
 import { Navigate, Route, Routes } from "react-router-dom";
 import { AuthGuard } from "@/components/AuthGuard";
 import { Layout } from "@/components/Layout";
 import { PlatformAdminGuard } from "@/components/PlatformAdminGuard";
+import { RouteBoundary } from "@/components/RouteBoundary";
 import { SetupGuard } from "@/components/SetupGuard";
-import { Spinner } from "@/components/ui/spinner";
 import { AppProvider } from "@/context/app-context";
 import { AuthProvider } from "@/context/auth-context";
 import { AppQueryPrefetch } from "@/hooks/use-app-queries";
 import { statusTabPath } from "@/lib/navigation";
 import { onGlobalQueryError, queryClient } from "@/lib/query-client";
 
-const AutomationsPage = lazy(() =>
-  import("@/pages/AutomationsPage").then(({ AutomationsPage }) => ({
-    default: AutomationsPage,
-  }))
+const lazyPage = <Name extends string>(
+  load: () => Promise<Record<Name, ComponentType>>,
+  name: Name
+) => lazy(async () => ({ default: (await load())[name] }));
+
+const AutomationsPage = lazyPage(
+  () => import("@/pages/AutomationsPage"),
+  "AutomationsPage"
 );
-const ChatPage = lazy(() =>
-  import("@/pages/ChatPage").then(({ ChatPage }) => ({ default: ChatPage }))
+const ChatPage = lazyPage(() => import("@/pages/ChatPage"), "ChatPage");
+const FilesPage = lazyPage(() => import("@/pages/FilesPage"), "FilesPage");
+const HistoryPage = lazyPage(
+  () => import("@/pages/HistoryPage"),
+  "HistoryPage"
 );
-const FilesPage = lazy(() =>
-  import("@/pages/FilesPage").then(({ FilesPage }) => ({ default: FilesPage }))
+const IntegrationsPage = lazyPage(
+  () => import("@/pages/IntegrationsPage"),
+  "IntegrationsPage"
 );
-const HistoryPage = lazy(() =>
-  import("@/pages/HistoryPage").then(({ HistoryPage }) => ({
-    default: HistoryPage,
-  }))
+const LoginPage = lazyPage(() => import("@/pages/LoginPage"), "LoginPage");
+const NotificationsPage = lazyPage(
+  () => import("@/pages/NotificationsPage"),
+  "NotificationsPage"
 );
-const IntegrationsPage = lazy(() =>
-  import("@/pages/IntegrationsPage").then(({ IntegrationsPage }) => ({
-    default: IntegrationsPage,
-  }))
+const ProfilesPage = lazyPage(
+  () => import("@/pages/ProfilesPage"),
+  "ProfilesPage"
 );
-const LoginPage = lazy(() =>
-  import("@/pages/LoginPage").then(({ LoginPage }) => ({ default: LoginPage }))
+const PublicArtifactSharePage = lazyPage(
+  () => import("@/pages/PublicArtifactSharePage"),
+  "PublicArtifactSharePage"
 );
-const NotificationsPage = lazy(() =>
-  import("@/pages/NotificationsPage").then(({ NotificationsPage }) => ({
-    default: NotificationsPage,
-  }))
+const SettingsPage = lazyPage(
+  () => import("@/pages/SettingsPage"),
+  "SettingsPage"
 );
-const ProfilesPage = lazy(() =>
-  import("@/pages/ProfilesPage").then(({ ProfilesPage }) => ({
-    default: ProfilesPage,
-  }))
+const SetupWizardPage = lazyPage(
+  () => import("@/pages/SetupWizardPage"),
+  "SetupWizardPage"
 );
-const PublicArtifactSharePage = lazy(() =>
-  import("@/pages/PublicArtifactSharePage").then(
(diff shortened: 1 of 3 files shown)


fajarhide:
@mamenesia 904 kB to 238 kB gzip on the entry is a real win, and measuring it with `--manifest` before and after is the right kind of evidence.

Two things about where the boundary sits.

`<Suspense>` wraps `<Routes>`, so it is outside `Layout`. The first visit to any route replaces the whole shell with the full-screen spinner, sidebar included, and it comes back when the chunk lands. Putting the boundary inside `Layout`, around the `<Outlet />`, keeps the shell up and spins only the content area. `/login`, `/setup` and `/s/:token` render outside the layout, so the full-screen fallback is right for those.

There is no error boundary anywhere in `apps/web`. A chunk that fails to load renders nothing now, and that failure mode did not exist while the pages were in the entry bundle. It shows up when a deploy replaces the hashed files under an open tab, which is the common case rather than a rare one. A small boundary around the Suspense with a reload prompt covers it.

One more to consider: `/` redirects to `/chat`, so the first paint after login now waits on a second request for the chat chunk. Keeping `ChatPage` eager, or warming it with an `import()` after mount, is worth measuring against time to first chat render rather than entry size.

Minor: the fourteen `.then(({ X }) => ({ default: X }))` wrappers could be one small helper, or a default export per page.

mamenesia:
Thanks @fajarhide - addressed in `5ee89fcd`.

- The authenticated boundary now wraps `<Outlet />` inside `Layout`, so the shell/sidebar stay mounted. `/login`, `/setup`, and `/s/:token` each retain a full-screen boundary.
- `RouteBoundary` now combines Suspense with an error boundary and reload prompt. I tested the deploy-stale case by making the built Chat and Login chunks unavailable: `/chat` kept the authenticated shell around the prompt, while `/login` showed only the full-screen prompt.
- I replaced the named-export adapters with one typed `lazyPage` helper.

For Chat, I compared lazy loading with a post-mount `import()` warm-up over five cold-profile runs under the same constrained Chrome profile (1.6 Mbps, 100 ms latency, 4x CPU slowdown):

| Variant | Median first chat render |
| --- | ---: |
| Lazy on navigation | 34.79 s |
| Warm after mount | 34.81 s |

The 0.02 s difference was noise, so I kept Chat lazy rather than fetching its dependency graph on authenticated visits that do not open Chat. The final entry remains 768.21 kB / 238.14 kB gzip.

Rechecked the production build, focused Ultracite, all 275 web tests, and React Doctor (90/100; same seven unrelated existing warnings).

fajarhide:
@mamenesia measuring the warm-up instead of arguing about it is the part I want to copy: 34.79 s against 34.81 s over five cold runs settles it, and keeping Chat lazy is the right read of that.

I built `5ee89fcd` here. The entry chunk comes out at 768,214 bytes, which is your 768.21 kB to the byte, 275 web tests pass, and `bun run --filter @nakama/web build` runs `tsc --noEmit` first, so the web side is type-clean. `RouteBoundary` wrapping Suspense in an error boundary is the shape I was hoping for, and the boundary now sits around `<Outlet />` inside `Layout` with the full-screen variant left on `/login`, `/setup` and `/s/:token`.

One thing the new placement brings with it: `key={location.pathname}` on that boundary. A changed key unmounts the subtree, so every path change remounts whatever `Outlet` renders. `syncChatUrl` in `use-chat-page.ts:181` navigates from `/chat` to `/chat/<profileId>/<sessionId>` as soon as a session exists, which happens mid-turn on the first message, so ChatPage unmounts and remounts while the reply is streaming. Its `messages`, `session`, `agentTodos` and streaming state are all `useState` in that hook.

The key is there to clear `failed` after a bad load, and that part is worth keeping. It just does not need a remount: a `resetKey` prop compared in `getDerivedStateFromProps`, or keying on the matched route pattern rather than the full pathname, clears the error state and leaves the page mounted. Worth a look at what the chat actually does after that remount before deciding how much it matters.

@ahmadrosid no check runs exist for this head, same as #312. Both are sitting at UNSTABLE, so the workflow probably needs your approval before either can show green.

ahmadrosid:
Thanks @mamenesia — this is a strong first contribution. The 904 kB → 238 kB gzip drop on the entry is a real win, and the follow-up on the layout boundary, stale-chunk reload, and Chat warm-up measurement is exactly the kind of evidence I want on perf PRs.

Merging this as-is. I'll follow up on Fajar's remaining note: `key={location.pathname}` remounts Chat when `syncChatUrl` writes `/chat/<profileId>/<sessionId>` on the first send. I'll drop that pathname key and reset the error state without unmounting the page.

ahmadrosid:
Follow-up landed in #317: dropped `key={location.pathname}` so the first send no longer remounts Chat when `syncChatUrl` writes `/chat/<profileId>/<sessionId>`. Failed chunk loads still clear on navigation via `resetKey`.

Issue #307, opened 2026-08-19
Org isolation stops at data: provider and channel config are still one per install


fajarhide:
@ahmadrosid this is a design question rather than a bug report. It came out of the
two fixes I opened today, #304 and #305, and I could not tell from the code which
answer you want.

#57 asked for isolated data, users, agents and configurations, with tenant context
enforced in the API, the database and the UI. Data and agents got there.
Configuration did not.

Still one per install, shared by every org:

| What | Where |
|---|---|
| provider instances: endpoint, API key, default | `userConfig` on disk |
| thinking, timezone, vision, transcription, image generation | same |
| Telegram, Discord, WhatsApp and Composio credentials | `WORKSPACE_SETTINGS_ID` |
| channel workers | one process per install, not per org |
| tools and MCP servers | platform-global, already deliberate |

Per org: profiles, sessions, automations, tasks, org memory, skills.

So an admin in org B can change the model, the endpoint and the bot token that org
A runs on, and org A's traffic is billed to whichever key is configured. #305 makes
those writes require an admin, since any logged-in user could reach them, a viewer
included. It cannot check *which* org's admin, because the record carries no org.
#304 closes the data side of the same boundary.

Three ways to go, and I am happy to build whichever you pick:

1. One install per tenant is the intended model. Then the docs should say so, and
   `requirePlatformAdmin` is the more honest guard on those routes.
2. Orgs are real tenants. Provider config needs an org id, which pulls in per-org
   worker routing and per-org usage accounting.
3. A middle step first: an org-level provider override that falls back to the
   workspace config, leaving channels alone.

Happy to write up whichever you choose as a sized issue.

ahmadrosid:
Rechecked against current main. The underlying concern remains valid, but this issue's scope is outdated: Telegram configuration is now org-scoped. Providers, Discord, WhatsApp, and the Composio API key remain shared, and an org admin can still update shared provider configuration.

Created #967 with the updated scope, code references, and the remaining design decision. Settings permission tests passed (47/47).

Closing this issue as superseded by #967, not as fixed.

PR #311, opened 2026-08-19
feat(coding-agent): optional harness-native vendor login

### LINKED ISSUE #143: Coding agents: optional harness-native vendor login with better UX
## Summary

Coding agents currently **always** use Nakama provider passthrough — credentials from Settings → Provider are injected at spawn time. Harness-native vendor login (`claude /login`, `codex login`, `opencode auth login`) is **not supported**.

This issue tracks adding an opt-in path for operators who want coding CLIs to use their own vendor accounts on the host instead of the Nakama-configured provider.

## Motivation

Some deployments may want:

- Separate billing/accounts for coding CLIs vs chat
- To avoid injecting org API keys into subprocess env on shared hosts
- Debugging harness auth independently of Nakama routing

We previously sketched a `coding_agent_provider_passthrough` workspace toggle but removed it to keep v1 simple (passthrough-only).

## Proposed UX (draft)

1. **Integrations → Coding agents** — clear toggle: “Use Nakama provider” (default on) vs “Use harness login on this server”
2. When harness-native mode is selected:
   - Show per-harness login instructions (`codex login`, etc.)
   - Readiness probe treats vendor auth as success (not Settings → Provider errors)
   - Hide/conflict provider passthrough status messaging
3. Document security tradeoffs (credential blast radius vs separate accounts)

## Technical notes

- DB column `coding_agent_provider_passthrough` was removed from schema; re-add if needed with migration
- API field `providerPassthroughEnabled` on `PUT /v1/settings/coding-harnesses` was removed from contract
- Probe logic now always returns `retry` + Settings → Provider guidance on auth failure

## Acceptance criteria

- [ ] UI toggle with clear copy (not API-only)
- [ ] When off: no spawn env injection; probes accept vendor login
- [ ] When on: current passthrough behavior (default)
- [ ] Docs updated with when to use each mode
- [ ] Tests for both paths

## Related

Provider passthrough refactor (replaces inference gateway): coding agents use Settings → Provider at spawn time.

akmalsyrf:
## Summary

Closes #143.

Coding agents currently always inject Nakama provider credentials at spawn. This PR restores an opt-in path for operators who want Codex / Claude Code / OpenCode / pi to use **harness-native vendor login on the host** instead.

Default remains **Use Nakama provider** (passthrough on). Turning it off:

- does not merge `ANTHROPIC_*` / `OPENAI_*` / temp harness config env into `bash` coding-agent runs
- treats an installed CLI as ready without requiring Settings → Provider
- on a failed auth probe, points at `codex login` / `claude auth login` / `opencode auth login` / `pi login` instead of Settings → Provider
- shows those login commands on **Integrations → Coding agents**

Cursor Agent is unchanged: it already uses host Cursor auth and never gets Nakama keys.

## Why

Some installs want separate billing for coding CLIs vs chat, or do not want org API keys in subprocess env on a shared host. That toggle existed in an earlier sketch (`coding_agent_provider_passthrough`) and was removed to keep v1 passthrough-only.

## Changes

- Workspace setting `coding_agent_provider_passthrough` (SQLite default `1`)
- `GET`/`PUT /v1/settings/coding-harnesses` with `providerPassthroughEnabled` (org-admin write)
- Integrations card with a single switch; login commands only when passthrough is off
- Spawn/probe/bash enrichment honor the flag
- `mergeWorkspaceSettings` so unrelated settings writes cannot drop the flag
- Docs: [Coding agent](https://github.com/ahmadrosid/nakama/blob/feat/coding-agent-harness-login/docs/website/content/docs/coding-agent.mdx) and Integrations

## Security tradeoff

| Mode | Credential blast radius |
|---|---|
| Nakama provider (default) | Org provider API key is injected into the coding CLI process on the host |
| Harness login | Nakama keys stay out of that process; the vendor account already logged in on the host is used instead |

## Test plan

- [x] Integrations → Coding agents: toggle defaults on; turn off and confirm login commands (`codex login`, `claude auth login`, `opencode auth login`, `pi login`)
- [x] Passthrough on: coding `bash` run still gets provider env (Claude Code / Codex)
- [x] Passthrough off: coding `bash` run has no `ANTHROPIC_API_KEY` / `OPENAI_API_KEY`
- [x] Passthrough off, no Nakama provider configured: installed harness still reports ready
- [x] Cursor Agent: still no Nakama env either way
- [x] Org member cannot PUT the setting; org admin can
- [x] Vision/transcription/token-optimizer saves do not reset the toggle

## Validation

- UI: Integrations toggle + login commands (admin vs member screenshots); Cursor Agent still has no Nakama env
- `bun test packages/db/src/workspace-settings.test.ts apps/server/src/services/coding-agent-bash-env.test.ts apps/server/src/services/coding-agent-harness-service.test.ts apps/server/src/http/coding-harnesses.test.ts` — 20 pass
- Extra: `AgentService` vision/transcription saves keep `codingAgentProviderPassthrough: false`

### DIFF
--- a/apps/server/src/http/app.ts
+++ b/apps/server/src/http/app.ts
@@ -8,6 +8,7 @@ import { createOrgContextMiddleware } from "./org-middleware";
 import { registerArtifactShareRoutes } from "./routes/artifact-shares";
 import { registerAuthRoutes } from "./routes/auth";
 import { registerAutomationRoutes } from "./routes/automations";
+import { registerCodingHarnessSettingsRoutes } from "./routes/coding-harnesses";
 import {
   registerComposioOAuthRoutes,
   registerComposioRoutes,
@@ -117,6 +118,7 @@ export function createHonoApp(options: ServerOptions) {
   registerAutomationRoutes(app, options);
   registerNotificationDestinationRoutes(app, options);
   registerTokenOptimizationRoutes(app, options);
+  registerCodingHarnessSettingsRoutes(app, options);
   registerComposioRoutes(app, options);
   registerTaskRoutes(app, options);
   registerPlatformOrgRoutes(app, options);
--- a/apps/server/src/http/coding-harnesses.test.ts
+++ b/apps/server/src/http/coding-harnesses.test.ts
@@ -8,9 +8,29 @@ import { AuthService } from "../services/auth-service";
 import { OrgService } from "../services/org-service";
 import { createHonoApp } from "./app";
 import {
+  loginPlatformAdminSession,
   loginUserSession,
   setupFreshInstallSession,
 } from "./test-session-helpers";
+
+function createHarnessApp() {
+  const databaseAdapter = createInMemoryDatabaseAdapter();
+  const authService = new AuthService();
+  const app = createHonoApp({
+    agent: new AgentService(null, null, databaseAdapter),
+    authService,
+    automationService: {} as never,
+    databaseAdapter,
+    mcpService: {} as never,
+    orgService: new OrgService(databaseAdapter, authService),
+    systemStatus: { getStatus: async () => ({ ok: true }) } as never,
+    taskService: {} as never,
+    webDistDir: null,
+    workerManager: {} as never,
+  });
+
+  return { app, authService, databaseAdapter };
+}
 
 describe("coding harness settings routes", () => {
   let configDir = "";
@@ -28,20 +48,7 @@ describe("coding harness settings routes", () => {
     configDir = await mkdtemp(join(tmpdir(), "nakama-coding-harness-route-"));
     process.env.NAKAMA_CONFIG_DIR = configDir;
 
-    const databaseAdapter = createInMemoryDatabaseAdapter();
-    const authService = new AuthService();
-    const app = createHonoApp({
-      agent: new AgentService(null, null, databaseAdapter),
-      authService,
-      automationService: {} as never,
-      databaseAdapter,
-      mcpService: {} as never,
-      orgService: new OrgService(databaseAdapter, authService),
-      systemStatus: { getStatus: async () => ({ ok: true }) } as never,
-      taskService: {} as never,
-      webDistDir: null,
-      workerManager: {} as never,
-    });
+    const { app, databaseAdapter } = createHarnessApp();
 
     const session = await setupFreshInstallSession(app, databaseAdapter);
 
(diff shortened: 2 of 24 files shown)


akmalsyrf:
In Admin
Screenshot 2026-08-19 at 20 35 52

In Member
Screenshot 2026-08-19 at 20 36 14

akmalsyrf:
Screenshot 2026-08-19 at 21 04 34

fajarhide:
@akmalsyrf I went looking for the usual failure modes here and did not find them.

The column is additive with a default and guarded by the `columnNames` check, `schema.sql`, the row type, the upsert and the mapper all move together, and `mergeWorkspaceSettings` closes the hole where saving the vision model would have reset the new flag to on. The two tests for that ("does not reset coding-agent passthrough when vision is saved" and the transcription one) are the ones I would have asked for. `PUT /v1/settings/coding-harnesses` calls `requireOrgAdminFromContext` and there is a test proving a member cannot write it, which is more than several older settings routes do. The early return in `enrichCodingAgentBashInput` skips the spawn bundle entirely rather than filtering keys out of it, and `runShellCommand` still spawns with `process.env` as the base, so `PATH` and `HOME` survive for a host login to find its own credentials.

One thing worth settling before merge, and it is not really about this PR. The setting is workspace-global, the guard is org admin, so an admin in org B flips coding-agent auth for org A. Same shape as #305 and the question in #307. If the answer there ends up "global config is platform admin", this route wants `requireOrgAdminOrPlatformAdmin` or `requirePlatformAdmin` to match.

Small: `HARNESS_LOGIN_ORDER` in the route re-lists four harnesses that `DEFAULT_HARNESSES` already names, so a fifth harness means editing two places. Deriving the list and keeping only the kinds that have a login command would drop the duplicate.

@ahmadrosid this one looks ready to me apart from the guard question above.

akmalsyrf:
@fajarhide thanks — both points are in `b498f6ee`.

**Guard.** This flag is still one row for the install, so org isolation is the same open question as #307. Until that is decided, the write bar matches #305: `requireOrgAdminOrPlatformAdmin` (org admin of the active org, or a platform admin who is a member of it). A normal org member still gets 403. `requirePlatformAdmin` would hide the Integrations toggle from org admins, which is not how the other workspace-global cards work today.

**Login list.** `listCodingHarnessLoginCommands()` walks `DEFAULT_HARNESSES` and keeps kinds that have a vendor login command. Cursor Agent stays out because `getCodingHarnessLoginCommand` returns null for it; a fifth login-capable harness only needs an entry in that map plus the default catalog.

fajarhide:
@akmalsyrf both confirmed on `b498f6ee`. The route now calls `requireOrgAdminOrPlatformAdminFromContext` on the write and `listCodingHarnessLoginCommands` derives from `DEFAULT_HARNESSES`, so the duplicate list is gone and Cursor Agent drops out on its own through the null login command. Your reasoning on the guard is the one I would have picked: matching #305 keeps the toggle where the dashboard already puts it, and #307 is where the org question belongs, not here.

@ahmadrosid this one reads ready to me. Additive column with a default behind a `columnNames` check, `schema.sql`, the row type, the upsert and the mapper all moved together, `mergeWorkspaceSettings` stops an unrelated settings save from resetting the flag, and there are tests for that reset, for the member 403, and for credentials not being merged when passthrough is off. CI is green and the branch is clean against main. The screenshots above cover the admin and member views.

ahmadrosid:
Looks good. Default stays Nakama-provider passthrough, harness-native login is opt-in, and the write guard plus login-command list now match the rest of workspace-global config. The remaining org-isolation question stays on #307.

Merging.

PR #308, opened 2026-08-19
feat: add Cloudflare Workers AI as LLM provider

syawqy:
## Summary

Adds Cloudflare Workers AI as a built-in LLM provider. Closes #294

## Changes

- Added `cloudflare` to `ProviderName` type union
- Added `cloudflare` to `USER_PROVIDER_NAMES` and `parseProviderName()`
- Added `CLOUDFLARE_API_TOKEN` env var mapping
- Created `cloudflare/` provider adapter (reuses OpenAI-compatible client)
- Created `cloudflare/catalog.ts` — 6 models (Llama 3.1 8B/70B, Llama 3.3 70B, Qwen 1.5 14B)
- Registered in factory with Account ID validation
- Added 6 models to `models.ts` catalog
- Added tests (6 tests, all passing)
- Updated `providers.mdx` documentation

## Setup

1. Get Cloudflare API token from https://dash.cloudflare.com/profile/api-tokens
2. Find Account ID on Cloudflare dashboard (right sidebar)
3. Settings → LLM providers → Add Cloudflare
4. API token as API key, Account ID as base URL
5. Select model (Llama 3.1 8B is free tier)

PR #312, opened 2026-08-19
feat(server): add Cloudflare Workers AI as an LLM provider option

### LINKED ISSUE #294: feat: Cloudflare Workers AI as an LLM provider option
## Summary

Add Cloudflare **Workers AI** as a first-class (or well-documented) LLM provider option for Nakama operators who want cloud-hosted inference without wiring a Custom OpenAI-compatible endpoint by hand.

Workers AI already exposes [OpenAI-compatible chat completions](https://developers.cloudflare.com/workers-ai/configuration/open-ai-compatibility/). This issue is about product surface and operator experience — not about running the Nakama server on Cloudflare Workers.

## Motivation

Nakama is provider-agnostic: operators bring their own keys (OpenAI, Anthropic, Gemini, OpenRouter, Ollama, Custom OpenAI-compatible, etc.). A Workers AI path fits that model.

Practical reasons operators ask for this:

- **Lower cost barrier for self-host / dogfood** — Workers AI currently offers a relatively generous free inference allowance compared with many hosted LLM APIs (exact limits change; treat Cloudflare’s docs as source of truth). Useful for contributors, demos, and small deployments that cannot start on paid frontier keys.
- **Cloud-hosted alternative to local Ollama** — same “cheap/easy to try” niche, without requiring a local GPU or always-on Ollama process.
- **Less friction than Custom** — today you can point **Custom (OpenAI-compatible)** at Cloudflare’s base URL and paste an API token, but you must know the account-scoped URL, model id format (`@cf/...`), and capability flags (vision, etc.) yourself.

This should remain **opt-in**. It is not a proposal to change Nakama’s default provider or to prefer Cloudflare over existing vendors.

## What already works today (workaround)

Using **Settings → LLM providers → Custom (OpenAI-compatible)**:

1. Base URL (pattern):  
   `https://api.cloudflare.com/client/v4/accounts/{account_id}/ai/v1`
2. API key: Cloudflare API token with Workers AI permissions
3. Model id: Workers AI model slug, e.g. `@cf/meta/llama-3.1-8b-instruct` (confirm against current Cloudflare catalog)

See Cloudflare docs: [OpenAI compatible API endpoints](https://developers.cloudflare.com/workers-ai/configuration/open-ai-compatibility/).

Gaps with Custom alone:

- No preset in the provider picker / setup wizard
- No curated shortlist of models that are known to work with Nakama’s tool loop
- Operators must discover URL shape and `@cf/` model ids themselves
- Capability metadata (vision / thinking) is manual

### Phase 0 — Docs only (smallest)

Document the Custom OpenAI-compatible recipe in [Providers](https://github.com/ahmadrosid/nakama/blob/main/docs/website/content/docs/providers.mdx) (or a short self-host note):

- Base URL template + token scopes
- Example model ids
- Explicit caveats (tool calling quality varies by model; provider-specific Nakama features still require OpenAI/Anthropic where documented)

### Phase 1 — First-class provider type (optional)

If Custom proves awkward enough in practice:

- New provider type (e.g. `cloudflare_workers_ai` or similar) in Settings / setup wizard
- Preset base URL construction from account id + token
- Curated or fetchable model list with clear tool-use guidance
- Reuse existing OpenAI-compatible client path where possible (same pattern as other compat backends)

### Non-goals

- Running Nakama itself on Cloudflare Workers / Pages as the primary runtime
- Making Workers AI the default in first-time setup
- Claiming feature parity with OpenAI/Anthropic for `web_search`, transcription, or vision — those stay provider-dependent as today
- Bundling or proxying Cloudflare credentials through Nakama Cloud in a special way (still BYO token on the operator’s install)

## Constraints / risks to call out

| Area | Note |
|---|---|
| Tool calling | Smaller / open models on Workers AI may be weaker at multi-step `skill_manage` / file tools than frontier models. Curate “known good” models or document expected quality. |
| Provider-specific features | `web_search` (OpenAI/Anthropic), Telegram transcription (OpenAI), etc. will not light up on Workers AI alone — same as Ollama/Custom today. |
| Free tier | Quotas and pricing change; docs should link to Cloudflare’s current limits rather than hard-coding numbers. |
| Auth | Account id + API token; avoid storing unnecessary Cloudflare account secrets beyond what the provider needs. |

### Phase 0

- [ ] User docs explain how to use Workers AI via Custom OpenAI-compatible (URL, token, example model)
- [ ] Caveats listed (tool quality, feature gaps, quota source of truth)

### Phase 1 (if pursued)

- [ ] Workers AI appears as a selectable provider in Settings / setup
- [ ] Chat + tool loop works with at least one documented model
- [ ] Tests cover config / client wiring (compat path)
- [ ] Providers docs updated for the native type

## Test plan

- [ ] Custom workaround: configure OpenAI-compatible → Workers AI base URL → send a chat turn with tools assigned
- [ ] (Phase 1) Setup wizard / Settings can add Workers AI without hand-editing base URL
- [ ] Failure modes: bad token, wrong account id, unsupported model id → clear errors

## References

- Cloudflare Workers AI OpenAI compatibility: https://developers.cloudflare.com/workers-ai/configuration/open-ai-compatibility/
- Nakama providers docs: `docs/website/content/docs/providers.mdx`
- Existing Custom / openai-compatible provider: `apps/server/src/providers/openai-compatible/`

puspoaditya:
## Summary
Adds Cloudflare **Workers AI** as a first-class LLM provider option for operators who want cloud-hosted inference without wiring a Custom OpenAI-compatible endpoint by hand.

Closes #294

## What changed
- New `createCloudflareProvider` targeting the Workers AI OpenAI-compatible endpoint (`/client/v4/accounts/{account_id}/ai/v1`)
- Env vars: `CLOUDFLARE_API_KEY` + `CLOUDFLARE_ACCOUNT_ID`
- Built-in model catalog: Llama 3.3 70B FP8 (default), Llama 3.1 8B, Llama 4 Scout 17B, Qwen 2.5 Coder 32B, DeepSeek R1 Distill Qwen 32B
- Custom model validation with `@cf/` / `@hf/` prefix
- Provider registered in contract, resolution, labels, factory, and instance validation
- Unit tests with mocked fetch (chat completion, JSON text gen, error labeling)

## Verification
- `bun run check` — clean (1135 files)
- `bun test packages/core apps/server/src/providers` — 755 pass, 0 fail

## Notes
- Live LLM cassette test not included (needs a Cloudflare account to record one real exchange per CONTRIBUTING.md). Happy to add it if the maintainers want.


puspoaditya:
Update: added the **live LLM cassette test** (real Workers AI call recorded + replayed offline via MSW) and fixed the model catalog — `llama-3.1-8b` was deprecated upstream (2026-05-30), replaced with `@cf/meta/infire-llama-3.1-8b-instruct`. All 5 catalog models verified live against the Workers AI API. `bun run check` clean, 4 cloudflare tests + 755 core/provider tests passing.

fajarhide:
@puspoaditya the wiring here is thorough: custom model validation, `resolveModel`, `getDefaultModel` and the instance helpers are all covered, more than #308 does. Three things I would change.

**The account ID is env only.** `create.ts` reads `CLOUDFLARE_ACCOUNT_ID` and falls back to `""`. Unset, the base URL becomes `https://api.cloudflare.com/client/v4/accounts//ai/v1` and every call fails with whatever Cloudflare returns for that path, with nothing pointing at the real cause. It also makes this the one provider an operator cannot configure from the dashboard, and changing it means restarting the server. Either let them paste the full endpoint into `baseUrl`, which the existing validation accepts, or add an `accountId` field to the instance schema and the provider form. A thrown error on empty would at least fail loudly.

**376 lines that already exist.** Workers AI speaks the OpenAI API, and the repo has `createOpenAICompatibleProvider` for exactly that: `ollama/index.ts` wires a provider in 50 lines with it. Cerebras and Fireworks are the older hand-rolled shape, so following them means a second copy of the streaming, tool call and usage parsing to keep in sync.

**Two registration points are missing.** tsc catches both:

```
packages/core/src/user-config.ts(70,7): error TS2741: Property 'cloudflare' is missing ... required in type 'Record<UserProviderName, string>'.
packages/core/src/document-content.ts(59,7): error TS2741: Property 'cloudflare' is missing ... required in type 'Record<ProviderName, ReadonlySet<string>>'.
```

CI runs `bun run test`, not tsc, which is why it passed. `provider-setup-prompt.ts` is untouched too, so Cloudflare will not appear in the CLI setup list, and `providers.mdx` has no row for it.

@ahmadrosid heads up that #308 adds the same provider and closes the same issue.

puspoaditya:
Update 2: added **Cloudflare to the provider setup wizard** (CLI prompt) and the **Supported providers docs table** (with the `CLOUDFLARE_ACCOUNT_ID` env requirement). PR now covers: implementation + unit tests + live cassette test + setup prompt + docs. All green: `bun run check` clean, 614 core + cloudflare tests passing.

puspoaditya:
Thanks for the review @fajarhide — all three points addressed in the latest push (cfba831):

1. **Account ID handling**: now fails loudly with a clear error when `CLOUDFLARE_ACCOUNT_ID` is unset, AND operators can paste the full Workers AI endpoint (`https://api.cloudflare.com/client/v4/accounts/<id>/ai/v1`) into the instance base URL — dashboard-configurable, no restart needed.

2. **Thin wrapper**: rewrote the provider as a ~50-line wrapper over `createOpenAICompatibleProvider` (ollama pattern) — streaming/tool-call/usage parsing no longer duplicated. Dropped the hand-rolled 376-line version and the unused `cloudflare/models.ts`.

3. **Registrations**: added `cloudflare` to `user-config.ts` `PROVIDER_TYPE_LABELS` and `document-content.ts` `NATIVE_DOCUMENT_MEDIA_TYPES`. Also covered: CLI setup wizard (`provider-setup-prompt.ts`) and the docs table (`providers.mdx`) in the previous commit.

Verified: full `tsc --noEmit` clean, `bun run check` clean, 614 core + cloudflare tests passing (incl. the live cassette test replay).

fajarhide:
@puspoaditya checked the new head (cfba831) rather than taking it on trust, and all three hold up.

`resolveCloudflareBaseUrl` prefers the instance base URL and falls back to the env account ID, with an error that names both ways out, so the dashboard path works: for a type with no special case, `buildProviderFieldsFromRequest` accepts any valid http(s) `baseUrl`, and that is what the operator pastes. `index.ts` is 47 lines over `createOpenAICompatibleProvider` now. `tsc --noEmit` no longer reports `user-config.ts` or `document-content.ts`.

The only cloudflare lines left in a full tsc run are `RequestInfo` in `index.test.ts`, and that is not yours to fix: the root config is `lib: ["ESNext"]` with `types: ["bun"]`, and the gemini, fireworks and openrouter tests already use the same name. Repo-wide noise, not this PR.

One open question, not a blocker. The cassette records a plain chat completion, no tools. The agent loop leans on tool calls for anything past a single reply, so it would be worth knowing whether Workers AI returns `tool_calls` in the OpenAI shape for the default model, and what happens if it does not. A second cassette with one tool in the request would answer it once and for good.

puspoaditya:
Great question — answered it empirically. Added a second cassette (6856f21): `cloudflare-llama-3-3-70b-tool-call` records a live chat completion **with one tool** (`get_weather`) against the default model.

**Result: Workers AI returns OpenAI-shaped tool_calls.** The recorded response has `finish_reason: tool_calls` with:

```json
{"id": "call_1787153123291", "type": "function", "function": {"name": "get_weather", "arguments": "{\"city\":\"Jakarta\"}"}}
```

So the agent loop's tool round-trips work out of the box with this provider. Both cassettes replay cleanly offline (2/2 pass), `bun run check` green.

(One note: during re-recording I hit a couple of transient 401s from the Workers AI endpoint — looked like free-tier rate limiting, not an auth/config issue. Existing cassettes are committed so CI replay is unaffected.)

fajarhide:
@puspoaditya answering it with a recorded call rather than a paragraph is the right instinct, and the cassette settles it: `finish_reason: tool_calls` with an OpenAI-shaped `function.arguments` string, so the agent loop's round trip works on the default model.

Checked out `6856f217` and ran it here:

- `bun test apps/server/src/providers packages/core/src`: 757 pass, 0 fail
- `bun test apps/server/src/providers/cloudflare`: 18 pass, both cassettes replaying
- `tsc --noEmit`: nothing new. The `provider-instance-helpers.ts` lines about `label` are already on main at 191, 309, 340 and 347, and `RequestInfo` is the pattern the gemini, fireworks and openrouter tests use. Neither is yours.

Registration is complete now: contract union, `provider-resolution` (names, parser, env var), `provider-label`, `provider-setup-prompt`, `user-config`, `document-content`, `AVAILABLE_MODELS`, `provider-instance-helpers`, the factory, the adapter, `providers.mdx`, and two cassettes.

One thing before merge: no checks have run on `6856f217`. The PR sits at UNSTABLE and GitHub reports no check runs for that head, so the workflow probably needs a maintainer approval.

@ahmadrosid this reads ready to me once CI runs on the current head. The one call left is #308, which implements the same feature and closed the same issue: this branch now also carries the thin adapter and the `user-config` row that #308 had, so it covers both, and @syawqy's work is worth a `Co-authored-by` if you take this one.

apps/server/src/providers/create.ts
```diff
@@ -89,6 +90,13 @@ function createProvider(options: CreateProviderOptions): ProviderClient {
         customModels: options.instance?.customModels,
         model,
       });
+    case "cloudflare":
+      return createCloudflareProvider({
+        accountId: readEnvValue(process.env, "CLOUDFLARE_ACCOUNT_ID") ?? "",
```
ahmadrosid, line 95: Please don't make `CLOUDFLARE_ACCOUNT_ID` a special required process env.

The existing pattern is:
1. Persist on the provider instance in `~/.nakama/config.ini` (`api_key`, `base_url`)
2. Fall back to env only when the instance field is empty (`readApiKeyForInstance` / `apiKeyEnvVarForProvider`)

`CLOUDFLARE_API_KEY` already follows that. Account ID does not — CLI setup only asks for the API key, so a freshly saved `[provider.*]` section has no `base_url` and every call depends on a server-process env var.

Please do this instead:
- CLI / dashboard: collect the account ID (or the full Workers AI URL) and write it as `base_url=https://api.cloudflare.com/client/v4/accounts/<id>/ai/v1` on the instance
- Keep `CLOUDFLARE_ACCOUNT_ID` as a fallback only when `instance.baseUrl` is empty (same shape as the API-key fallback)
- Drop the docs line that presents `CLOUDFLARE_ACCOUNT_ID` as a required server env

That keeps Cloudflare configurable from the dashboard without a restart, and matches OpenAI / Ollama / custom.

ahmadrosid (commented):
One consistency ask before merge: treat Cloudflare credentials like every other provider — `~/.nakama/config.ini` first, process env only as fallback.

ahmadrosid:
Merged as-is. The `config.ini` consistency change (account ID saved as `base_url`, `CLOUDFLARE_ACCOUNT_ID` as fallback only) is in a follow-up: https://github.com/ahmadrosid/nakama/pull/318

apps/server/src/providers/create.ts
ahmadrosid: Not fixing this on #312 — merged as-is. Follow-up: https://github.com/ahmadrosid/nakama/pull/318

PR #330, opened 2026-08-20
fix(server): bound the agent-browser version probe with a timeout

### LINKED ISSUE #326: agent-browser version probe has no timeout, so GET /v1/settings/agent-browser never returns when the CLI hangs
#261 fixed this class for the coding agent harness probe, which now kills the child after 5s (`coding-agent-harness-service.ts:714`, `:741`). The agent-browser sibling was never covered.

`probeAgentBrowserVersion` (`agent-browser-service.ts:61-97`) resolves only on `close` or `error`, with no timer. `getAgentBrowserRuntimeStatus` runs it twice, once directly and once more after `ensureProcessPath()`. `GET /v1/settings/agent-browser` (`models.ts:1347`) awaits that, so an `agent-browser` binary that hangs on `--version` leaves the Settings request open forever.

Repro against a stub that hangs:

```sh
mkdir -p /tmp/stubbin
printf '#!/bin/sh\nsleep 120\n' > /tmp/stubbin/agent-browser
chmod +x /tmp/stubbin/agent-browser

cat > probe-repro.ts <<'TS'
import { getAgentBrowserStatus } from "./apps/server/src/services/agent-browser-service";
const started = Date.now();
const result = await Promise.race([
  getAgentBrowserStatus().then(() => "probe resolved"),
  new Promise((r) => setTimeout(() => r("still hanging after 5000ms"), 5000)),
]);
console.log(`${result} (elapsed ${Date.now() - started}ms)`);
process.exit(0);
TS

NAKAMA_DISABLE_FIX_PATH=1 PATH="/tmp/stubbin:$PATH" bun run probe-repro.ts
```

```
still hanging after 5000ms (elapsed 5002ms)
```

`NAKAMA_DISABLE_FIX_PATH=1` is required because `fixPath()` replaces `PATH` with the login shell's and drops the stub. Without it the probe returns in about 1.1s having found nothing, which is the false pass.

Fix: the same 5s timer and SIGTERM the harness probe already uses, resolving as not installed when it fires.

One more site with the same shape and no timeout: `packages/core/src/omni-install.ts:204` spawns `tar` for the OMNI auto-install and resolves only on `close` or `error`.

rioaguspermana:
`GET /v1/settings/agent-browser` waits on `probeAgentBrowserVersion()`. That spawn only resolved on `close` or `error`, so a CLI that hung on `--version` left Settings open.

This copies the coding-agent harness bound: 5s then SIGTERM, resolve as not installed. Timeout sets `missing: false` so `getAgentBrowserRuntimeStatus` does not pay the 5s again on the PATH retry.

Tests use `NAKAMA_DISABLE_FIX_PATH=1` so `fixPath()` cannot hide the stub (the trap from the issue). Hang and SIGTERM-ignoring stubs are Node processes, not `sh`+`sleep`, so the signal hits the PID we spawn.

`omni-install` `tar` spawn is still unbounded; left for a follow-up as agreed on the issue.

## Validation

Added hang / stubborn / Settings GET coverage in `apps/server/src/services/agent-browser-service.test.ts`. I could not run `bun test` to completion on this Windows machine: the local `openai` install is missing files (`version.mjs`, then `resources/videos.mjs`). CI on Linux should be the check.

## Related

Fixes #326

### DIFF
--- a/apps/server/src/services/agent-browser-service.test.ts
+++ b/apps/server/src/services/agent-browser-service.test.ts
@@ -57,6 +57,29 @@ describe("agent-browser service", () => {
     expect(status.ready).toBe(true);
     expect(status.nextStep).toBeNull();
   });
+
+  test("a CLI that hangs on --version resolves as not installed", async () => {
+    await installFakeBinary(tempBinDir, "agent-browser", "hangs");
+
+    const started = Date.now();
+    const status = await getAgentBrowserStatus();
+
+    expect(status.installed).toBe(false);
+    expect(status.ready).toBe(false);
+    expect(status.nextStep).toBe("install");
+    expect(Date.now() - started).toBeLessThan(15_000);
+  }, 20_000);
+
+  test("a CLI that ignores SIGTERM still resolves the readiness probe", async () => {
+    await installFakeBinary(tempBinDir, "agent-browser", "stubborn");
+
+    const started = Date.now();
+    const status = await getAgentBrowserStatus();
+
+    expect(status.installed).toBe(false);
+    expect(status.ready).toBe(false);
+    expect(Date.now() - started).toBeLessThan(15_000);
+  }, 20_000);
 });
 
 describe("agent-browser settings routes", () => {
@@ -133,6 +156,43 @@ describe("agent-browser settings routes", () => {
     expect(body.version).toBe("agent-browser 1.0.0");
   });
 
+  test("org admin status request returns when agent-browser hangs", async () => {
+    await installFakeBinary(tempBinDir, "agent-browser", "hangs");
+
+    const databaseAdapter = createInMemoryDatabaseAdapter();
+    const authService = new AuthService();
+    const app = createHonoApp({
+      agent: new AgentService(null, null, databaseAdapter),
+      authService,
+      automationService: {} as any,
+      databaseAdapter,
+      mcpService: {} as any,
+      orgService: new OrgService(databaseAdapter, authService),
+      systemStatus: { getStatus: async () => ({ ok: true }) } as any,
+      taskService: {} as any,
+      webDistDir: null,
+      workerManager: {} as any,
+    });
+
+    const session = await setupFreshInstallSession(app, databaseAdapter);
+    const started = Date.now();
+
+    const response = await app.fetch(
+      new Request("http://localhost:4310/v1/settings/agent-browser", {
+        headers: session.headers(),
+      })
+    );
+
+    expect(response.status).toBe(200);
+    const body = (await response.json()) as {
+      installed: boolean;
+      ready: boolean;
+    };
+    expect(body.installed).toBe(false);
+    expect(body.ready).toBe(false);
+    expect(Date.now() - started).toBeLessThan(15_000);
+  }, 20_000);
+
   test("install stream emits progress events", async () => {
     await installFakeBinary(tempBinDir, "npm", "noop");
     await installFakeBinary(tempBinDir, "agent-browser", "installable");
@@ -174,7 +234,13 @@ describe("agent-browser settings routes", () => {
 async function installFakeBinary(
   binDir: string,
   name: string,
-  mode: "ready" | "login-required" | "noop" | "installable"
+  mode:
+    | "ready"
+    | "login-required"
+    | "noop"
+    | "installable"
+    | "hangs"
+    | "stubborn"
 ): Promise<void> {
   const scriptPath = join(binDir, name);
   let script = "";
@@ -211,6 +277,18 @@ if [ "$1" = "install" ]; then
   exit 0
 fi
 exit 0
+`;
+  } else if (mode === "hangs") {
+    // Direct Node process so SIGTERM hits this PID. A shell wrapper would
+    // leave `sleep` running after kill(), and PATH in these tests is only
+    // the stub dir so `sleep` would not be found anyway.
+    script = `#!${process.execPath}
+setInterval(() => {}, 1000);
+`;
+  } else if (mode === "stubborn") {
+    script = `#!${process.execPath}
+process.on("SIGTERM", () => {});
+setInterval(() => {}, 1000);
 `;
   }
 
--- a/apps/server/src/services/agent-browser-service.ts
+++ b/apps/server/src/services/agent-browser-service.ts
@@ -64,14 +64,38 @@ async function probeAgentBrowserVersion(): Promise<{
   missing: boolean;
 }> {
   const { spawn } = await import("node:child_process");
+  const timeoutMs = 5000;
 
   return new Promise((resolve) => {
-    const child = spawn(AGENT_BROWSER_COMMAND, ["--version"], {
-      env: getToolExecutionEnv(),
-      stdio: ["ignore", "pipe", "pipe"],
-    });
+    let child: ReturnType<typeof spawn>;
+
+    try {
+      child = spawn(AGENT_BROWSER_COMMAND, ["--version"], {
+        env: getToolExecutionEnv(),
+        stdio: ["ignore", "pipe", "pipe"],
+      });
+    } catch {
+      resolve({
+        installed: false,
+        missing: true,
+        version: null,
+      });
+      return;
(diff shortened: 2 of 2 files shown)


fajarhide:
I pulled the branch into a worktree and ran it. It does what the issue asked, and the tests are real: reverting `agent-browser-service.ts` alone fails all three (stubborn at 19849ms against the 15000ms bound, both `hangs` time out at 20s).

Same repro as the issue, before and after:

```
main:   still hanging after 20000ms (elapsed 20001ms)
branch: probe resolved (installed=false) (elapsed 5009ms)
```

Full `bun run test` on the branch is green apart from `coding-agent-bash-env` timing out at 5s, which is the known fresh-worktree flake and not yours. Copying `probeHarnessVersion` line for line was right, and your comment explains the two parts that look wrong out of context.

One thing I found that is not yours to fix here: we SIGTERM and return, so a CLI that ignores SIGTERM keeps running. Against a stub that traps it, the probe returned in 5006ms and the child was still alive after. That is one orphan process per Settings load while the CLI is broken. `probeHarnessVersion` (#268, mine) has the same hole, so it wants a follow-up covering both, alongside the `omni-install` tar spawn.

Two smaller ones, your call. The two probes are now identical apart from the command and the comment, and `extractVersion` was already duplicated verbatim; one `probeCliVersion(command)` would collapse them, but that touches the harness file so it is probably its own PR. And the three tests add about 15s to the server suite: the route-level hang test covers the service path too, so the service-level `hangs` one may not be earning its 5s. I would keep `stubborn`, it is the one pinning the resolve-on-timer design.

I did not measure real `agent-browser --version` latency against the 5s bound (not installed here); the harness CLIs were 0.08 to 0.87s when we picked that number. Nothing checked on Windows.

Reads ready to me, @ahmadrosid.

apps/server/src/services/agent-browser-service.ts
```diff
@@ -64,35 +64,61 @@ async function probeAgentBrowserVersion(): Promise<{
...
     let stderr = "";
 
+    // Settings awaits this probe, so a CLI that never answers --version
+    // would otherwise hold GET /v1/settings/agent-browser open forever.
+    // Resolve on the timer instead of waiting for close, because a child
+    // that ignores SIGTERM never emits one. `missing: false` keeps
+    // getAgentBrowserRuntimeStatus from paying the timeout a second time
+    // on its PATH retry.
+    const timeoutId = setTimeout(() => {
+      child.kill("SIGTERM");
```
vincentri, line 96: This inherits the gap #338 just fixed on the twin: a binary that traps SIGTERM resolves the probe at 5s but the child keeps running, and its open stdio pipes pin handles in the server process. Main's `probeHarnessVersion` now escalates via `killTimeoutId = setTimeout(() => child.kill("SIGKILL"), SIGTERM_GRACE_MS)` + `clearTimeout` in both handlers (`coding-agent-harness-service.ts:735-768`, merged as bae24849). Copying those ~4 lines here closes it. Not blocking for this issue's AC — Settings does return — but merging as-is knowingly ships the #338 bug in a sibling, so I'd add it here rather than as a follow-up.

apps/server/src/services/agent-browser-service.test.ts
```diff
@@ -57,6 +57,29 @@ describe("agent-browser service", () => {
...
+    const started = Date.now();
+    const status = await getAgentBrowserStatus();
+
+    expect(status.installed).toBe(false);
+    expect(status.ready).toBe(false);
+    expect(status.nextStep).toBe("install");
+    expect(Date.now() - started).toBeLessThan(15_000);
+  }, 20_000);
+
+  test("a CLI that ignores SIGTERM still resolves the readiness probe", async () => {
```
vincentri, line 73: This test passes today only because the probe resolves on its timer — it never asserts the stubborn child actually died, so it cannot catch the missing-SIGKILL escalation (see my other comment). If you record the child pid (or have the stub write it to a file), an assertion that the process is gone after resolution would pin the escalation too.

apps/server/src/services/agent-browser-service.ts
```diff
@@ -64,35 +64,61 @@ async function probeAgentBrowserVersion(): Promise<{
   missing: boolean;
 }> {
   const { spawn } = await import("node:child_process");
+  const timeoutMs = 5000;
```
vincentri, line 67: Judgement call (Duplicated Code, anchoring on the nearby added line): `probeAgentBrowserVersion` is now a line-for-line twin of `probeHarnessVersion` (`coding-agent-harness-service.ts:708-769`) apart from the command name, and `extractVersion` just above duplicates `coding-agent-harness-service.ts:771-778` verbatim. The twins have already drifted once — SIGKILL escalation landed on one side only — which is the usual argument for extracting a shared spawn-with-timeout helper in a follow-up.

vincentri (commented):
Two-axis review (standards + spec vs #326), run in an isolated worktree at 6ddf0544 with tests executed.

**Bottom line: the issue's ask is implemented faithfully and the repro is properly encoded — Settings returns, the probe resolves as not installed at ~5s, and the `NAKAMA_DISABLE_FIX_PATH` trap is handled. One important gap inherited from the twin probe (#338, fixed on main by #339 after this PR was opened): no SIGKILL escalation, so a SIGTERM-trapping binary leaks a live child per status call. Judged follow-up-grade rather than blocking — but I'd add the ~4 lines here instead of shipping the known bug in a sibling.**

## Spec

- **Implemented per the issue:** "the same 5s timer and SIGTERM the harness probe already uses, resolving as not installed when it fires" — both probe call sites in `getAgentBrowserRuntimeStatus` are covered; on timeout the user-visible result is `{installed: false, ready: false, nextStep: "install"}`, matching the issue verbatim. The `missing: false` internal refinement (skip paying the 5s bound again on the PATH retry) is endorsed by the maintainer on #326.
- **Inherited gap (partial vs. current parity):** main merged #339 (`bae24849`) adding SIGTERM → SIGKILL escalation after `SIGTERM_GRACE_MS = 2000` to `probeHarnessVersion`, precisely because SIGTERM-only orphans children that trap SIGTERM (#338). This PR's copy has no escalation, and #338 names it explicitly: "probeAgentBrowserVersion in #330 copies this function, so it inherits the gap." Not a blocker for this issue's stated AC (Settings must return — it does), but parity with the harness probe now includes escalation; recommend adding it here.
- **Test blind spot:** the new "stubborn" test only asserts resolution, never that the child died — so it cannot catch the missing escalation even if it were present.
- **omni-install tar spawn** named in the issue is legitimately deferred: the author proposed the follow-up on #326 and the maintainer agreed.
- **Scope creep:** none material (a defensive `try/catch` around `spawn` copied verbatim from the twin; spawn failures actually arrive via the `'error'` event). The second SIGTERM site at line ~197 is the pre-existing 120s install timeout, untouched.

Tests run: `bun test apps/server/src/services/agent-browser-service.test.ts` → 7 pass / 0 fail (~17s; hang tests resolve ≈5.0–5.1s as designed).

## Standards

No hard violations. `bun x ultracite check` clean on both files; ts-taste.md respected (single boundary cast only); tests assert behavior (installed/ready/nextStep + wall-clock bound), not error copy.

Judgement calls:

- **Duplicated Code:** `probeAgentBrowserVersion` is a line-for-line twin of `probeHarnessVersion` apart from the command name, and `extractVersion` duplicates its counterpart verbatim. The twins have already drifted once (escalation on one side only) — a shared spawn-with-timeout helper is warranted in a follow-up, not premature now.
- Weak: inline `const timeoutMs = 5000` vs the harness's named constants (matches this file's existing style); defensive try/catch around compile-time-constant args (verbatim from the twin).

**Summary:** Standards — 0 hard violations, 2 judgement calls (worst: the line-for-line probe twin). Spec — 0 missing ACs against the issue text, worst finding: the inherited missing-SIGKILL escalation, recommended for inclusion here or as an immediate follow-up.

Commenting rather than requesting changes: the diff satisfies everything #326 asked for at time of writing; the escalation gap comes from main moving under the PR and deserves a deliberate small addition before or right after merge.

Issue #329, opened 2026-08-20
Research: evaluate alternatives to Composio for SaaS integrations UX


ahmadrosid:
## Problem

Setting up the Composio integration is too complicated. Today it requires multiple moving parts before a user gets a working connection:

- API key + `config.ini` on disk (`composio-config.ts`)
- Per-org and per-user ID mapping (`nakama:org:<id>`, `nakama:user:<id>`)
- OAuth connection flow to connect each SaaS account
- Toolkit assignment via the picker UI
- MCP generation wiring

Each of these is a separate setup step with its own failure modes. The UX for a first-time user is bad.

## Goal

Research alternatives to Composio that give the same capability (connect user SaaS accounts, expose them as tools) with a simpler setup path. We care about UX first, then coverage and maintenance cost.

## What to evaluate

For each candidate (at minimum: composio, plus plausible alternatives), document:

1. **Setup path** — how many steps from zero to first working connection? API key? OAuth? SDK vs MCP vs HTTP?
2. **Auth model** — does it handle per-user OAuth out of the box, or do we build it?
3. **Tool exposure** — MCP server, REST, SDK? How do connected apps become callable tools?
4. **Self-host / BYO-key fit** — does it match Nakama's self-hosting dev-team segment, or is it a hosted dependency?
5. **Pricing & licensing** — per-seat, per-connection, OSS?
6. **Migration cost** — how much of our existing `composio-*` surface would we replace?

## Acceptance criteria

- A short written comparison (a doc or issue comment) ranking candidates on setup UX.
- A recommendation with a clear reasoning: keep Composio, or pick a specific alternative.
- If an alternative wins, a follow-up spike issue to prototype the setup flow.

## Non-goals

- Do not start the migration in this issue. Research + recommendation only.

fajarhide:
@ahmadrosid **Recommendation: keep Composio, close this without a vendor comparison.**

Measured end to end on a throwaway instance: the project API key is the only thing that comes from composio.dev. `resolveAuthConfigId` (`composio-api-client.ts:164`) creates the auth config itself, so the Google Cloud project walkthrough in Composio's own docs never applies here. Following their docs instead of ours is most of the friction.

The rest was ours:

- Two of the six steps in `composio.mdx` were already automatic. `/health` hardcodes `composioAvailable: false` (`system.ts:178`), so step 2 reads as a broken integration, and `completeOAuth` already syncs tools (`composio-service.ts:496`). #335 drops both.
[...]

PR #335, opened 2026-08-20
docs(composio): drop two dead setup steps and rewrite for a first-time reader

fajarhide:
Setup in `composio.mdx` listed six steps, two of which the product already does, and the page opened with the SDK name and the tenancy model before telling anyone how to get the API key.

**The two dead steps.** Step 2 sent readers to `GET /health` to confirm the key was saved. That endpoint runs local checks only and hardcodes `composioAvailable: false` (`system.ts:178`), so a permanently false flag sat next to the one they were told to check and read as a broken integration. Step 5 asked for a Sync tools click, but `completeOAuth` calls `syncUserToolkit` as soon as the connection is stored (`composio-service.ts:496`).

Measured on a throwaway instance with only the API key copied across and nothing configured on the Composio dashboard:

```
/v1/system/status                        composioAvailable true
GET  /v1/composio/toolkits               catalog 200 toolkits, catalogError null
POST /v1/composio/toolkits/gmail/enable  enabled, lastError null
POST .../gmail/connect                   redirectUrl https://connect.composio.dev/link/lk_...
after OAuth in the browser               cachedTools 61, connection connected
```

No Sync click happened in that run. The auth config came from `resolveAuthConfigId`, which is why the page now says the API key is the only thing needed from Composio: their own docs walk you through a Google Cloud project and consent screen that Nakama never uses.

**The rewrite.** Getting the key is now the first thing on the page, by the exact menu path, with the `ck_` consumer key called out because that is the one people paste by mistake. The mechanics that were mixed into the steps, MCP sessions, config file paths, callback URLs, the health endpoint, moved to a section at the end. A symptom table covers the two failures people actually hit: an unconnected account and an unassigned profile both look like the agent ignoring the app.

Setup reads as five steps now, four of which are clicks in Nakama.

`bun run build:docs` passes.

Refs #329

**Screenshots.** Four shots from a real instance now sit next to the steps: the empty Project API key card, the toolkit list, a Gmail row waiting to be connected, and the profile picker.

Steps 4 and 5 describe the Integrations page as it behaves with #337, so this should merge after that one.


ahmadrosid (approved):
Nice rewrite. Setup now starts where a first-time reader actually starts — the click path to the project API key — instead of the SDK name and the tenancy model.

Checked the two dropped steps against main:

- `GET /health` still hardcodes `composioAvailable: false` in `system.ts` so the old “confirm /health” step would have looked like a broken integration. Reachability belongs on `/v1/system/status`.
- `completeOAuth` already calls `syncUserToolkit` once the connection is stored, so the Sync click was leftover. Sync remains in the overflow menu for a reconnect, which is the right place.

The `ck_` callout, the symptom table, and the four screenshots from a real instance are the parts that will save people time. The “How it works underneath” section is the right home for MCP sessions, config paths, and the health endpoint.

One merge constraint: steps 4 and 5, plus those screenshots, describe the Integrations page as it exists in #337 (Connect on the row, popular-apps list, profiles picker, enable assigning the default profile). **Merge this after #337**, otherwise the page will document buttons that are not on main yet.

Two small follow-ups, neither blocking:

- The troubleshooting row still says “press Connect your account”; the button in #337 (and in the screenshot here) is **Connect**.
- Telegram / WhatsApp / Discord / CLI still map to the earliest human org admin’s Composio connection (`resolveComposioActingUserId`). The old page said that; the new one only covers the public-URL failure. Worth one line in the symptom table or “How it works”, since “the agent uses whoever is talking” is true on web and not on those bridges.

Approved. Hold the merge until #337 is in.

fajarhide:
@ahmadrosid Fixed in the latest commit! I've updated the button text to '**Connect**' and added a note about how bridges resolve the acting user in the '**How it works**' section.

PR #337, opened 2026-08-20
feat(composio): connect, auto-assign and a shorter app list on Integrations

fajarhide:
Enabling a Composio toolkit used to leave it usable by nobody, and for most org admins there was no way out of that.

There was no Connect button anywhere in `apps/web` (nothing called `client.connectComposioToolkit`), and the chat fallback only offers `composio__connect_account` for toolkits already assigned to a profile (`composio-tool-bridge.ts:185`), so a freshly enabled toolkit had no path to a connection at all. `composio.mdx` pointed at a button that did not exist. Assignment was worse: `PUT /v1/profiles/:profileId/composio-toolkits` allows an org admin, but its only UI lives on `/profiles`, which `PlatformAdminGuard` restricts to platform admins.

Four changes, all on the Integrations page.

**Connect.**

| before | after |
| --- | --- |
| before | after |

**Enabling assigns the toolkit to the org's default profile**, and disabling removes it from every profile in the org. The Profiles page then already reads 1 assigned without anyone going there:

assigned

**The list opens on popular apps** instead of all 200, with a toggle for the full catalog. Search always covers everything, so a niche app stays one query away.

list

**A profiles button on each enabled row** for everything beyond the default, so an org admin never needs the platform admin page:

row

dialog

Super profiles are labelled as running bash, because that is the assignment worth a second thought. Assignment deliberately does not follow a disconnect: connections are per member, so dropping it when one member disconnects would take the toolkit from everyone else. The bridge already filters on the chatting user's connection.

Verified against a live instance, reading the server rather than the UI. Clicking Connect through Google consent left one connected row and gmail with 61 cached tools, both empty beforehand, with no Sync click. On a fresh org, enable moved assignments 0 to 1 (default profile only, not Super Bot), disable moved it back to 0, and re-enable restored it. Toggling Super Bot in the new dialog wrote the second row:

```
profile      toolkit_slug
Default Bot  gmail
Super Bot    gmail
```

The two new service tests fail without the change (10 pass 2 fail, then 12 pass).

`bun run test`: 2284 tests across 11 workspaces, 0 fail. `bun run check`: clean.

Refs #329

### DIFF
--- a/apps/server/src/services/composio-service.test.ts
+++ b/apps/server/src/services/composio-service.test.ts
@@ -154,6 +154,78 @@ describe("ComposioService", () => {
     }
   });
 
+  test("enableToolkit assigns the toolkit to the org default profile", async () => {
+    const { db, service, restore } = await createConfiguredService();
+
+    try {
+      await db.upsertProfile({
+        createdAt: new Date().toISOString(),
+        id: "profile_default",
+        isDefault: true,
+        isSuper: false,
+        model: null,
+        name: "Default",
+        orgId: ORG_ID,
+        systemPrompt: "",
+      });
+      await db.upsertProfile({
+        createdAt: new Date().toISOString(),
+        id: "profile_super",
+        isDefault: false,
+        isSuper: true,
+        model: null,
+        name: "Super Bot",
+        orgId: ORG_ID,
+        systemPrompt: "",
+      });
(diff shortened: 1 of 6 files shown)


ahmadrosid (approved):
LGTM

PR #339, opened 2026-08-20
fix(server): kill a SIGTERM-proof harness after the version probe times out

### LINKED ISSUE #338: probeHarnessVersion leaves the child running when a harness traps SIGTERM
#275 bounded the SIGTERM-proof case at the probe run and the install: SIGTERM on timeout, then SIGKILL after `SIGTERM_GRACE_MS`. It missed the third site. `probeHarnessVersion` (`coding-agent-harness-service.ts:741`) still only sends SIGTERM, so a harness that traps it outlives the probe. That one is mine.

Repro with a stub `claude` that traps SIGTERM and never answers `--version`, driving `listCodingAgentHarnessStatuses` against an in-memory adapter:

```
probe returned in 5007ms, claude installed=false

  PID  ELAPSED  COMMAND
70586    00:12  bun run orphan-repro.ts
70593    00:11  .../stub-harness/claude --version
```

The probe resolves on its timer as designed. The child is still running 12 seconds later, and its open stdio pipes keep the parent process from exiting at all, which is the part that hurts: the repro script never returned to the shell.

`listCodingAgentHarnessStatuses` probes every configured harness (`:190`), so a harness in that state adds one orphan per Settings load.

Fix is the `killTimeoutId` escalation the two sibling sites already carry.

`probeAgentBrowserVersion` in #330 copies this function, so it inherits the gap.

fajarhide:
`probeHarnessVersion` sends SIGTERM when the 5s bound fires and returns. A harness that traps SIGTERM keeps running, and its open stdio pipes keep the parent process alive with it. #275 added the SIGKILL escalation to the probe run and the install and missed this third site.

Same repro as the issue: a stub `claude` that traps SIGTERM and never answers `--version`, driven through `listCodingAgentHarnessStatuses`, then `ps` 12 seconds later.

before
```
probe returned in 5007ms, claude installed=false
70586    00:12 bun run orphan-repro.ts
70593    00:11 .../stub-harness/claude --version
```

after
```
probe returned in 5012ms, claude installed=false
(nothing left in ps, and the script exits on its own)
```

The new test writes the stub's pid to a file, because the probe never exposes the child, then waits for that pid to disappear. With the service change stashed it fails:

```
(fail) a harness that traps SIGTERM is killed once the version probe times out [20121.02ms]
```

`bun run test` on this branch: 883 server tests, 0 fail, and web, cli and the platform workspaces green.

`probeAgentBrowserVersion` in #330 is a copy of this function, so it inherits the same gap. I have not touched that branch; @rioaguspermana, the three lines here drop straight in if you want them, otherwise it is a follow-up once yours lands.

Fixes #338

### DIFF
--- a/apps/server/src/services/coding-agent-harness-service.test.ts
+++ b/apps/server/src/services/coding-agent-harness-service.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, test } from "bun:test";
-import { mkdtemp, rm, writeFile } from "node:fs/promises";
+import { mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import path from "node:path";
 import { createInMemoryDatabaseAdapter } from "@nakama/db";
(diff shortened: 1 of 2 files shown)


ahmadrosid (approved):
LGTM — this is the third probe site #275 missed.

`probeHarnessVersion` now matches the run/install paths: SIGTERM on the 5s bound, then SIGKILL after `SIGTERM_GRACE_MS`. The test is the right shape for a child the probe never exposes: stub writes its pid, traps SIGTERM, and we wait for that pid to disappear.

Noted on `probeAgentBrowserVersion` in #330 — same three lines once that lands.

Merging.

PR #350, opened 2026-08-21
fix(agent): scale pruneToolOutputs thresholds off the model context window (#342)

### LINKED ISSUE #342: fix(agent): tool-output pruning ignores the context window and deletes stored tool output
`pruneToolOutputs` truncates old tool results, and the truncation is written back to the session's stored messages. Its thresholds are module constants; the model's context window is never read. On a 1M-token model a session therefore loses old tool output while sitting at a fraction of the usable window, and the loss is a hard delete with no archive.

Measured on `f1be072b`.

## What the code does

`pruneToolOutputs` takes only the message array (`packages/agent/src/history-compaction.ts:221`). Its thresholds are module constants: `PRUNE_MINIMUM = 20_000` (`history-compaction.ts:13`) and `PRUNE_PROTECT = 40_000` (`history-compaction.ts:14`). `CompactionConfig` is not a parameter, so `contextWindow` cannot reach the decision.

`compactHistory` prunes first (`history-compaction.ts:286`) and computes overflow afterwards (`history-compaction.ts:292`), so pruning does not depend on pressure.

`sendMessage` calls `runCompaction(false)` on every turn and discards the response (`packages/agent/src/chat.ts:410`), so an ordinary turn reports nothing about it. `send` and `sendStream` both route through that same `sendMessage` (`packages/agent/src/chat.ts:285`, `packages/agent/src/chat.ts:304`), so the streaming path the dashboard uses is not a different one. `prunedTokens` is reported only in the response to an explicit compact request (`apps/server/src/http/routes/sessions.ts:88`).

`runCompaction` bumps the history revision for any action other than `none`, which sends the wrapper down the full-replace branch (`apps/server/src/services/session-persistence.ts:104`) into `replaceSessionHistory` (`session-persistence.ts:90`). The SQLite adapter implements that as `DELETE FROM session_messages WHERE session_id = ?` and re-insert (`packages/db/src/adapters/sqlite.ts:627`). `session_messages` is the only one of the 33 tables in `packages/db/sql/schema.sql` that holds chat messages, so the original text is gone.

The record that is overwritten is the one `GET /v1/sessions/:sessionId/messages` serves (`apps/server/src/services/agent-service.ts:1602`), the one `branchSession` copies into the new session (`agent-service.ts:1679`), and the one the web transcript renders as the tool result (`apps/web/src/lib/chat-history.ts:479`).

`resolveCompactionConfig` takes `contextWindow` straight from the model catalog (`agent-service.ts:3585`). Six of the 38 entries in `apps/server/src/providers/models.ts` are `1_000_000`; the smallest is `7968`.

For contrast, the caps that exist today are applied at the tool, before the output ever enters the history: `web_fetch` stops at `MAX_CONTENT_CHARS = 16_000` (`packages/core/src/tools/web-fetch.ts:62`), and the other tool-output optimiser is a workspace setting with its own route, `tokenOptimizerEnabled` (`packages/db/src/types.ts:179`, `apps/server/src/http/routes/token-optimization.ts:161`). This one has no switch and edits the history afterwards.

## Measured

One seeded history, one ordinary turn, three windows:

| contextWindow | usable | estimated tokens before the turn | share of usable | tool results truncated |
|---|---|---|---|---|
| 1,000,000 | 991,808 | 90,121 | 9.09% | 2 of 3 |
| 200,000 | 191,808 | 90,121 | 46.99% | 2 of 3 |
| 128,000 | 119,808 | 90,121 | 75.22% | 2 of 3 |

In the 1M case the session's stored payload went from 360,804 bytes to 120,937: 239,867 bytes of tool output deleted from the database while the context was 9% full. Running the same turn against a real SQLite database, with the migrations and the foreign keys, removes the same 239,867 bytes. That figure is read off the rows themselves, so it does not depend on the token estimate being right. The provider was called once, for the turn itself, so nothing was summarized, and branching cannot recover the text because `branchSession` copies the stored rows.

The trigger is the accumulated tool output, not the percentage: the estimate is the same 90,121 in every row, and only the window moves. The two constants also interact in a way worth knowing. On the same shape with a single tool result, 40,000 estimated tokens survives and 40,001 is truncated: the message that crosses `PRUNE_PROTECT` is counted into the pruned total in full, so `PRUNE_MINIMUM` does not hold back a single result that large. That is about 156 KB in one tool result, and nothing in this repo is that big, the largest first-party TypeScript file being `packages/db/src/adapters/sqlite.ts` at 102 KB. The accumulating case in the table needs no single large result at all, and that is the one I would expect to happen. The seed is three tool results of 120,000 characters each. That is not an unusual size. `read_file` returns the whole file when `limit` is null and is bounded only by `maxFileBytes`, which defaults to 10 MB (`packages/core/src/tools/paths.ts:15`, `packages/core/src/tools/builtin.ts:799`), so a single call can return about 29 times the entire seed.

## Reproduce

```bash
git clone https://github.com/ahmadrosid/nakama.git nakama-prune
cd nakama-prune
git checkout f1be072b
bun install
# save the block below as apps/server/src/services/prune-window-probe.test.ts, then:
bun test apps/server/src/services/prune-window-probe.test.ts
```

```ts
import { createAgentHarness, usableContextTokens } from "@nakama/agent";
import type {
  ChatCompletionResult,
  ChatMessage,
  ProviderClient,
} from "@nakama/core";
import {
  createInMemoryDatabaseAdapter,
  createSqliteDatabase,
} from "@nakama/db";
import {
  loadSessionHistory,
  replaceSessionHistory,
  wrapPersistedSession,
} from "./session-persistence";

const TRUNCATION = "[output truncated by compaction]";
// 120_000 characters is 30_000 tokens at the 4-chars-per-token ratio the
// estimator uses, so three of them clear PRUNE_PROTECT plus PRUNE_MINIMUM.
const TOOL_CHARS = 120_000;
const SESSION_ID = "session_prune_probe";

function toolOutput(marker: string): string {
  return marker.repeat(TOOL_CHARS / marker.length);
}

/** Three turns that each ran one large tool, then one quiet turn. */
function seedHistory(): ChatMessage[] {
  const history: ChatMessage[] = [];

  for (const turn of [1, 2, 3]) {
    history.push({ content: `turn ${turn}`, role: "user" });
    history.push({
      content: "",
      role: "assistant",
      toolCalls: [
        { arguments: { path: `/f${turn}` }, id: `call_${turn}`, name: "read" },
      ],
    });
    history.push({
      content: toolOutput(`t${turn}-`),
      name: "read",
      role: "tool",
      toolCallId: `call_${turn}`,
    });
    history.push({ content: `answer ${turn}`, role: "assistant" });
  }

  history.push({ content: "turn 4", role: "user" });
  history.push({ content: "answer 4", role: "assistant" });

  return history;
}

function countingProvider(): { calls: number[]; provider: ProviderClient } {
  const calls: number[] = [];
  const reply: ChatCompletionResult = {
    assistantMessage: { content: "ok", role: "assistant" },
    content: "ok",
    toolCalls: [],
  };

  return {
    calls,
    provider: {
      generateChat() {
        calls.push(1);
        return Promise.resolve(reply);
      },
      generateText() {
        return Promise.resolve({ content: "ok" });
      },
      name: "openai",
      streamChat(_input, handlers) {
        calls.push(1);
        handlers.onChunk("ok");
        return Promise.resolve(reply);
      },
    },
  };
}

function toolBodies(history: readonly ChatMessage[]): string[] {
  return history
    .filter((message) => message.role === "tool")
    .map((message) => message.content.slice(0, TRUNCATION.length));
}

function truncatedCount(history: readonly ChatMessage[]): number {
  return toolBodies(history).filter((body) => body === TRUNCATION).length;
}

describe("tool-output pruning ignores the model context window", () => {
  test("one ordinary turn on a 1M window deletes stored tool output", async () => {
    const compaction = { contextWindow: 1_000_000, maxOutputTokens: 8192 };
    const { calls, provider } = countingProvider();

    await databaseAdapter.upsertSession({
      agentQuestionnaire: null,
      agentTodos: [],
      channel: "web",
      createdAt: new Date().toISOString(),
      id: SESSION_ID,
      orgId: "org_1",
      profileId: "profile_1",
      title: null,
      userId: "user_1",
    });
    await replaceSessionHistory(databaseAdapter, SESSION_ID, seedHistory());

    const storedBytes = async () =>
      (await databaseAdapter.listMessagesForSession(SESSION_ID)).reduce(
        (total, record) => total + JSON.stringify(record.payload).length,
        0
      );
    const bytesBefore = await storedBytes();

    const session = wrapPersistedSession(
      SESSION_ID,
      createAgentHarness({ provider }).createChatSession({
        compaction,
        initialHistory: await loadSessionHistory(databaseAdapter, SESSION_ID),
      }),
      databaseAdapter
    );
    const usedBefore = session.getContextUsage()?.usedTokens ?? 0;

    await session.send("turn 5");

    const bytesAfter = await storedBytes();
    const stored = await loadSessionHistory(databaseAdapter, SESSION_ID);
    console.log({
      bytesAfter,
      bytesBefore,
      shareOfUsable: `${((usedBefore / usableContextTokens(compaction)) * 100).toFixed(2)}%`,
      truncated: truncatedCount(stored),
      usableContextTokens: usableContextTokens(compaction),
      usedBefore,
    });

    // One provider call, for the turn itself: no summarization ran, because
    // nothing was anywhere near overflowing.
    expect(calls).toHaveLength(1);
    expect(usedBefore).toBeLessThan(usableContextTokens(compaction) * 0.1);

    // Two of the three tool results are gone from the stored transcript that
    // GET /v1/sessions/:id/messages serves and branchSession reads.
    expect(truncatedCount(stored)).toBe(2);
    expect(bytesAfter).toBeLessThan(bytesBefore / 2);
  });

  test("the same turn prunes the same way at 1M, 200k and 128k", async () => {
    const outcomes: string[][] = [];

    for (const contextWindow of [1_000_000, 200_000, 128_000]) {
      const compaction = { contextWindow, maxOutputTokens: 8192 };
      const { provider } = countingProvider();
      const session = createAgentHarness({ provider }).createChatSession({
        compaction,
        initialHistory: seedHistory(),
      });
      const usedBefore = session.getContextUsage()?.usedTokens ?? 0;

      await session.send("turn 5");

      const history = session.getHistory();
      console.log({
        contextWindow,
        shareOfUsable: `${((usedBefore / usableContextTokens(compaction)) * 100).toFixed(2)}%`,
        truncated: truncatedCount(history),
        usableContextTokens: usableContextTokens(compaction),
        usedBefore,
      });
      expect(truncatedCount(history)).toBe(2);
      outcomes.push(toolBodies(history));
    }

    expect(outcomes[0]).toEqual(outcomes[1] as string[]);
    expect(outcomes[1]).toEqual(outcomes[2] as string[]);
  });

  test("a single tool result is truncated one token past PRUNE_PROTECT", async () => {
    const single = (tokens: number): ChatMessage[] => [
      { content: "turn 1", role: "user" },
      {
        content: "",
        role: "assistant",
        toolCalls: [{ arguments: {}, id: "c1", name: "read" }],
      },
      {
        content: "x".repeat(tokens * 4),
        name: "read",
        role: "tool",
        toolCallId: "c1",
      },
      { content: "answer 1", role: "assistant" },
      { content: "turn 2", role: "user" },
      { content: "answer 2", role: "assistant" },
    ];

    const survives = async (tokens: number) => {
      const { provider } = countingProvider();
      const session = createAgentHarness({ provider }).createChatSession({
        compaction: { contextWindow: 1_000_000, maxOutputTokens: 8192 },
        initialHistory: single(tokens),
      });
      await session.send("turn 3");
      return truncatedCount(session.getHistory()) === 0;
    };

    // PRUNE_MINIMUM does not hold this back: the message that crosses
    // PRUNE_PROTECT is counted into the pruned total in full.
    expect(await survives(40_000)).toBe(true);
    expect(await survives(40_001)).toBe(false);
  });

  test("real SQLite, with migrations and foreign keys, does the same", async () => {
    const { provider } = countingProvider();
    const database = await createSqliteDatabase(":memory:");
    const databaseAdapter = database.adapter;
    const now = new Date().toISOString();

    // sessions.profile_id is a real foreign key here, unlike in the map-backed
    // adapter, so the profile has to exist.
    await databaseAdapter.upsertProfile({
      createdAt: now,
      id: "profile_1",
      isSuper: false,
      model: null,
      name: "probe",
      orgId: null,
      systemPrompt: "",
      updatedAt: now,
    });
    await databaseAdapter.upsertSession({
      agentQuestionnaire: null,
      agentTodos: [],
      channel: "web",
      createdAt: now,
      id: SESSION_ID,
      profileId: "profile_1",
      title: null,
      userId: null,
    });
    await replaceSessionHistory(databaseAdapter, SESSION_ID, seedHistory());

    const storedBytes = async () =>
      (await databaseAdapter.listMessagesForSession(SESSION_ID)).reduce(
        (total, record) => total + JSON.stringify(record.payload).length,
        0
      );
    const bytesBefore = await storedBytes();

    const session = wrapPersistedSession(
      SESSION_ID,
      createAgentHarness({ provider }).createChatSession({
        compaction: { contextWindow: 1_000_000, maxOutputTokens: 8192 },
        initialHistory: await loadSessionHistory(databaseAdapter, SESSION_ID),
      }),
      databaseAdapter
    );

    await session.send("turn 5");

    const bytesAfter = await storedBytes();
    console.log({ bytesAfter, bytesBefore, removed: bytesBefore - bytesAfter });
    expect(
      truncatedCount(await loadSessionHistory(databaseAdapter, SESSION_ID))
    ).toBe(2);
    expect(bytesBefore - bytesAfter).toBe(239_867);
    database.close();
  });
});
```

The probe drives the real pieces: `createChatSession` with the same `compaction` and `initialHistory` shape `AgentService` passes at `agent-service.ts:3183`, wrapped by the same `wrapPersistedSession` it uses at `agent-service.ts:3327`. The last of the four tests runs the same turn against `createSqliteDatabase(":memory:")`, so the schema, the foreign keys and the `DELETE` are the real ones and not a map standing in for them. It is a reproduction, not a test I am proposing to add, and the exact token counts depend on the system prompt and will drift. It registers no tools, so the estimate leaves out tool schemas; adding the builtin set moves it by about 1,400 tokens.

## What I am not claiming

Not that pruning is wrong. Replaying 90k tokens of stale tool output every turn costs money on a 1M model too, so a cost-driven prune is a defensible choice.

Not that any real session has hit this. I measured the code path, not usage.

## The question

Should the prompt view and the stored transcript be the same array? Right now they are, so an optimisation aimed at the prompt also rewrites the history a user scrolls through and branches from.

Two directions, and I would build either:

1. Scale the thresholds off `contextWindow` rather than leaving them as constants, so a 1M model prunes under 1M-scale pressure and the `7968` model prunes before the fixed thresholds are anywhere near reached.
2. Leave the stored transcript intact and prune only on the way to the provider, beside `rehydrateMessagesForProvider`, which already exists as a per-request transform.

I lean toward 2 because it also fixes branching, but that is a judgement, not a measurement, and 1 is the much smaller change.

Per CONTRIBUTING I will not open a PR before this is assigned.

vincentri:
## Summary

`pruneToolOutputs` rewrote stored tool output using module-level constants (`PRUNE_PROTECT = 40_000`, `PRUNE_MINIMUM = 20_000`) regardless of the model's context window. On a 1M-token model a session with three 30k-token tool results lost **240 kB of stored tool output on a single ordinary turn** while the context was at 9% of its usable window. The truncation is written back to `session_messages`, so the same rows `GET /v1/sessions/:id/messages`, `branchSession`, and the web transcript all read are the rows that got rewritten. `branchSession` cannot recover the text because it copies the stored rows.

This PR scales the thresholds off `usableContextTokens(compaction)` (direction 1 of the two options the issue lays out). `compactHistory` already received the session's `CompactionConfig`, so no changes were needed in `chat.ts`; the PR teaches `pruneToolOutputs` to read that config, which is what lets the per-turn `runCompaction(false)` path prune against the real window size.

## What the code did before

| | File | Line |
|---|---|---|
| Module constants | `packages/agent/src/history-compaction.ts` | 13–14 |
| `pruneToolOutputs` (no `compaction` arg) | `packages/agent/src/history-compaction.ts` | 221 |
| Call from `compactHistory` (drops `compaction`) | `packages/agent/src/history-compaction.ts` | 286 |
| Per-turn compaction driver | `packages/agent/src/chat.ts` | 410 |
| Config built from model catalog | `apps/server/src/services/agent-service.ts` | 3569–3588 |

`pruneToolOutputs` took only `messages: ChatMessage[]`. `contextWindow` was never on the call path.

## What it does now

```ts
const PRUNE_PROTECT_FRACTION = 0.5; // protect tool output up to 50% of usable
const PRUNE_MINIMUM_FRACTION = 0.1; // only act when pruning >= 10% of usable

export function pruneToolOutputs(
  messages: ChatMessage[],
  compaction: CompactionConfig
): { prunedTokens: number } {
  const usable = usableContextTokens(compaction);
  if (usable <= 0) return { prunedTokens: 0 }; // degenerate configs: no-op
  const protect = Math.floor(usable * PRUNE_PROTECT_FRACTION);
  const minimum = Math.floor(usable * PRUNE_MINIMUM_FRACTION);
  // ...same loop, using `protect`/`minimum` instead of the module constants
}
```

`compactHistory` now passes `input.compaction` through. `usableContextTokens` is the same helper the existing overflow check (`isOverflow`) uses, so the two are consistent.

## Behavior change

A note on call shapes first, because it changes the numbers. `seedHistory`
returns 11 messages and the unit tests call `pruneToolOutputs` on that array
directly. On the per-turn path `compactHistory` sees the array after the new
turn has been appended, so the two-turn tail slides forward and all three tool
results become candidates instead of two. The per-turn path is the one that
writes back to `session_messages`, so the table below uses that shape
throughout. Every cell was measured by running both implementations over the
same seeds, not derived by hand.

Before/after on the issue's seed (three 30k-token tool results, one ordinary turn sent on top):

| Window | Usable | Old result | New result |
|---|---:|---|---|
| 1,000,000 | 991,808 | **2 of 3 truncated, 240 kB deleted** | no prune ✓ |
| 200,000 | 191,808 | 2 of 3 truncated | no prune ✓ |
| 128,000 | 119,808 | 2 of 3 truncated | 2 of 3 truncated |

Called directly on the 11-message array, as the unit tests do, the old code
truncates 1 of 3 at every window (its constants are window-blind) and the new
code truncates 1 of 3 at 128k and 100k and 0 elsewhere. Both "no prune" cells
hold under either shape.

Two rows from the test suite, each its own seed, called directly:

| Seed | Old | New |
|---|---|---|
| existing test's transcript (one 50,000-token result, 100k window) | 50k truncated | 50k truncated (test unchanged) |
| degenerate guard test (`seedHistory(200_000)`, 4,096 window) | 2 of 3 truncated (100k tokens) | no prune (usable ≤ 0 guard) |

The 200k row is a deliberate behavior change: under 50% of usable, the history
is left alone. The bug report's intent is that pruning is reserved for actual
overflow pressure, not as a flat cost optimisation.

Two things worth knowing about the 128k row: it is also where
`resolveCompactionConfig` lands for any model the catalog does not carry
(`contextWindow: model?.contextWindow ?? 128_000`), so uncataloged models keep
today's behavior; and on the smallest catalog entry (7968) the old code's
stored-byte counter reads 0 only because overflow summarization drops the
whole head afterwards — the text was lost either way, so the small model
pruning earlier is intended behavior, not a safety baseline.

## Tests

`packages/agent/src/history-compaction.test.ts` (11/11 pass, 28 expects):

- **existing** `prunes old tool outputs while protecting recent turns` — still passes (100k window, 50k tool result)
- **new** `does not prune when tool output is well below the model's usable window (#342)` — pins the bug fix (1M window, 60k tool output → no prune)
- **new** `still prunes when accumulated tool output crosses the protect fraction` — pins preserved overflow behavior (128k window)
- **new** `does not truncate when reclaimed tokens do not clear the minimum fraction` — pins the below-minimum no-op branch (200k window; the reclaim equals the minimum, so nothing is rewritten)
- **new** `does not prune when usable tokens are non-positive` — pins the degenerate guard (4k window, 8k maxOutput → no prune)

Wider sweep:

```
bun test packages/                 # 808 pass, 0 fail, 1696 expects
bun test apps/server/src/services  # 409 pass, 0 fail, 1010 expects
bun x ultracite check              # 1175 files clean
```

## Out of scope (deliberately)

- **Direction 2** from the issue — separate prompt-time prune (`rehydrateMessagesForProvider`) from stored-transcript edits. That also fixes `branchSession` losing text, but it touches `rehydrateMessagesForProvider`, `session-persistence`, and the SQLite adapter. Worth a follow-up.
- The follow-up timeout bug in `omni-install.ts:204` from #326 — separate issue.
- Adding a kill-switch / config knob to disable pruning. With the new fractions, the worst-case deletion is bounded by usable context, so a knob is no longer required for correctness. If product wants it later, add it on top.

## Reproduction (from the issue)

```bash
git clone https://github.com/ahmadrosid/nakama.git nakama-prune
cd nakama-prune
git checkout f1be072b
bun install
# paste the prune-window-probe.test.ts block from the issue body
bun test apps/server/src/services/prune-window-probe.test.ts
# 1M window: 240 kB removed from session_messages while context was 9% full
```

After this PR, on the same seed and a 1M window the probe shows `prunedTokens: 0` and no tool output is touched; the stored payload ends up 69 bytes larger, which is the two rows the turn itself writes (`turn 5` / `ok`).

Closes #342

### DIFF
--- a/packages/agent/src/history-compaction.test.ts
+++ b/packages/agent/src/history-compaction.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, test } from "bun:test";
 import type {
   ChatCompletionResult,
   ChatMessage,
+  CompactionConfig,
   ProviderClient,
 } from "@nakama/core";
 import {
@@ -14,8 +15,18 @@ import {
   usableContextTokens,
 } from "./history-compaction";
 
-const compaction = {
+const compaction: CompactionConfig = {
   contextWindow: 100_000,
+  maxOutputTokens: 8192,
+};
+
+const largeWindow: CompactionConfig = {
+  contextWindow: 1_000_000,
+  maxOutputTokens: 8192,
+};
+
+const smallWindow: CompactionConfig = {
+  contextWindow: 128_000,
   maxOutputTokens: 8192,
 };
 
@@ -30,6 +41,26 @@ function createToolMessage(content: string): ChatMessage {
     role: "tool",
     toolCallId: "call_1",
   };
+}
+
+// Builds the 14-message transcript used by the prune tests: three turns
+// each carrying one tool result, followed by a tail turn with no tool call.
+// The last two user turns are protected, so only the older tool messages
+// are pruning candidates.
+function seedHistory(toolChars: number): ChatMessage[] {
+  return [
+    { content: "turn 1", role: "user" },
+    createToolMessage(repeat("a", toolChars)),
+    { content: "done 1", role: "assistant" },
+    { content: "turn 2", role: "user" },
+    createToolMessage(repeat("b", toolChars)),
+    { content: "done 2", role: "assistant" },
+    { content: "turn 3", role: "user" },
+    createToolMessage(repeat("c", toolChars)),
+    { content: "done 3", role: "assistant" },
+    { content: "turn 4", role: "user" },
+    { content: "done 4", role: "assistant" },
+  ];
 }
 
 describe("history compaction", () => {
@@ -55,7 +86,7 @@ describe("history compaction", () => {
       { content: "done 4", role: "assistant" },
     ];
 
-    const result = pruneToolOutputs(messages);
+    const result = pruneToolOutputs(messages, compaction);
 
     expect(result.prunedTokens).toBeGreaterThan(0);
     expect(messages[1]?.role === "tool" && messages[1].content).toContain(
@@ -63,6 +94,56 @@ describe("history compaction", () => {
     );
     expect(messages[10]?.role === "assistant" && messages[10].content).toBe(
       "done 4"
+    );
+  });
+
+  test("does not prune when tool output is well below the model's usable window (#342)", () => {
+    // Three 30k-token tool results; the last 2 turns are protected, so two
+    // candidates (60k tokens) are considered. On a 1M window the protect
+    // fraction is ~496k, so the loop must leave the stored transcript intact.
+    const messages = seedHistory(120_000);
+    const result = pruneToolOutputs(messages, largeWindow);
+
+    expect(result.prunedTokens).toBe(0);
+    expect(messages[1]?.role === "tool" && messages[1].content).toBe(
+      repeat("a", 120_000)
+    );
+    expect(messages[4]?.role === "tool" && messages[4].content).toBe(
+      repeat("b", 120_000)
+    );
+    expect(messages[7]?.role === "tool" && messages[7].content).toBe(
+      repeat("c", 120_000)
+    );
+  });
+
+  test("still prunes when accumulated tool output crosses the protect fraction", () => {
+    // Same three 30k-token tool results against a 128k window. The protect
+    // fraction is ~60k, so the older candidate crosses the threshold and the
+    // function must truncate it.
+    const messages = seedHistory(120_000);
+    const result = pruneToolOutputs(messages, smallWindow);
+
+    expect(result.prunedTokens).toBeGreaterThan(0);
+    expect(messages[1]?.role === "tool" && messages[1].content).toContain(
+      "truncated"
+    );
+    expect(messages[4]?.role === "tool" && messages[4].content).toBe(
+      repeat("b", 120_000)
+    );
+  });
+
+  test("does not prune when usable tokens are non-positive", () => {
+    const degenerate: CompactionConfig = {
+      contextWindow: 4096,
+      maxOutputTokens: 8192,
+    };
+
+    const messages = seedHistory(200_000);
+    const result = pruneToolOutputs(messages, degenerate);
+
+    expect(result.prunedTokens).toBe(0);
+    expect(messages[1]?.role === "tool" && messages[1].content).toBe(
+      repeat("a", 200_000)
     );
   });
 
--- a/packages/agent/src/history-compaction.test.ts
+++ b/packages/agent/src/history-compaction.test.ts
@@ -132,6 +132,39 @@ describe("history compaction", () => {
     );
   });
 
+  test("does not truncate when reclaimed tokens do not clear the minimum fraction", () => {
+    // usable = 200k (protect = 100k, minimum = 20k). Walking newest→oldest,
+    // the 95k-token result stays under protect and the 20k-token one crosses
+    // it, but the reclaim equals the minimum, so nothing may be rewritten.
+    const messages: ChatMessage[] = [
+      { content: "turn 1", role: "user" },
+      createToolMessage(repeat("a", 80_000)),
+      { content: "done 1", role: "assistant" },
+      { content: "turn 2", role: "user" },
+      createToolMessage(repeat("b", 380_000)),
+      { content: "done 2", role: "assistant" },
+      { content: "turn 3", role: "user" },
+      createToolMessage(repeat("c", 4000)),
+      { content: "done 3", role: "assistant" },
+      { content: "turn 4", role: "user" },
+      { content: "done 4", role: "assistant" },
+    ];
+    const window: CompactionConfig = {
+      contextWindow: 208_192,
+      maxOutputTokens: 8192,
+    };
+
+    const result = pruneToolOutputs(messages, window);
+
+    expect(result.prunedTokens).toBe(0);
+    expect(messages[1]?.role === "tool" && messages[1].content).toBe(
+      repeat("a", 80_000)
+    );
+    expect(messages[4]?.role === "tool" && messages[4].content).toBe(
+      repeat("b", 380_000)
+    );
+  });
+
   test("does not prune when usable tokens are non-positive", () => {
     const degenerate: CompactionConfig = {
       contextWindow: 4096,
--- a/packages/agent/src/history-compaction.test.ts
+++ b/packages/agent/src/history-compaction.test.ts
@@ -2,11 +2,11 @@ import { describe, expect, test } from "bun:test";
 import type {
   ChatCompletionResult,
   ChatMessage,
-  CompactionConfig,
   ProviderClient,
 } from "@nakama/core";
 import {
   buildCompactionPrompt,
+  type CompactionConfig,
   compactHistory,
   estimateHistoryTokens,
   isOverflow,
--- a/packages/agent/src/history-compaction.ts
+++ b/packages/agent/src/history-compaction.ts
@@ -10,8 +10,12 @@ import {
 } from "@nakama/core";
 
 const COMPACTION_BUFFER = 20_000;
-const PRUNE_MINIMUM = 20_000;
-const PRUNE_PROTECT = 40_000;
+// Pruning thresholds are fractions of the model's usable context so that
+// large-window models keep their history intact when tool output is small
+// relative to the window, while small-window models still reclaim tokens
+// before overflow. See #342.
+const PRUNE_PROTECT_FRACTION = 0.5;
+const PRUNE_MINIMUM_FRACTION = 0.1;
 const TAIL_TURNS = 2;
 const TOKEN_ESTIMATE_RATIO = 4;
 const PRUNE_TRUNCATION = "[output truncated by compaction]";
@@ -218,9 +222,20 @@ export function selectCompactionRange(
   };
 }
 
-export function pruneToolOutputs(messages: ChatMessage[]): {
-  prunedTokens: number;
-} {
+export function pruneToolOutputs(
+  messages: ChatMessage[],
+  compaction: CompactionConfig
+): { prunedTokens: number } {
+  const usable = usableContextTokens(compaction);
+  const protect = Math.floor(usable * PRUNE_PROTECT_FRACTION);
+  const minimum = Math.floor(usable * PRUNE_MINIMUM_FRACTION);
+
+  // Degenerate configs (small contextWindow vs. maxOutputTokens) yield a
+  // non-positive usable budget; pruning would have nothing to protect.
+  if (usable <= 0) {
+    return { prunedTokens: 0 };
+  }
+
   let total = 0;
   let pruned = 0;
   const toPrune: Extract<ChatMessage, { role: "tool" }>[] = [];
@@ -260,7 +275,7 @@ export function pruneToolOutputs(messages: ChatMessage[]): {
     const estimate = estimateTokens(message.content);
     total += estimate;
 
-    if (total <= PRUNE_PROTECT) {
+    if (total <= protect) {
       continue;
     }
 
@@ -268,7 +283,7 @@ export function pruneToolOutputs(messages: ChatMessage[]): {
     toPrune.push(message);
   }
 
-  if (pruned <= PRUNE_MINIMUM) {
+  if (pruned <= minimum) {
     return { prunedTokens: 0 };
   }
 
@@ -283,7 +298,7 @@ export async function compactHistory(
   input: CompactHistoryInput
 ): Promise<CompactionResponse> {
   const messagesBefore = input.history.length;
-  const { prunedTokens } = pruneToolOutputs(input.history);
+  const { prunedTokens } = pruneToolOutputs(input.history, input.compaction);
   const usedTokens = estimateHistoryTokens(
     input.history,
     input.systemPrompt,


packages/agent/src/history-compaction.test.ts
```diff
@@ -2,6 +2,7 @@ import { describe, expect, test } from "bun:test";
 import type {
   ChatCompletionResult,
   ChatMessage,
+  CompactionConfig,
```
vincentri: Bug: `CompactionConfig` is not exported by `@nakama/core` — it is defined in `packages/agent/src/history-compaction.ts:63`. Bun strips type-only imports at runtime so the tests still pass, but this is a real type error (`tsc --noEmit` flags it). Import it from `./history-compaction` instead.

packages/agent/src/history-compaction.ts
```diff
@@ -260,15 +275,15 @@ export function pruneToolOutputs(messages: ChatMessage[]): {
...
+    if (total <= protect) {
       continue;
     }
 
     pruned += estimate;
     toPrune.push(message);
   }
 
-  if (pruned <= PRUNE_MINIMUM) {
+  if (pruned <= minimum) {
```
vincentri, line 286: Coverage gap: this below-minimum branch (`0 < pruned <= minimum` → return 0) has no test. The new tests cover the protect fraction on both sides and the degenerate `usable <= 0` case, but not this one. It is reachable — e.g. usable=100k (protect=50k, minimum=10k): a 45k tool result stays under protect, then an 8k one crosses it and gets pruned, but 8k <= minimum so nothing is truncated. A small-window test pinning that outcome would close the gap.

vincentri (commented):
Re-review of the PR at head c2d47ae1 against main (de670a6e). Two-axis review (standards + spec vs #342), run in a full worktree with tests executed.

**Bottom line: the fix is correct and reaches the destructive path. One small real bug (a type-only bad import in the new test), one coverage gap, one stale claim in the PR description. Nothing merge-blocking at runtime — but fix the import before merging.**

## Standards

No hard violations. Evidence: `bun test packages/agent/src/history-compaction.test.ts` → 10 pass / 0 fail; `bun x ultracite check` clean on both changed files; single commit, single concern (#342) per CONTRIBUTING.

- Judgement call (Duplicated Code): the pre-existing "prunes old tool outputs while protecting recent turns" test still builds the same 14-message transcript inline that the new `seedHistory()` helper now encodes. Migrating it would leave one shape in the file — fine as a follow-up to keep this diff minimal.
- Judgement call, no action: `protect`/`minimum` are two derived values of the same `usable × fraction` shape (possible Data Clump); too small to warrant extraction.

## Spec

Direction 1 of #342 ("scale the thresholds off contextWindow") is implemented faithfully and verified end-to-end on the path the issue describes as destructive: `resolveCompactionConfig` (real catalog `contextWindow`, 128k fallback) → `createChatSession({compaction})` → per-turn `runCompaction(false)` → `compactHistory({compaction})` → `pruneToolOutputs(history, compaction)`. All callers pass genuine configs; no scope creep; the new tests encode the issue's three-×-30k-on-a-1M-window scenario plus the protect-fraction boundary on both sides and the degenerate `usable <= 0` case.

Findings:

1. **Bug (type-only):** `history-compaction.test.ts:5` imports `CompactionConfig` from `@nakama/core`, which does not export it — it is defined in `packages/agent/src/history-compaction.ts:63`. Bun erases type-only imports so tests still pass, but `tsc --noEmit` flags it. One-line fix: import from `./history-compaction`.
2. **Coverage gap:** the `pruned <= minimum → return {prunedTokens: 0}` branch (`history-compaction.ts:286`) is untested. It is reachable (usable=100k: protect=50k, minimum=10k; a 45k result stays under protect, an 8k one crosses it and gets pruned, but 8k ≤ minimum so nothing truncates). A small-window test pinning that outcome would close it.
3. **Stale PR-body claim:** the description says the PR "threads CompactionConfig through the call path so the per-turn `runCompaction(false)` in `chat.ts` can finally see the window size", but `chat.ts` is untouched — and correctly so: `CompactHistoryInput.compaction` and the `runCompaction` pass-through already existed on main. Only `pruneToolOutputs` learned to read it. Worth correcting the description so reviewers don't hunt for a `chat.ts` change.

**Summary:** Standards — 0 hard violations, 2 judgement calls (worst: duplicated transcript fixture). Spec — 0 missing requirements, worst finding: the wrong-module type-only import.

vianbas:
I filed #342, so I re-ran the probe from that issue body against this branch,
unchanged, and then swept the model catalog. Measured on `6a9e59a3` against
`de670a6e`, every number run twice.

The fix reaches the path the issue measured. I read the description as it
stands now, after the test counts and the `chat.ts` sentence were corrected.
One thing I would act on before merge: the behavior table mixes two call
shapes, and the shape behind its 200k and 128k rows is not the one that writes
to the database.

## The issue's probe, unchanged, on this branch

All four tests now fail, each in the direction that means the bug is gone:

| test | on `de670a6e` | on `6a9e59a3` |
|---|---|---|
| one ordinary turn on a 1M window | 2 truncated | 0 |
| the same turn at 1M, 200k, 128k | 2 truncated at all three | 0 at 1M, then the test stops |
| single 40,001-token result, 1M window | truncated | survives |
| real SQLite, bytes removed | 239,867 | **-69** |

The companion values are unchanged from the issue: `usedBefore=90121`,
`usableContextTokens=991808`, `9.09%`.

The `-69` is worth spelling out. The issue reported the stored payload going
from 360,804 bytes to 120,937 on that turn. On this branch `bytesBefore` is
360,804 to the byte and `bytesAfter` is 360,873: no tool output is touched,
and the session grows by the two rows the turn itself writes, 69 bytes
together.

```json
{"content":"turn 5","role":"user"}
{"content":"ok","role":"assistant"}
```

The description's closing line says the byte count after the turn is
unchanged. It is those 69 bytes larger, for the reason above.

```bash
# in a clone of ahmadrosid/nakama
git fetch origin pull/350/head:pr350 && git checkout pr350
bun install
# paste the prune-window-probe.test.ts block from the #342 body into
# apps/server/src/services/prune-window-probe.test.ts
bun test apps/server/src/services/prune-window-probe.test.ts
```

## The 128k and 131k band is unchanged, and 128k is also the fallback

The issue's seed shape, three turns each carrying one 30,000-token tool
result, one ordinary turn sent on top, run against every `contextWindow` and
`maxOutputTokens` pair in `apps/server/src/providers/models.ts`. The count is
how many of the three tool results end up truncated:

| contextWindow / maxOutputTokens | catalog entries | `de670a6e` | `6a9e59a3` |
|---|---:|---:|---:|
| 7968 / 4096 | 1 | 0 (see below) | 3 |
| 128000 / 8192, 16384, 40960 | 5 | 2 | 2 |
| 131072 / 40960 | 9 | 2 | 2 |
| 200000 / 8192 | 2 | 2 | 0 |
| 204800 / 131072 | 4 | 2 | 0 |
| 256000 / 64000 | 1 | 2 | 0 |
| 262144 / 65536, 262144 | 10 | 2 | 0 |
| 1000000 / 8192, 384000 | 6 | 2 | 0 |

For this seed, 23 of 38 entries keep their stored transcript, 14 behave
exactly as before, and the `7968` entry is its own case, at the end.

A larger tool result still prunes on the wider windows. The boundary for a
single result moved from a flat 40,001 tokens to
`floor(usable * 0.5) + 1`. Flat is measured, not assumed: on `de670a6e` a
40,000-token result survives and a 40,001-token one is truncated at 128k,
200k, 262144 and 1M alike. With `maxOutputTokens: 8192` the new boundary is
495,905 tokens at 1M and 59,905 at 128k.

The unchanged band is also where `resolveCompactionConfig` lands when the
catalog does not know the model: `contextWindow: model?.contextWindow ??
128_000` (`apps/server/src/services/agent-service.ts:3585`).

On a 128k window the same seed still removes **239,867 bytes** from real
SQLite: `before=360804`, `after=120937`. Those are the same three numbers
`de670a6e` produces there, to the byte, so that band is unchanged in size and
not only in count.

This is not an objection. The issue offered direction 1 and this is direction
1, faithfully, and the description already shows the 128k row still
truncating. The two things I would add to it are that the same band is the
fallback for any model the catalog does not carry, and that the loss there is
larger than the table says, which is the next section.

## The behavior table mixes two call shapes

`seedHistory` returns 11 messages and the unit tests call `pruneToolOutputs`
on it directly. On the per-turn path `compactHistory` sees that array after
the new turn has been appended, so the two-turn tail slides forward and all
three tool results become candidates instead of two.

Your own seed, both shapes, both branches, each run twice:

| shape | window | `de670a6e` | `6a9e59a3` |
|---|---|---:|---:|
| direct call | 1,000,000 | 1 | 0 |
| direct call | 200,000 | 1 | 0 |
| direct call | 128,000 | 1 | 1 |
| one ordinary turn | 1,000,000 | 2 | 0 |
| one ordinary turn | 200,000 | 2 | 0 |
| one ordinary turn | 128,000 | 2 | 2 |

Against the published table: the 1,000,000 old cell reads 2 of 3, which is the
per-turn number and the one the issue reported. The 200,000 and 128,000 old
cells read 1 of 3, which is the direct-call number. So the first row is
measured one way and the rest another.

It matters because the per-turn path is the one that writes back to
`session_messages`. Put on that one shape, three cells change and the rest
stand: 200,000 old becomes 2 of 3, 128,000 old becomes 2 of 3, and 128,000 new
becomes 2 of 3. Both "no prune" cells hold under either shape.

The last two rows are a separate seed. The caption names the issue's seed, but
`50k truncated` is the existing test's single 50,000-token result; the issue's
seed gives 1 of 3 by direct call and 2 of 3 (100,000) or 0 (4,096) per turn.
The 100,000 row is labelled `(existing test)`; the 4,096 row carries no such
label.

## The 7968 row, and why its 0 is not a baseline

The `7968` row shows 0 on `de670a6e`, which looks like the small model was
already safe. It is not. There, `pruneToolOutputs` truncates two of the three
results, `prunedTokens=60000`, and then the overflow check triggers
summarization, which drops the whole head: the 14-message,
360,056-character seed becomes 5 messages and 22 characters with no tool
messages left, so the counter reads 0 because the messages are gone. On this
branch the same seed ends at 16 messages and 158 characters, three truncation
markers.
The text is lost either way, and the issue's direction 1 asked for the small
model to prune earlier, so this row is intended behavior rather than a
regression. I mention it only so the 0 is not read as a safety baseline.

Direction 2 stays open, and I agree it is a separate change. Happy to take it
as a follow-up if that is useful.

vincentri:
Thanks for running the probe and the catalog sweep — the shape mix you found is real, and I've corrected the description rather than arguing with it.

I re-measured every cell by running both implementations over the same seeds instead of deriving them by hand, and your table reproduces exactly: per-turn old is 2 of 3 at 200k and 128k (the published 1-of-3 cells were direct-call numbers), new at 128k is 2 of 3 per-turn, and both no-prune cells hold under either shape.

Changes to the description:

- The behavior table now states the shape up front and uses the per-turn path (the one that writes `session_messages`) throughout, with the direct-call numbers in a footnote.
- The last two rows are labelled as their own seeds — the degenerate row's old cell was also on the wrong seed; measured it's 2 of 3 (100k tokens), not 50k.
- The closing line now says the stored payload ends **69 bytes larger** (the two rows the turn itself writes), not unchanged.
- Added both things you asked for: the 128k band is also the `resolveCompactionConfig` fallback for uncataloged models, and the 7968 row's 0 is called out as summarization dropping the head, not a safety baseline.

On direction 2 as a follow-up: agreed it's the cleaner end state for branching; leaving that call to the maintainers.

PR #351, opened 2026-08-21
feat(tools): read another profile's sessions from within the same org

### LINKED ISSUE #332: Allow an agent to read sessions from other agents
## Problem

Each agent (profile) owns its own sessions, scoped per org. There is no way for one agent to read the conversation history of another agent in the same org. That blocks multi-agent workflows: a coordinator agent that needs to inspect what a sub-agent already discussed, or an agent taking over a task mid-flight.

## Proposed behavior

- Allow an agent to read (and list) sessions belonging to another profile in the same org.
- Read-only by default. Writing, compacting, or purging another agent's session is a separate decision.
- Keep the existing org boundary intact. Cross-org access must stay rejected (see `apps/server/src/http/routes/sessions.org-scope.test.ts`).

## Scope / open questions

- Read-only vs. also compact/purge other agents' sessions.
- Permission model: all profiles in an org by default, or opt-in per profile.
- Surfaces: HTTP sessions route, agent tools (so the agent can call it during a run), or both.

## Acceptance criteria

- A profile can fetch another profile's session messages within the same org.
- Cross-org access remains rejected.
- Covered by tests.

vianbas:
## Summary

Fixes #332.

Two read-only tools, `list_profile_sessions` and `read_profile_session`, so an assigned profile can look at another profile's sessions inside its own organization. They follow the shape you asked for: a row in `tools` with its own handler type, resolved by name, assigned per profile. Nothing is appended in `resolveProfileTools`, and `list_profiles` is untouched.

Super Bot receives both at seed time, the way `bash` does.

## The org boundary

It goes through the profile, never through the session row. `sessions` has no `org_id` column, and the optional `orgId` on `StoredSessionRecord` is never persisted: the in-memory adapter keeps it, SQLite drops it. A check written against that field would pass its own unit test and do nothing in production. Both tools call into `AgentService`, which resolves the session's profile and calls `getProfileForOrg`.

A profile in another organization fails exactly the way a profile id that does not exist fails, and a session in another organization fails exactly the way an unknown session id fails. Neither tool confirms whether an id exists somewhere else.

## Decisions I made, and where I would expect pushback

Three of the questions in my earlier comment are still open, so I picked defaults rather than stall. Each is cheap to change.

- **No extra role gate.** These tools read the rows the two sessions routes already return to every role in the org, viewers included: `list_profile_sessions` calls the same `listSessions`, and the transcript comes from `listMessagesForSession`, the same rows `GET /v1/sessions/{sessionId}/messages` returns when no turn is in flight. Both routes check the org and nothing else. A viewer denial in the tool would not protect anything, since the same user can call the route. Org memory denies viewers because its data has no such route.
- **Every profile in the org is a valid target, Super Bot included.** No special case. One consequence is worth stating here rather than leaving it to be found in the code: you said Super Bot can have the tools by default, and Super Bot already holds `list_profiles` (`apps/server/src/tools/super-bot-tools.ts:41`, appended for `isSuper` profiles at `apps/server/src/services/agent-service.ts:3159` on this branch), which returns every profile id in the organization. So Super Bot can find the ids first; it does not need one it was already given. An ordinary profile assigned these tools has no `list_profiles` and does have to know the id already. If you would rather Super Bot not have them by default, the change is the two `ensureSuperBotSessionTools` calls in `packages/db/src/org-profiles.ts`.
- **Persisted messages only.** `read_profile_session` never consults the live turn registry, so a session with a turn in flight comes back as of its last completed turn. That is the cheaper read and it has no side effects.

I also changed one default I published. I had proposed that the list verb merge all eight channels when none is given, then corrected the cost of that in a follow-up: the `sessions` filter is a scan, not a seek, so it is eight full passes. Rather than merge or make `channel` required, `list_profile_sessions` now takes an optional `channel` defaulting to `web`, which is what `GET /v1/sessions` itself does. One query, no adapter change.

## Test plan

Three new test files, sixteen tests.

- `apps/server/src/tools/session-tools.test.ts`, nine tests: a same-org list, a same-org read, paging, the channel default, an unknown channel, the org context requirement, and three refusals. Each refusal asserts that the two failures are indistinguishable rather than pinning the message text, and each carries a positive control, so a tool that failed on everything would not pass them. One refusal runs against `createSqliteDatabase(":memory:")` rather than the map-backed adapter, because that is the case where a check against the dead `orgId` field would have looked green and done nothing. Its foreign keys are live: the first version of that test failed with `SQLITE_CONSTRAINT_FOREIGNKEY` until the organization rows were created.
- `packages/db/src/seed-session-tools.test.ts`, five tests: both rows are upserted with the `session` handler type, they survive `removeUnsupportedTools`, seeding does not assign them to an ordinary profile, Super Bot gets both, and a second boot re-assigns them after an admin unassigns.
- `apps/server/src/services/tool-resolver-session.test.ts`, two tests: the `session` handler type resolves through the registry, and resolves to nothing when the tools were never registered.

Measured twice on each side, with the commands CI runs. `main` below means `de670a6e`, this branch's merge base, since that is what the diff is against; `main` has moved on since:

- `bun run test`: exit 0 on both sides, `main` 2284 pass and 0 fail across 329 files, this branch 2300 and 0 across 332. The difference is the three new files and their sixteen tests.
- `bun run knip -- --cache --reporter github-actions`: exit 0, output byte-identical to `main`. It failed once during this work, on two exported types nothing outside the file used; they are no longer exported.
- `bun run check`: no fixes on either side, 1175 files on `main` and 1179 here.

Three things CI does not check on a pull request, so I ran them here:

- **The org check is load-bearing, not decorative.** I replaced `getSessionRecordForOrg` with a plain `getSession` in `readSessionTranscript`, and the two session refusals failed, including the SQLite one. Put back, they pass.
- **`bun run build:docs` succeeds**, and the built `builtin-tools` page carries both tool names. The Docs workflow runs on pushes to `main` and on manual dispatch, never on a pull request, so a broken MDX file would land before anything caught it.
- **No new type errors.** `tsc --noEmit -p tsconfig.json` is not clean on `main` and there is no typecheck job, so the useful gate is the delta rather than the total: `main` and this branch both report 8381 error lines in my checkout, and the two sets are identical once line and column numbers are stripped. The four new files report none. The total is not portable: it moves with whatever `bun install` resolved into `node_modules`, which is how I came to publish 7851 for it earlier. The delta is the part that reproduces.

## Remaining risks

- `list_profile_sessions` inherits `HAVING COUNT(m.id) > 0` from `listSessionSummaries`, so a session with no messages is invisible to the tool. The docs section says so.
- Reading a long transcript is a large tool result, which is the kind of payload history compaction truncates. That is why the read verb pages, default 50 and capped at 200, rather than returning everything.
- The read verb pages from the start of the session, not the end, so the recent part of a long transcript takes two calls: one to learn `totalMessages`, one with the offset. The response carries `totalMessages`, so nothing else is needed, but it is a round trip a caller has to know about.
- Neither tool has been exercised against a live model, only against the service and both database adapters.

## Corrections to an earlier version of this description

Three things I had wrong. Listing them rather than editing them away, since the first one carried an argument.

- I wrote that there is no `list_profiles`, so a caller has to know the id already. `list_profiles` exists and Super Bot holds it, so that was false for the one profile these tools are seeded to by default. The second bullet above now says what actually happens.
- I described the first test file as covering "the two refusals". It has three.
- I gave 7851 as the `tsc` total on both sides. In my checkout it is 8381 on both sides. The delta was zero then and is zero now; the absolute was never portable.


apps/server/src/tools/session-tools.test.ts
```diff
@@ -0,0 +1,237 @@
...
+      "second",
+    ]);
+  });
+
+  test("refuses a profile in another org with the same error as an unknown one", async () => {
+    const { listTool } = await setUp(createInMemoryDatabaseAdapter());
+
+    await expect(
+      listTool.run({ profileId: "profile_other_org" }, context(ORG_A))
+    ).rejects.toThrow("Profile not found.");
```
vincentri: Documented-standard breach (AGENTS.md "Writing Tests: assert behavior, not prompt/description/error copy"; CONTRIBUTING.md has the same rule): this pins the literal error copy. The behavioral property you actually want — a foreign profile is indistinguishable from an unknown one — survives rewording if you assert equality instead: catch both rejections and `expect(foreignError.message).toBe(unknownError.message)`. Same applies to line 194.

apps/server/src/tools/session-tools.test.ts
```diff
@@ -0,0 +1,237 @@
...
+
+  test("rejects an unknown channel rather than silently listing web", async () => {
+    const { listTool } = await setUp(createInMemoryDatabaseAdapter());
+
+    await expect(
+      listTool.run(
+        { channel: "carrier-pigeon", profileId: "profile_target" },
+        context(ORG_A)
+      )
+    ).rejects.toThrow("Unknown channel: carrier-pigeon.");
```
vincentri: Same documented standard as the refusals below (assert behavior, not error copy): asserting the exact `Unknown channel: carrier-pigeon.` wording couples the test to copy. Asserting rejection — or a stable fragment like `Unknown channel` — keeps the behavior check without pinning the sentence.

apps/server/src/tools/session-tools.test.ts
```diff
@@ -0,0 +1,237 @@
...
+  });
+
+  test("refuses a session in another org", async () => {
+    const { foreignSessionId, readTool } = await setUp(
+      createInMemoryDatabaseAdapter()
+    );
+
+    await expect(
+      readTool.run({ sessionId: foreignSessionId }, context(ORG_A))
+    ).rejects.toThrow("Session not found.");
```
vincentri: Error-copy assertion again (here and on line 217). The refusal itself is the behavior; `rejects.toThrow()` without the literal — or a match on a stable fragment — asserts it without coupling the test to the sentence.

apps/server/src/tools/session-tools.test.ts
```diff
@@ -0,0 +1,237 @@
...
+
+    expect(allowed.totalMessages).toBe(3);
+  });
+
+  test("requires an organization context", async () => {
+    const { listTool, readTool } = await setUp(createInMemoryDatabaseAdapter());
+
+    await expect(
+      listTool.run({ profileId: "profile_target" }, context(undefined))
+    ).rejects.toThrow("Organization context is required.");
```
vincentri: Lines 232 and 235 pin "Organization context is required." verbatim — same AGENTS.md rule as the other refusals. A bare `rejects.toThrow()` preserves the intent.

apps/server/src/tools/session-tools.ts
```diff
@@ -0,0 +1,202 @@
+import type { AgentChannel, ToolContext, ToolDefinition } from "@nakama/core";
+
+const AGENT_CHANNELS: AgentChannel[] = [
```
vincentri, line 4: Judgement call (Primitive Obsession / Duplicated Code): `AGENT_CHANNELS` hand-enumerates the `AgentChannel` union as a runtime array — a third copy beside the channel validator in `apps/server/src/http/shared.ts`. If the union grows, this list drifts silently. Deriving both from one source (or reusing the existing validator) would keep them in lockstep.

apps/server/src/services/tool-resolver.ts
```diff
@@ -98,6 +103,10 @@ async function resolveStoredTool(
     return serverTools.get(record.name) ?? null;
   }
 
+  if (record.handlerType === "session") {
+    return serverTools.get(record.name) ?? null;
```
vincentri, line 100: Judgement call (Duplicated Code): this is now the fourth identical branch — `bash`, `sub_agent`, `generate_image`, and `session` all return `serverTools.get(record.name) ?? null`. A single set-membership guard over the server-tool handler types would collapse them and make the next handler kind a one-line change.

vincentri (changes requested):
Two-axis review (standards + spec vs #332), run in an isolated worktree at b1de3cd9 with tests executed.

**Bottom line: the spec is fully satisfied — all three acceptance criteria are implemented and tested, and the org boundary holds end-to-end. One documented-standard breach in the new tests (exact error-copy assertions — trivially fixable without losing intent); everything else is judgement calls.**

## Spec

All three ACs from #332 verified:

- **"A profile can fetch another profile's session messages within the same org"** — `list_profile_sessions` → `AgentService.listSessions`, `read_profile_session` → new `readSessionTranscript`; paging, channel default, and unknown-channel behavior tested.
- **"Cross-org access remains rejected"** — traced end-to-end: tools pass the caller's `orgId` → `requireProfile` / `getSessionRecordForOrg` → `db.getProfileForOrg` (`WHERE id = ? AND org_id = ?`). The claimed identical "Profile not found." for a cross-org profile and a nonexistent one is real in code and asserted by test; session reads return "Session not found." — no existence leak on either surface. `sessions.org-scope.test.ts` still passes 9/9.
- **"Covered by tests"** — 16 new tests, including the cross-org read against real SQLite (`createSqliteDatabase(":memory:")`) where the dead-`orgId` trap would have looked green on the map adapter.

The PR body's org-boundary claim checks out: `sessions` has no `org_id` column, the SQLite insert drops `StoredSessionRecord.orgId`, and neither tool ever reads an org field off a session row. Seeding claims hold: both tools in `PROTECTED_TOOL_IDS`, `"session"` added to `SUPPORTED_TOOL_HANDLER_TYPES` so they survive `removeUnsupportedTools`, Super Bot gets them in both seed branches (fresh seed and re-seed after unassign), ordinary profiles unassigned. `resolveProfileTools` untouched, no write/compact/purge path, HTTP surface unchanged.

Scope creep: one small item — `tool-source.ts` gains a `readToolSource` branch for `handlerType === "session"` so the dashboard can render the seeded rows. Not asked for by the issue, but needed for the tools UI; benign.

Observations, no action required: the tools follow plain per-profile assignment on every surface including messaging channels and automations (the issue was silent on gating; this matches the `bash` precedent, and org context is still enforced per call); `readSessionTranscript` casts persisted payloads `as ChatMessage` without validation (mirrors the pre-existing sibling in `getSessionMessages`).

## Standards

**Hard violation — error-copy assertions.** AGENTS.md ("Writing Tests: assert behavior, not prompt/description/error copy") and CONTRIBUTING.md forbid asserting exact error strings; the new test file does it 7× (`session-tools.test.ts:135, 191, 194, 204, 217, 232, 235`). The cross-org test's *intent* — foreign profile indistinguishable from unknown — is behavioral and worth keeping; assert error equality instead of the literal. Fix is mechanical; see inline comments.

Judgement calls (not merge-blocking):

- `resolveStoredTool` now has four identical branches returning `serverTools.get(record.name) ?? null` (`bash`, `sub_agent`, `generate_image`, `session`) — a set-membership guard would collapse them.
- `AGENT_CHANNELS` in `session-tools.ts` hand-enumerates the `AgentChannel` union as a runtime array — a third copy beside the validator in `http/shared.ts`.
- Docs: mostly compliant with the AGENTS.md docs rules (task-oriented, why/value/limits structure matching siblings), but cites route path `GET /v1/sessions` — the only route path in `builtin-tools.mdx` — and "re-applied on every boot" is seed-internals detail (though a useful admin warning that unassign won't stick).
- `readSessionTranscript` repeats the `getSessionRecordForOrg` → channel-parse prologue of four sibling methods (pre-existing pattern, noted for completeness).

Clean: `bun x ultracite check` on all 12 changed files; ts-taste.md cast rules respected (single boundary casts only); `SessionReader` interface exercised by tests.

**Summary:** Standards — 1 hard violation (error-copy assertions ×7), 4 judgement calls (worst: the resolver branch duplication). Spec — 0 missing/partial requirements, worst finding: none beyond the benign `readToolSource` addition.

Requesting changes solely for the test-copy breach; everything else is merge-quality.

apps/server/src/tools/session-tools.test.ts
vianbas: Done, in `ef806e5e`. This thread is pinned to the old hunk, so here is what those lines say now, `apps/server/src/tools/session-tools.test.ts:213`:

```ts
const foreign = await captureError(
  listTool.run({ profileId: "profile_other_org" }, context(ORG_A))
);
const unknown = await captureError(
  listTool.run({ profileId: "profile_does_not_exist" }, context(ORG_A))
);

expect(foreign).not.toBeNull();
expect(unknown).not.toBeNull();
expect(foreign?.message).toBe(unknown?.message);
```

`captureError` is the shape `super-bot-tools.test.ts:536` already uses. That one is a file-local function rather than an export, so this is the same helper written the same way rather than a new invention.

Your suggestion made me check something I had not. Equality on its own is weaker than the assertion it replaces: a tool that threw the same error on every call would satisfy it. I made both tools throw one error unconditionally. With the positive controls stripped back out of the three refusal tests, all three passed that mutation. So each refusal test now carries one; here it is the same list call inside the org, which still returns the seeded session. With them in place the mutation fails all nine tests in the file, which is what the string assertions did before this change.

apps/server/src/tools/session-tools.test.ts
vianbas: Done, in `ef806e5e`. The thread shows the old hunk; the assertion is now at `session-tools.test.ts:141`:

```ts
await expect(
  listTool.run(
    { channel: "carrier-pigeon", profileId: "profile_target" },
    context(ORG_A)
  )
).rejects.toThrow();
```

I took the bare rejection rather than the `Unknown channel` fragment you offered as the alternative, because a fragment is still copy, only less of it. What keeps the check meaningful is the positive control right below it: a known channel still lists the seeded session, so the rejection above is the channel check rather than a tool that fails on everything.

apps/server/src/tools/session-tools.test.ts
vianbas: Done in `ef806e5e`, both of them. Line 204 was the in-memory read and 217 the SQLite one, and both threads are frozen on the old hunks.

In-memory, now `session-tools.test.ts:241`:

```ts
const foreign = await captureError(
  readTool.run({ sessionId: foreignSessionId }, context(ORG_A))
);
const unknown = await captureError(
  readTool.run({ sessionId: "session_does_not_exist" }, context(ORG_A))
);

expect(foreign).not.toBeNull();
expect(unknown).not.toBeNull();
expect(foreign?.message).toBe(unknown?.message);
```

SQLite, now `session-tools.test.ts:271`: the same lines character for character, against `database.adapter` instead of the map adapter.

I went past the bare `rejects.toThrow()` you suggested here and asserted the two failures are equal instead, because for these two the indistinguishability is the property worth keeping, same as on the profile thread. Both carry a positive control.

The SQLite one picked up one thing you did not ask for. It was closing its database on the last line of the test, so a failed assertion leaked the handle. It is wrapped now, `session-tools.test.ts:288`:

```ts
    } finally {
      database.close();
    }
```

the way the other three tests in the repo that open `createSqliteDatabase(":memory:")` do.

apps/server/src/tools/session-tools.test.ts
vianbas: Done, in `ef806e5e`. Now `session-tools.test.ts:298`:

```ts
await expect(
  listTool.run({ profileId: "profile_target" }, context(undefined))
).rejects.toThrow();
await expect(
  readTool.run({ sessionId: webSessionId }, context(undefined))
).rejects.toThrow();
```

One change beyond the copy. The read used to pass `"session_x"`, an id that does not exist. `requireOrgId` runs before the lookup today, so that test was checking the org gate either way, but with the literal gone a bare rejection on a missing id would keep passing if the lookup ever moved ahead of the gate. It passes a real session id now, and the line below reads that same id successfully once the org context is there.

apps/server/src/tools/session-tools.ts
vianbas: Leaving this one, and one correction to the count.

It is the fourth copy rather than the third. Besides the union at `packages/core/src/contract.ts:77` and `parseChannel` at `apps/server/src/http/shared.ts:383`, there is `parseAgentChannel` at `apps/server/src/services/agent-service.ts:3641`, the one `readSessionTranscript` itself calls. Both of those are older than this PR.

Reusing the existing validator does not fit what this array is for. It is the `enum` in the tool's JSON schema, so it has to exist at runtime, and neither `parseChannel` nor `parseAgentChannel` exposes a list. `parseChannel` also throws `NakamaApiError(..., 400)`, which is not something a tool handler should be raising.

The change that would hold all four in lockstep is one exported array in `packages/core` with the union derived from it, and both parsers reading it. That edits `contract.ts` and two call sites outside this feature, which is a second concern (`CONTRIBUTING.md:115`, "One concern per PR"). Happy to open a follow-up issue for it if you want it tracked.

apps/server/src/services/tool-resolver.ts
vianbas: Leaving this one as well. I checked the count: four identical branches, and three of them (`bash`, `sub_agent`, `generate_image`) are older than this PR. A set-membership guard changes how those three resolve as well, which is a different concern from adding a fourth handler type (`CONTRIBUTING.md:115`, "One concern per PR"). I would rather not fold a refactor of three existing paths into a PR you are reviewing for one new one. Happy to open a follow-up issue for it if you want it tracked.

vianbas:
Thanks for the two-axis review. The documented-standard breach is fixed in `ef806e5e`, and I answered each of the six threads with the code as it reads now, since the four on the test file are pinned to the old hunks.

The docs point from your summary is done too. `GET /v1/sessions` is gone, and there are no route paths left in `builtin-tools.mdx` at all. While I was there I also dropped the two error sentences the page quoted verbatim. No rule covers docs prose, but pinning a user-facing page to wording the code is free to change has the same failure mode as pinning a test to it, and nothing in CI reads docs prose. Both are behaviour descriptions now.

I left the two judgement calls where they are, with the reasoning in their threads and an offer to open follow-up issues.

One more thing, in the description rather than the code. Reading the PR back against the branch turned up three things I had written wrong, so the description now ends with a section listing them. The one worth flagging here: I had claimed there is no `list_profiles`, so a caller has to know the profile id already. That is false for Super Bot, which holds `list_profiles` and receives these two tools at seed time, so that profile can find the ids first rather than needing one it was handed. The decision itself is unchanged and is the one you asked for; I would rather it be written down than found later.

CI on the merge with `main` is green: 2306 pass, 0 fail.

vincentri (approved):
Re-checked at ef806e5e. The documented-standard breach from the first review is resolved: no exact error-copy assertions remain in `session-tools.test.ts` — the cross-org and session refusals now assert error *equality* (the indistinguishability property) with positive controls so mutations fail, the channel and org-context checks use bare rejection with positive controls, and the SQLite test closes its handle in `finally`. All 16 tests across the three new files pass at this head.

The two declined judgement calls (AGENT_CHANNELS enumeration, resolver branch guard) are fair calls under CONTRIBUTING's one-concern-per-PR — both refactors would touch pre-existing paths outside this feature. Follow-up issues would be welcome.

Approving.

ahmadrosid:
Ponytail pass applied: dropped SessionReader / readSessionTranscript, reused getSessionMessages({ persistedOnly }). LGTM — merging.

PR #352, opened 2026-08-21
feat(tools): add custom Python tool handler (#331)

### LINKED ISSUE #331: Add support for custom tools that run Python code
## Summary

Custom tools currently run JavaScript/TypeScript only. Add the ability to author a custom tool whose handler runs Python code.

## Current state

Custom tools are stored with a `handler_type` + `handler_config`. The supported handler types today are:

- `builtin`
- `bash`
- `sub_agent`
- `generate_image`
- `javascript`

The `javascript` handler loads a module from `~/.nakama/tools/` (via `handlerConfig.modulePath`) and calls its `run(input, context)` export. See `apps/server/src/services/javascript-tool-loader.ts` and `apps/server/src/services/tool-source.ts`.

## Proposed change

Add a new `handler_type` of `python` so a tool can point at a Python file and expose a callable handler, following the same shape as the JavaScript path:

- `handlerConfig.modulePath` resolves to a Python file under `~/.nakama/tools/` (or equivalent).
- The module exposes a `run(input, context)` function (or an equivalent entrypoint) that returns the tool result.
- `readToolSource` returns the Python source for the editor/playground view.

This likely touches:

- DB seed/schema for the new handler type (`packages/db/src/seed.ts`, `migrate.ts`).
- A `python-tool-loader` service (mirror of `javascript-tool-loader.ts`).
- `tool-source.ts`, `tool-resolver.ts`, and `apps/server/src/http/routes/tools.ts`.
- Docs (`docs/website/content/docs/builtin-tools.mdx` and the tool-authoring docs).

## Open questions

- Python runtime: is Python assumed present on the host, or do we ship/pin a runtime? What minimum version?
- Execution model: run inline via a subprocess per invocation, or a long-lived runner? How do timeouts and cancellation map to the Python process?
- Isolation/security: custom tools already run arbitrary code; does Python need any extra sandboxing beyond what JavaScript custom tools get? See the `security` label concerns.
- `ToolContext` parity: how much of the JS `ToolContext` (workspace paths, config dir, etc.) gets passed to Python, and in what form (env vars, JSON on stdin, both)?
- Parameters: reuse `parameters` JSON schema like the JS path, or derive from a Python signature?

## Acceptance criteria

- A custom tool with `handler_type: python` can be created, assigned to a profile, and invoked end-to-end.
- Tool source is viewable in the editor/playground.
- Error/timeout handling is at least as clear as the JavaScript path.
- Docs cover authoring a Python custom tool.

vincentri:
## Summary

Custom tools previously ran JavaScript only. This adds a `python` handler type so a tool can point at a `.py` file under `~/.nakama/tools/`, expose `run(input, context)`, and be created, assigned to profiles, and invoked end-to-end — through the chat tool loop and the playground.

Closes #331.

## Contract

Nakama spawns `python3 <module>` once per call:

- Call arguments arrive on **stdin as JSON**; the tool prints **one JSON value to stdout**, which becomes the result the agent sees.
- Non-zero exit codes and non-JSON stdout are returned as error messages with stderr included.
- Calls time out after **30 s** (SIGTERM, then SIGKILL after a 5 s grace).
- The child inherits the parent environment plus `NAKAMA_WORKSPACE_ROOT` (active profile workspace) and `NAKAMA_CONFIG_DIR`.
- `spawn` receives `context.signal`, so a cancelled chat turn kills the child — same precedent as the bash tool.
- `NAKAMA_PYTHON_BIN` overrides the interpreter for hosts where `python3` is not on `PATH`.

```python
# ~/.nakama/tools/word_count.py
import json, sys

def run(input, context):
    return {"words": len(str(input.get("text", "")).split())}

if __name__ == "__main__":
    sys.stdout.write(json.dumps(run(json.loads(sys.stdin.read() or "{}"), {})))
```

## Handler registry

Dispatch previously lived as if-chains in four files (`tool-resolver`, `tool-source`, `profile-service`, `agent-service`), so every new handler type meant editing each call site — and missing one fails silently (this PR's own first draft widened the playground guards but left both bodies on the JS loader; python tools returned an import-error stub and nothing caught it).

`custom-tool-handlers.ts` now owns the mapping:

```ts
CUSTOM_TOOL_HANDLERS = {
  javascript: { extension: ".js", language: "javascript", load, resolveModulePath, validateModule },
  python:     { extension: ".py", language: "python",     load, resolveModulePath, validateModule },
}
```

All four call sites look up from the registry. Adding a future handler type (Ruby, WASM, …) is one loader module plus one registry entry — no existing-file edits. Shared helpers (`createErrorTool`, path containment, config parsing) moved to `custom-tool-shared.ts`, removing the copies that existed between the JS and Python loaders.

## What changed

| Area | Change |
|---|---|
| `python-tool-loader.ts` (new) | `loadPythonTool` / `validatePythonToolModule` / `resolvePythonModulePath`; missing-`run` fails at load time with a clear message |
| `javascript-tool-loader.ts` | Switched to shared helpers; behavior unchanged |
| `custom-tool-shared.ts` (new) | Helpers shared by all custom tool loaders |
| `custom-tool-handlers.ts` (new) | The registry |
| `tool-resolver.ts` / `tool-source.ts` / `profile-service.ts` / `agent-service.ts` | Dispatch through the registry |
| `packages/core/contract.ts` | `ToolSourceResponse.language` union gains `"python"` |
| Web | `.py` syntax highlighting; playground gates on any registered custom handler type |
| Docs | New "Custom tools" section in `builtin-tools.mdx` per the acceptance criteria |

## Design decisions (from the issue's open questions)

- **Runtime**: host `python3`; no bundled/pinned runtime. Trust posture matches the bash tool.
- **Execution**: subprocess per invocation. A long-lived runner would need its own restart/error-recovery story; not justified for v1.
- **Isolation**: same posture as JavaScript custom tools — arbitrary code by design.
- **Parameters**: reuse the `handlerConfig.parameters` JSON schema exactly like JS; no Python signature introspection.

## Why no DB migration

The issue listed `seed.ts` / `migrate.ts` as likely touch points. Not needed: `tools.handler_type` is a plain TEXT column with no CHECK constraint, so new values require no schema change.

## Known limitations

- Python tools cannot opt into `parallelSafe` — each call spawns a process. Sequential by default, documented.
- Super Bot authoring (`super-bot-tools.ts`) still validates JavaScript modules only, so Super Bot cannot author Python tools yet. Follow-up candidate.

## Tests

New coverage in `python-tool-loader.test.ts` (8 tests):

- Happy path incl. `NAKAMA_WORKSPACE_ROOT` propagation asserted through the child process
- Missing file → error tool; path traversal rejected
- Missing `run` → load-time error (clearer than a runtime JSON parse failure)
- Non-zero exit → exit code + stderr surfaced
- Non-JSON stdout → parse error + stderr surfaced
- Resolver dispatch from storage records
- **Regression test**: real `AgentService.runToolPlayground` executes a real python tool end-to-end (`ok: true`) — the test that would have caught the dispatch bug described above

Updated: `tool-source.test.ts` (+python source read), `profile-service.test.ts` (+python createTool, rejection message updated).

```
bun test packages/                                          # 804 pass, 0 fail
bun test apps/server/src/services apps/server/src/http/routes/tools.test.ts   # 424 pass, 0 fail
bun x ultracite check                                       # 1179 files clean
```

Two pre-existing failures in the full `apps/server` sweep are environmental and unrelated (bash env-var port mismatch and an Anthropic live-API connection error); both fail identically on worktrees without this change.

### DIFF
--- a/apps/server/src/services/profile-service.ts
+++ b/apps/server/src/services/profile-service.ts
@@ -55,9 +55,12 @@ import {
   ensureProfileDefaultBundledSkills,
 } from "@nakama/db";
 import {
-  loadJavascriptTool,
-  validateJavascriptToolModule,
-} from "./javascript-tool-loader";
+  CUSTOM_TOOL_HANDLERS,
+  type CustomToolType,
+  customToolTypesLabel,
+  getCustomToolHandler,
+  isCustomToolType,
+} from "./custom-tool-handlers";
 import { toMcpServerSummaries } from "./mcp-service";
 import { toSkillSummaries } from "./skills-service";
 import { readToolSource } from "./tool-source";
@@ -420,11 +423,14 @@ export class ProfileService {
     }
 
     const handlerType = readToolHandlerType(request.handlerType);
-    const handlerConfig = readJavascriptToolHandlerConfig(
+    const handlerConfig = readCustomToolHandlerConfig(
+      handlerType,
       request.handlerConfig
     );
 
-    await validateJavascriptToolModule(handlerConfig.modulePath);
+    await CUSTOM_TOOL_HANDLERS[handlerType].validateModule(
+      handlerConfig.modulePath
+    );
 
     const now = new Date().toISOString();
     const record: StoredToolRecord = {
@@ -782,7 +788,9 @@ async function enrichToolParameters(
   detail: ToolDetail,
   record?: StoredToolRecord
 ): Promise<ToolDetail> {
-  if (detail.handlerType !== "javascript") {
+  const handler = getCustomToolHandler(detail.handlerType);
+
+  if (!handler) {
     return detail;
   }
 
@@ -798,7 +806,7 @@ async function enrichToolParameters(
       updatedAt: detail.updatedAt,
     } satisfies StoredToolRecord);
 
-  const loaded = await loadJavascriptTool(source);
+  const loaded = await handler.load(source);
   if (!loaded?.parameters) {
     return detail;
   }
@@ -817,14 +825,44 @@ function toToolDetail(record: StoredToolRecord): ToolDetail {
 
 export type { ProfileDetail };
 
-function readToolHandlerType(handlerType: string | undefined): "javascript" {
-  if (handlerType === undefined || handlerType === "javascript") {
+function readToolHandlerType(handlerType: string | undefined): CustomToolType {
+  if (handlerType === undefined) {
     return "javascript";
   }
 
+  if (isCustomToolType(handlerType)) {
+    return handlerType;
+  }
+
   throw new Error(
-    'Only JavaScript tools can be created. Use handlerType "javascript".'
+    `Only ${customToolTypesLabel()} tools can be created. Use handlerType ${customToolTypesLabel()}.`
   );
+}
+
+function readCustomToolHandlerConfig(
+  handlerType: CustomToolType,
+  handlerConfig: unknown
+): { modulePath: string } {
+  const { extension } = CUSTOM_TOOL_HANDLERS[handlerType];
+
+  if (typeof handlerConfig !== "object" || handlerConfig === null) {
+    throw new Error(
+      `Custom tools require handlerConfig.modulePath ending in "${extension}".`
+    );
+  }
+
+  const modulePath = (handlerConfig as Record<string, unknown>).modulePath;
+
+  if (
+    typeof modulePath !== "string" ||
+    !modulePath.trim().endsWith(extension)
+  ) {
+    throw new Error(
+      `Custom tools require handlerConfig.modulePath ending in "${extension}".`
+    );
+  }
+
+  return { modulePath: modulePath.trim() };
 }
 
 async function writeGeneratedSoulFiles(
@@ -873,24 +911,4 @@ function validateGeneratedSoulFiles(
       throw new Error(`Soul file content must be a string: ${key}`);
     }
   }
-}
-
-function readJavascriptToolHandlerConfig(handlerConfig: unknown): {
-  modulePath: string;
-} {
-  if (typeof handlerConfig !== "object" || handlerConfig === null) {
-    throw new Error(
-      'JavaScript tools require handlerConfig.modulePath ending in ".js".'
-    );
-  }
-
-  const modulePath = (handlerConfig as Record<string, unknown>).modulePath;
-
-  if (typeof modulePath !== "string" || !modulePath.trim().endsWith(".js")) {
-    throw new Error(
-      'JavaScript tools require handlerConfig.modulePath ending in ".js".'
-    );
-  }
-
-  return { modulePath: modulePath.trim() };
 }
--- a/apps/server/src/services/profile-service.ts
+++ b/apps/server/src/services/profile-service.ts
@@ -9,6 +9,7 @@ import type {
   DeleteKnowledgeBaseResponse,
   DocumentAttachment,
   ImageAttachment,
+  JsonSchema,
   ListKnowledgeBaseResponse,
   ListProfilesResponse,
   ListToolsResponse,
@@ -842,7 +843,7 @@ function readToolHandlerType(handlerType: string | undefined): CustomToolType {
 function readCustomToolHandlerConfig(
   handlerType: CustomToolType,
   handlerConfig: unknown
-): { modulePath: string } {
+): { modulePath: string; parameters?: JsonSchema } {
   const { extension } = CUSTOM_TOOL_HANDLERS[handlerType];
 
   if (typeof handlerConfig !== "object" || handlerConfig === null) {
@@ -851,7 +852,8 @@ function readCustomToolHandlerConfig(
     );
   }
 
-  const modulePath = (handlerConfig as Record<string, unknown>).modulePath;
+  const config = handlerConfig as Record<string, unknown>;
+  const modulePath = config.modulePath;
 
   if (
     typeof modulePath !== "string" ||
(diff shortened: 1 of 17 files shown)


apps/server/src/services/profile-service.ts
```diff
@@ -817,16 +825,46 @@ function toToolDetail(record: StoredToolRecord): ToolDetail {
...
+  if (
+    typeof modulePath !== "string" ||
+    !modulePath.trim().endsWith(extension)
+  ) {
+    throw new Error(
+      `Custom tools require handlerConfig.modulePath ending in "${extension}".`
+    );
+  }
+
+  return { modulePath: modulePath.trim() };
```
ahmadrosid: This drops `handlerConfig.parameters`. Python has no module-level schema export, so the LLM gets an open object. Persist `{ modulePath, parameters }`.

apps/server/src/services/python-tool-loader.ts
```diff
@@ -0,0 +1,235 @@
...
+        clearTimeout(sigtermTimer);
+        reject(error);
+      }
+    }
+  );
+
+  const trimmed = result.stdout.trim();
+
+  if (!trimmed) {
+    return null;
```
ahmadrosid: Empty stdout is success (`null`). Nakama never calls `def run` — a harness-less module that passed create-time validation comes back `{ ok: true, result: null }`. Treat empty stdout as an error.

apps/server/src/services/python-tool-loader.ts
```diff
@@ -0,0 +1,235 @@
...
+
+      child.once("error", (error) => {
+        clearTimeout(sigtermTimer);
+        reject(error);
+      });
+
+      child.once("close", (exitCode) => {
+        clearTimeout(sigtermTimer);
+
+        if (exitCode === 0) {
```
ahmadrosid, line 204: Exit 0 after SIGTERM still succeeds. If `timedOut`, reject.

docs/website/content/docs/builtin-tools.mdx
```diff
@@ -388,6 +388,48 @@ Exports are `.zip` backups and should be handled as sensitive files because they
...
+The contract:
+
+- Nakama invokes `python3 <file>` once per call and writes the call arguments to stdin as JSON.
+- Whatever the script prints to stdout must be one JSON value — that becomes the tool result the agent sees.
+- A non-zero exit code or non-JSON stdout is returned to the agent as an error message (stderr is included).
+- Calls time out after 30 seconds.
+- The child process inherits your environment plus two extras: `NAKAMA_WORKSPACE_ROOT` (the active profile workspace) and `NAKAMA_CONFIG_DIR`.
+- Set `NAKAMA_PYTHON_BIN` before starting Nakama if your interpreter is not on `PATH` as `python3`.
+
+Parameters are declared as a JSON Schema in the tool's `handlerConfig.parameters`, the same way as JavaScript tools.
```
ahmadrosid: This promises `handlerConfig.parameters`, but `createTool` currently strips them. Also line 39 still says only JS tools can opt into `parallelSafe`.

apps/server/src/services/python-tool-loader.test.ts
```diff
@@ -0,0 +1,311 @@
...
+  });
+
+  test("returns an error tool when the module file is missing", async () => {
+    const { configDir: dir } = await setupToolsDir();
+    configDir = dir;
+
+    const tool = await loadPythonTool(makeRecord());
+    const result = (await tool!.run({}, {})) as { error: string };
+
+    expect(result.error).toMatch(/not found/i);
```
ahmadrosid: Assert the failure mode (missing file / exit code / non-JSON), not the error string.

ahmadrosid (changes requested):
Requesting changes on three P1s:

1. Persist `handlerConfig.parameters` on create
2. Empty stdout must be an error (or invoke `run()` from a harness)
3. Honor `timedOut` on exit 0

Leftover JS-only copy (not in this diff): `AGENTS.md` custom-tools table, OpenAPI `runTool` summary.

vincentri:
All three P1s are addressed in 09ffc498, plus the two inline nits and the leftover copy in a55e3f06.

**1. Persist `handlerConfig.parameters`** — `readCustomToolHandlerConfig` now validates and returns `{ modulePath, parameters }` (parameters optional, must be a plain JSON-schema object; the key is omitted entirely when absent so JS round-trips are byte-identical). Traced the full path: create → stored record → loader; no other spot reconstructed handlerConfig with only modulePath. Tests: create-with-parameters reads back through the service with both stored and resolved schema equal; create-without still works on the permissive fallback; array/string/null parameters reject creation.

**2. Empty stdout is an error** — a module that prints nothing now fails with "Python tool produced no output; it must print its JSON result to stdout" (stderr tail included) instead of resolving `{ ok: true, result: null }`.

**3. `timedOut` honored on exit 0** — the close handler rejects with a timeout error before the exit-code check, so a child that traps SIGTERM and exits 0 after the deadline can't succeed. To keep that behavior testable without a 30s test, the timeout is now readable from `NAKAMA_PYTHON_TOOL_TIMEOUT_MS` at call time (default unchanged at 30s, same idiom as `NAKAMA_PYTHON_BIN`) — say the word if you'd rather keep it hard-coded and I'll drop it.

**Inline: assert failure modes, not strings** — the loader tests now match on the failure shape (`/exit code/`, `/non-json/`, `/timed out/`, module path containment) instead of exact copy; new tests pin the silent-module and exit-0-after-timeout cases.

**Leftover JS-only copy** — builtin-tools.mdx now says parameters is an optional JSON Schema with a permissive open object when omitted (line 39 stays as-is: verified Python tools cannot opt into parallelSafe); AGENTS.md custom-tools rows cover `.py` modules, the `NAKAMA_WORKSPACE_ROOT` env var, and sequential-only execution; the playground runTool summary now reads "custom JavaScript or Python tool".

Sweep after the changes: 60 pass / 0 fail across loader/JS-loader/profile-service/tools-routes/resolver tests, ultracite clean.

ahmadrosid:
Follow-up: Super Bot Python authoring is tracked in #397.

Issue #385, opened 2026-08-21
feat(providers): direct Qwen DashScope integration (intl + CN)


vincentri:
## Summary

Add **Qwen (Alibaba DashScope)** as a direct first-party provider. DashScope runs split endpoints:

| Platform | Endpoint | Type name | Env key | Label |
|---|---|---|---|---|
| International | `https://dashscope-intl.aliyuncs.com/compatible-mode/v1` | `qwen` | `QWEN_API_KEY` | Qwen (DashScope) |
| China | `https://dashscope.aliyuncs.com/compatible-mode/v1` | `qwen_cn` | `QWEN_CN_API_KEY` | Qwen (DashScope CN) |

OpenAI-compatible mode → DeepSeek template. Today Qwen is reachable only via opencode-go.

## Capability matrix (verify against docs at implementation)

| Capability | Status |
|---|---|
| Chat vision (image input) | ✅ qwen-vl / qvq variants |
| Audio transcription (STT) | ⚠️ qwen-audio exists as a separate product line — include only if exposed through compatible-mode chat or a documented STT endpoint |
| Image generation | ⚠️ wan/flux via separate DashScope APIs — opt into Image generation settings only if a compatible endpoint is documented |

## Implementation touch points

Standard six: `ProviderName` union · `apiKeyEnvVarForProvider` · `create.ts` case ×2 · `models.ts` curated entries per region (text + qwen-vl) · `pricing.ts` · tests.

## Acceptance criteria

- [ ] `qwen` / `qwen_cn` appear as distinct cards with own key fields and base URLs
- [ ] Curated entries include text + qwen-vl vision variants with correct flags
- [ ] Chat works end-to-end with a live key on both platforms; cassettes committed
- [ ] Non-vision fallback clean on text-only entries

⚠️ Re-verify endpoints and current Qwen3-era lineup against docs during implementation.

Per CONTRIBUTING: comment to claim before starting.

akmalsyrf:
Claiming this — implementing direct Qwen DashScope (intl + CN) as first-party providers.

Issue #384, opened 2026-08-21
feat(providers): direct Moonshot Kimi integration (intl + CN)


vincentri:
## Summary

Add **Moonshot Kimi** as a direct first-party provider. Moonshot runs split platforms:

| Platform | Endpoint | Type name | Env key | Label |
|---|---|---|---|---|
| International | `https://api.moonshot.ai/v1` | `moonshot` | `MOONSHOT_API_KEY` | Moonshot Kimi |
| China | `https://api.moonshot.cn/v1` | `moonshot_cn` | `MOONSHOT_CN_API_KEY` | Moonshot Kimi (CN) |

OpenAI-compatible → DeepSeek template. Today Kimi is reachable only via Fireworks / opencode-go aggregators.

## Capability matrix (verify against docs at implementation)

| Capability | Status |
|---|---|
| Chat vision (image input) | ⚠️ CN platform exposes vision variants (`*vision` model ids); intl is text-only K2.x |
| Audio transcription (STT) | ❌ |
| Image generation | ❌ |

## Implementation touch points

Standard six: `ProviderName` union · `apiKeyEnvVarForProvider` · `create.ts` case ×2 · `models.ts` curated entries per region · `pricing.ts` · tests (`models.test.ts` + `*.llm.test.ts` cassettes).

## Acceptance criteria

- [ ] `moonshot` / `moonshot_cn` appear as distinct cards with own key fields and base URLs
- [ ] Curated entries show correct context window, pricing, vision flags
- [ ] Chat works end-to-end with a live key on both platforms; cassettes committed
- [ ] Non-vision fallback clean when attaching images to K2.x text entries

⚠️ Re-verify endpoints and current K-series lineup against docs during implementation.

Per CONTRIBUTING: comment to claim before starting.

Issue #381, opened 2026-08-21
feat(providers): direct Zhipu GLM integration (Z.ai intl + BigModel CN)


vincentri:
## Summary

Add **Zhipu GLM** as a direct first-party provider. Zhipu runs **split platforms** under two brands — Z.ai (international) and Zhipu BigModel (China) — with separate keys and catalogs, so this ships as two provider entries under one ticket:

| Platform | Endpoint | Type name | Env key | Label |
|---|---|---|---|---|
| International (Z.ai) | `https://api.z.ai/api/paas/v4` | `zhipu` | `ZHIPU_API_KEY` | GLM (Z.ai) |
| China (BigModel) | `https://open.bigmodel.cn/api/paas/v4` | `zhipu_cn` | `ZHIPU_CN_API_KEY` | GLM (CN) |

Both endpoints are OpenAI-compatible → follows the existing DeepSeek template. Today GLM is reachable only via aggregators (Z.ai catalog entry, Fireworks, opencode-go).

**Open naming question:** intl type name `zhipu` (uniform `_cn` convention, proposed) vs `zai` (matches the lab's intl brand). Maintainer preference welcome at review; rename is one line.

## Capability matrix (verify against Zhipu/Z.ai docs at implementation)

| Capability | Status |
|---|---|
| Chat vision (image input) | ✅ GLM-4V variants (separate model ids from text GLM) |
| Audio transcription (STT) | ⚠️ voice models exist but are separate products — exclude from Transcription settings unless a compatible STT endpoint is documented |
| Image generation | ✅ CogView — opt into Image generation settings |

## Implementation touch points

1. `ProviderName` union (`@nakama/core`) — add `"zhipu"`, `"zhipu_cn"`
2. `apiKeyEnvVarForProvider` (`@nakama/core`) — `ZHIPU_API_KEY`, `ZHIPU_CN_API_KEY`
3. `create.ts` cases → `createOpenAIProvider({ baseUrl, providerName })` ×2
4. `apps/server/src/providers/models.ts` — curated GLM text + GLM-4V vision entries per region with `contextWindow`, `maxOutputTokens`, USD pricing (CNY→USD conversion noted in PR), `supportsVision`
5. `pricing.ts`
6. Tests: extend `models.test.ts`; live-call tests named `*.llm.test.ts` with committed MSW cassettes

## Acceptance criteria

- [ ] `zhipu` and `zhipu_cn` appear as distinct cards, each with its own API-key field and default base URL
[...]

PR #392, opened 2026-08-21
feat(providers): direct MiniMax integration with dynamic model discovery (#380)

### LINKED ISSUE #380: feat(providers): direct MiniMax integration (intl + CN)
## Summary

Add **MiniMax** as a direct first-party provider. MiniMax runs **split platforms** — separate keys, model availability, and pricing per region — so this ships as two provider entries under one ticket:

|---|---|---|---|---|
| International | `https://api.minimax.io/v1` ✅ confirmed | `minimax` | `MINIMAX_API_KEY` | MiniMax |
| China | `https://api.minimaxi.com/v1` ✅ confirmed | `minimax_cn` | `MINIMAX_CN_API_KEY` | MiniMax (CN) |

Both endpoints are OpenAI-compatible → follows the existing DeepSeek template (`createOpenAIProvider` with a base URL). Today MiniMax models are reachable only via the opencode-go aggregator.

## Capability matrix (verify against MiniMax docs at implementation)

| Capability | Status |
|---|---|
| Chat vision (image input) | ❌ M2.x series is text-only |
| Audio transcription (STT) | ❌ T2A is speech *output* only — exclude from Transcription settings |
| Image generation | ✅ image-01 — opt into Image generation settings |

1. `ProviderName` union (`@nakama/core`) — add `"minimax"`, `"minimax_cn"`
2. `apiKeyEnvVarForProvider` (`@nakama/core`) — `MINIMAX_API_KEY`, `MINIMAX_CN_API_KEY`
3. `create.ts` cases → `createOpenAIProvider({ baseUrl, providerName })` ×2
4. `apps/server/src/providers/models.ts` — curated M2.x entries per region (catalogs may differ between platforms) with `contextWindow`, `maxOutputTokens`, USD pricing (CNY→USD conversion noted in PR), `supportsVision: false`
5. `pricing.ts`
6. Tests: extend `models.test.ts`; live-call tests named `*.llm.test.ts` with committed MSW cassettes

## Acceptance criteria

- [ ] `minimax` and `minimax_cn` appear as distinct cards, each with its own API-key field and default base URL
- [ ] Curated entries show correct context window, pricing, and vision flags
- [ ] Sending an image to a MiniMax entry produces the clean non-vision fallback, not an error
- [ ] image-01 available via Image generation settings; no Transcription entry

⚠️ Both URLs are reporter-confirmed; re-verify model lineups per region against current docs during implementation.

vincentri:
## Summary

Adds **MiniMax** as a direct first-party provider — the user brings their own MiniMax API key. MiniMax runs split international/China platforms with separate keys and catalogs, so this ships as two provider entries:

|---|---|---|---|---|
| International | `https://api.minimax.io/v1` ✅ confirmed | `minimax` | `MINIMAX_API_KEY` | MiniMax |
| China | `https://api.minimaxi.com/v1` ✅ confirmed | `minimax_cn` | `MINIMAX_CN_API_KEY` | MiniMax (CN) |

Both are OpenAI-compatible; each follows the DeepSeek template (`createOpenAIProvider` with a region base URL).

Closes #380 (chat integration + region entries; image-01 deferred — see below).

## Dynamic model discovery — no hardcoded catalog

MiniMax model lists are **fetched live**, not hardcoded:

1. Adding a MiniMax instance prefills the region base URL (shared constant, below)
2. The settings card gets the same remote-browse editing flow as OpenAI-compatible endpoints — Browse hits `{baseUrl}/models`
3. Fetched models are stored as instance `customModels`; resolution, defaults, and vision flags derive from that list
4. New MiniMax releases appear without code changes

Region constants live in `@nakama/core/minimax-provider-config` so the server factory and web prefill share one source (mirrors `ollama-provider-config`).

## Region-safe model resolution (bug found by tests)

The first draft hardcoded six catalog entries and hit a real bug: `resolveModel` validated ids via **global** `getModelById`, which returns the first match. Region variants expose identical ids (`MiniMax-M3` on both platforms — the API must receive the lab's real id), so `minimax_cn` rejected its own models and silently fell back to the default.

Fix: validation is now **provider-scoped** (`getModelsForProvider(provider).some(...)`), which also generalizes to every future region-split provider on the board (#381 Zhipu, #384 Moonshot, #385 Qwen). The dynamic-model pivot then made the static catalog unnecessary entirely.

## Capability handling

| Capability | Status |
|---|---|
| Chat vision | Opt-in via discovered metadata (`supportsVision` on custom models); defaults to false — M2.x/M3 chat is text-only |
| Audio transcription | Not offered — MiniMax T2A is speech *output* only |
| Document input | Empty native sets (text-only), matching the deepseek precedent |
| Image generation | **Deferred** — see below |

### Deferred: image-01

Image generation is hard-allowlisted to `openai::gpt-image-2` in `agent-service.ts` (`isAllowedImageGenerationSelection`). Wiring MiniMax image-01 requires a new image-API call shape plus allowlist/settings extension — a follow-up PR referencing #380 rather than scope creep here.

## Wiring checklist

- `ProviderName` union += `"minimax"`, `"minimax_cn"` (`contract.ts`)
- `apiKeyEnvVarForProvider` → `MINIMAX_API_KEY` / `MINIMAX_CN_API_KEY`
- `create.ts` cases ×2 with region base URLs
- Labels across `user-config.ts`, `provider-label.ts`, `provider-setup-prompt.ts`
- `document-content.ts` empty native-doc sets
- Web: `PROVIDER_OPTIONS` cards, label chain, compatibleLike editing with region URL prefill
- New `packages/core/src/minimax-provider-config.ts`

## Tests

`models.test.ts`:
- customModels passthrough for discovered ids; unknown id falls back to instance default
- `getDefaultModel` resolves per-instance default for both regions
- Vision opt-in via discovered metadata (false by default, true when flagged)

`provider-resolution.test.ts`:
- `parseProviderName` accepts both types
- Distinct env keys per region

```
bun test packages/ + models.test + web models.test   # 847 pass, 0 fail
bun x ultracite check                                # clean
bun run knip                                         # clean
tsc error count                                      # identical to untouched-worktree baseline
```

## Live-key caveat

Discovered ids come straight from MiniMax's `/models` response, so context-window and cost metadata for usage tracking are only as good as that endpoint reports (same tradeoff openai_compatible users have today). Exact direct-API model ids should be spot-checked against platform.minimax.io with a live key before release — both base URLs themselves are reporter-confirmed.

### DIFF
--- a/apps/server/src/providers/create.ts
+++ b/apps/server/src/providers/create.ts
@@ -8,6 +8,10 @@ import {
   readEnvValue,
   type UserConfig,
 } from "@nakama/core";
+import {
+  MINIMAX_CN_DEFAULT_BASE_URL,
+  MINIMAX_DEFAULT_BASE_URL,
+} from "@nakama/core/minimax-provider-config";
 import { resolveDefaultModelForInstance } from "../services/provider-instance-helpers";
 import { createAnthropicProvider } from "./anthropic";
 import { createCerebrasProvider } from "./cerebras";
@@ -72,6 +76,20 @@ function createProvider(options: CreateProviderOptions): ProviderClient {
         baseUrl: baseUrlOverride ?? DEFAULT_DEEPSEEK_BASE_URL,
         model,
         providerName: "deepseek",
+      });
+    case "minimax":
+      return createOpenAIProvider({
+        apiKey: options.apiKey,
+        baseUrl: baseUrlOverride ?? MINIMAX_DEFAULT_BASE_URL,
+        model,
+        providerName: "minimax",
+      });
+    case "minimax_cn":
+      return createOpenAIProvider({
+        apiKey: options.apiKey,
+        baseUrl: baseUrlOverride ?? MINIMAX_CN_DEFAULT_BASE_URL,
+        model,
+        providerName: "minimax_cn",
       });
     case "opencode_go":
       return createOpenCodeGoProvider({
--- a/apps/server/src/providers/models.test.ts
+++ b/apps/server/src/providers/models.test.ts
@@ -142,9 +142,40 @@ describe("resolveModel", () => {
       "accounts/fireworks/models/glm-5p2"
     );
   });
+
+  test("resolves MiniMax models from discovered custom models", () => {
+    const customModels = [
+      { default: true, id: "MiniMax-M3" },
+      { id: "MiniMax-M2.7" },
+      { id: "MiniMax-M2.5" },
+    ];
+
+    expect(resolveModel("minimax", "MiniMax-M2.7", customModels)).toBe(
+      "MiniMax-M2.7"
+    );
+    expect(getDefaultModel("minimax", customModels)).toBe("MiniMax-M3");
+    expect(getDefaultModel("minimax_cn", customModels)).toBe("MiniMax-M3");
+  });
+
+  test("falls back to instance default for unknown MiniMax ids", () => {
+    const customModels = [{ default: true, id: "MiniMax-M3" }];
+    expect(resolveModel("minimax_cn", "not-a-real-model", customModels)).toBe(
+      "MiniMax-M3"
+    );
+  });
 });
 
 describe("modelSupportsVision", () => {
+  test("keeps MiniMax models opt-in only (discovered lists)", () => {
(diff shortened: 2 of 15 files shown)


ahmadrosid (approved):
nice, thanks for adding more provider here 👍

PR #393, opened 2026-08-21
feat(orgs): let platform admins archive an organization

ahmadrosid:
## Summary

Platform admins can archive the current org from **System → Organization**. Members, profiles, and on-disk files stay; Nakama cannot restore it.

Org admins cannot archive. The last live org, and a platform admin’s last remaining membership, return 409. Archived orgs drop out of the switcher, org APIs, automations, curator, and notify webhooks. Stale Telegram/Discord/WhatsApp orgs prompt instead of silently rebinding.

## Test plan

- [ ] Platform admin with two orgs: Delete → confirm; switcher lands on the remaining org
- [ ] Org admin does not see Delete
- [ ] Last remaining org (and last remaining membership) returns 409
- [ ] After archive, that org is 404 on org-scoped routes; automations, curator, and `/v1/notify` skip it
- [ ] Telegram/Discord/WhatsApp with a stored archived org prompts instead of hopping tenants

---

Cursor

### DIFF
--- a/apps/server/src/http/org-middleware.ts
+++ b/apps/server/src/http/org-middleware.ts
@@ -74,6 +74,12 @@ export function createOrgContextMiddleware(
       return;
     }
 
+    const organization = await databaseAdapter.getOrganizationById(orgId);
+    if (!organization || organization.archivedAt) {
+      c.res = errorResponse("Not found", 404);
+      return;
+    }
+
     const member = await databaseAdapter.getOrgMember(orgId, auth.user.id);
     if (!member) {
       c.res = errorResponse("Not found", 404);
--- a/apps/server/src/services/notification-webhook-service.ts
+++ b/apps/server/src/services/notification-webhook-service.ts
@@ -60,6 +60,13 @@ export class NotificationWebhookService {
       throw new NakamaApiError("Invalid notification credentials.", 401);
     }
 
+    const organization = await this.databaseAdapter.getOrganizationById(
+      destination.orgId
+    );
+    if (!organization || organization.archivedAt) {
+      throw new NakamaApiError("Not found", 404);
+    }
+
     const normalized = normalizeNotificationWebhookRequest(payload);
     const result = await this.telegram.send({
       chatIds: [destination.config.chatId],
--- a/apps/server/src/services/org-service.test.ts
+++ b/apps/server/src/services/org-service.test.ts
@@ -489,6 +489,47 @@ describe("OrgService", () => {
     expect(resolved).toBe(second.organization.id);
   });
 
+  test("clears a stale session org when the user has no remaining memberships", async () => {
+    const { orgService, authService, databaseAdapter } = createOrgService();
+    const bootstrapped = await orgService.bootstrapInitialSetup({
+      admin: {
+        email: "admin@acme.com",
+        name: "Acme Admin",
+        passwordHash: await authService.hashPassword("password123"),
+        phone: "",
+      },
+      organization: { name: "Acme", slug: "acme-session-clear" },
+    });
+    await databaseAdapter.createBrowserSession({
+      activeOrgId: bootstrapped.organization.id,
+      createdAt: new Date().toISOString(),
+      csrfTokenHash: "csrf",
+      expiresAt: "2099-01-01T00:00:00.000Z",
+      id: "session_stale",
+      lastUsedAt: null,
+      revokedAt: null,
+      sessionTokenHash: "token_stale",
+      userId: bootstrapped.user.id,
+    });
+    await databaseAdapter.upsertOrganization({
+      ...(await databaseAdapter.getOrganizationById(
+        bootstrapped.organization.id
+      ))!,
+      archivedAt: new Date().toISOString(),
+    });
+
+    const resolved = await orgService.resolveActiveOrgId(
+      bootstrapped.user.id,
+      "session_stale",
+      bootstrapped.organization.id
+    );
+    expect(resolved).toBeNull();
+
+    const session =
+      await databaseAdapter.getBrowserSessionBySessionTokenHash("token_stale");
+    expect(session?.activeOrgId).toBeNull();
+  });
+
   test("refuses to archive the actor's last remaining membership", async () => {
     const { orgService, authService } = createOrgService();
     const bootstrapped = await orgService.bootstrapInitialSetup({
@@ -510,7 +551,10 @@ describe("OrgService", () => {
         bootstrapped.organization.id,
         bootstrapped.user.id
       )
-    ).rejects.toMatchObject({ status: 409 });
+    ).rejects.toMatchObject({
+      message: "Cannot archive your last remaining organization.",
+      status: 409,
+    });
   });
 
   test("refuses updates on an archived org", async () => {
@@ -549,7 +593,10 @@ describe("OrgService", () => {
 
     await expect(
       orgService.archiveOrganization(created.organization.id)
-    ).rejects.toMatchObject({ status: 409 });
+    ).rejects.toMatchObject({
+      message: "Cannot archive the last remaining organization.",
+      status: 409,
+    });
 
     const stillListed = await orgService.listOrganizations();
     expect(stillListed).toHaveLength(1);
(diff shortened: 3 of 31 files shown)


apps/server/src/http/org-middleware.ts
```diff
@@ -74,6 +74,12 @@ export function createOrgContextMiddleware(
       return;
     }
 
+    const organization = await databaseAdapter.getOrganizationById(orgId);
+    if (!organization || organization.archivedAt) {
```
vincentri, line 78: Judgement call (Duplicated Code / mild Shotgun Surgery): the "org missing or archived → 404" predicate is now written inline in five places — here, internal-automations.ts (schedule set + run guard), internal-curator.ts, notification-webhook-service.ts — while OrgService already extracted `requireActiveOrganization()`. Gathering them behind one helper would make the archived rule single-sourced for the next surface.

apps/server/src/services/org-service.test.ts
```diff
@@ -443,4 +443,225 @@ describe("OrgService", () => {
...
+      slug: "other-last-membership",
+    });
+
+    await expect(
+      orgService.archiveOrganization(
+        bootstrapped.organization.id,
+        bootstrapped.user.id
+      )
+    ).rejects.toMatchObject({
+      message: "Cannot archive your last remaining organization.",
```
vincentri: Documented-standard breach (AGENTS.md "Writing Tests: assert behavior, not prompt/description/error copy"; CONTRIBUTING.md §Testing): lines 555 and 597 pin the exact error strings. The 409 status alone asserts the behavior — the two cases are already distinguished by which call was made. Same for the exact `message` field if you want to keep some discrimination: assert on status, not copy.

apps/web/src/components/settings/OrgArchiveCard.tsx
```diff
@@ -0,0 +1,102 @@
...
+      setFormError(formatError(error));
+    } finally {
+      setPending(false);
+    }
+  }
+
+  return (
+    <Card className="gap-0 py-0">
+      <div className="flex flex-wrap items-center justify-between gap-3 px-4 py-4 sm:px-5">
+        <h2 className="font-medium text-sm">Delete organization</h2>
```
vincentri, line 45: Judgement call (Mysterious Name / language drift): one concept carries three user-facing names — OpenAPI says "Archive an organization" over an HTTP DELETE, this UI says "Delete organization", and the docs heading says "Remove an organization". Integrators reading the API docs and admins reading the UI will wonder if they are different operations. Worth picking one term across API/UI/docs before merge or in a fast follow-up.

apps/server/src/services/notification-webhook-service.ts
```diff
@@ -60,6 +60,13 @@ export class NotificationWebhookService {
       throw new NakamaApiError("Invalid notification credentials.", 401);
     }
 
+    const organization = await this.databaseAdapter.getOrganizationById(
+      destination.orgId
+    );
+    if (!organization || organization.archivedAt) {
```
vincentri, line 66: Judgement call (Feature Envy, weak): this reaches into `databaseAdapter.getOrganizationById` directly and re-implements the archived guard, where OrgService already owns that rule (`requireActiveOrganization`). Going through the service would keep the archived semantics in one place — same theme as my comment on org-middleware.ts.

apps/web/src/lib/org-archive.ts
```diff
@@ -0,0 +1,10 @@
+export function canArchiveOrganization(isPlatformAdmin: boolean): boolean {
```
vincentri, line 1: Judgement call (Middle Man / Speculative Generality): `canArchiveOrganization(isPlatformAdmin) { return isPlatformAdmin; }` returns its argument — the caller already knows the answer. Inline the check, or give the function real logic (e.g. org-admin allowance) when that lands.

vincentri (changes requested):
Two-axis review (standards + spec vs the PR description, which is its own spec — no linked issue), run in an isolated worktree at 900b1a01 with tests executed.

**Bottom line: every behavioral claim in the description verifies line-by-line — platform-admin-only guard (403 for org admins), both 409 paths, hide-not-wipe across membership lists / org APIs / automations / curator / notify webhooks, X-Org-Id → 404 with session self-heal, channel stale-org prompt instead of silent rebind. The last-org archive really is one conditional UPDATE, race-safe. One documented-standard breach in the new tests (exact error-copy assertions); the rest is judgement calls.**

## Spec

All eight claims from the description verified in code and tests (76 pass / 0 fail across the eight touched test files):

- **"Org admins do not get this control"** — `DELETE /v1/platform/orgs/:orgId` sits behind `requirePlatformAdminFromContext`; test asserts org admin gets 403.
- **"The last live organization cannot be removed… Last-org is one write"** — the sqlite adapter archives via a single `UPDATE … WHERE archived_at IS NULL AND (SELECT COUNT(*) …) > 1` with success = `changes > 0`; statement-level atomicity makes concurrent deletes race-safe; the in-memory adapter mirrors the semantics.
- **"Hide, do not wipe"** — `archived_at` added by a PRAGMA-guarded migration consistent with the existing pattern; `schema.sql` in step; members/profiles/files untouched; platform GET still returns archived rows; automations (schedule filter + run guard), curator (service filter + route), notify webhooks (404) all exclude archived orgs.
- **"After archive, X-Org-Id on that org is 404"** — middleware checks `archivedAt` before membership → 404; switching to an archived org 404s; accept-invite requires an active org; session self-heals to a remaining membership or clears.
- **Channels prompt instead of silently rebinding** — `channel-org.ts` prompts when the stored org is gone and one other remains, with a text-pick escape hatch; covered by test.
- **Web** — `OrgArchiveCard` on System → Organization, platform-admin-gated; archiving refetches and switches to a remaining membership.

Minor notes, no action required:

- The **last-membership** check in `archiveOrganization` is check-then-write rather than atomic — two concurrent archives of different orgs by the same user can both pass it. The atomic UPDATE backstops data loss; worst case is a user momentarily at zero memberships, which `resolveActiveOrgId` handles by clearing the session. Acceptable, but worth knowing it's the one non-atomic edge.
- No test exercises the `ALTER TABLE … ADD COLUMN archived_at` upgrade on a pre-column database; the migration follows the existing guarded pattern, so risk is low.
- Scope creep: none of consequence — no restore endpoint, no hard delete, guarding update/invite against archived orgs falls under "org-scoped APIs treat the org as gone".

## Standards

**Hard violation — error-copy assertions.** `org-service.test.ts:555` and `:597` pin exact error strings ("Cannot archive your last remaining organization." / "Cannot archive the last remaining organization."), against AGENTS.md ("assert behavior, not prompt/description/error copy") and CONTRIBUTING.md §Testing. The 409 status already asserts the behavior; drop the copy match. (File precedent does the same — still against the documented standard.)

Judgement calls:

- **Duplicated Code:** the "org missing or archived → 404" predicate is inline in five places while `requireActiveOrganization()` already exists — gather behind one helper.
- **Naming drift:** one concept, three names — API "Archive" over HTTP DELETE, UI "Delete", docs "Remove". Pick one term for users/integrators.
- **Middle Man:** `canArchiveOrganization(isPlatformAdmin)` returns its argument — inline it or give it real logic later.
- **Feature Envy (weak):** notification-webhook-service reaches into the adapter directly instead of OrgService for the archived guard.

Clean: ultracite on all 30 files; ts-taste.md respected (single interop casts only); OrgArchiveCard follows the React rules (no subtitle under heading, justified destructive-confirm copy, hooks before early return); docs use dashboard navigation names with why→how structure; adapters semantically consistent.

**Summary:** Standards — 1 hard violation (error-copy assertions ×2), 4 judgement calls (worst: the five-site archived predicate). Spec — 0 missing/partial requirements, worst finding: the non-atomic last-membership edge, judged acceptable.

Requesting changes solely for the test-copy breach; everything else is merge-quality.

apps/server/src/services/org-service.test.ts
ahmadrosid: > The 409 status alone asserts the behavior — the two cases are already distinguished by which call was made.

Addressed: both last-membership and last-org tests now assert `{ status: 409 }` only. Last-org still checks the org stays listed and unarchived.

ahmadrosid:
> Requesting changes solely for the test-copy breach

Addressed in 3fe9d995 — the two 409 archive tests now assert status only, not error copy.

PR #400, opened 2026-08-21
fix(auth): require platform admin to rotate local auth token (#370)

### LINKED ISSUE #370: Security: require platform-admin for local auth token rotation
Found during the enterprise-compliance audit (2026-08-21); verified against source.

# Summary

`POST /v1/auth/local-token/rotate` requires only *any* browser session — there is no platform-admin guard. Any signed-in user, including a **viewer**, can rotate the installation-wide `tc_local_` token.

# Impact

- Availability sabotage: rotation invalidates the shared credential used by the CLI and all channel workers (Telegram/Discord/WhatsApp bridges drop until an admin re-syncs the token).
- Least-privilege violation: the lowest-privilege role can mutate an installation-wide credential.

# Evidence

- Route handler: `apps/server/src/http/routes/auth.ts:593-626` — calls `getRequestAuth(c)` and asserts CSRF, but no `requirePlatformAdminFromContext` / role check.
- The token is installation-wide: `packages/core/src/local-auth.ts` (single `tc_local_` credential owned by synthetic `local-client@nakama.internal`, admin in every org).
- Contrast: sibling sensitive routes consistently use platform-admin guards.

# Fix

1. Add `requirePlatformAdminFromContext(c)` at the top of the rotate handler.
2. Add a test asserting viewer/member get 403 on the rotate route.

Trivial fix — good first PR.

akmalsyrf:
## Summary

`POST /v1/auth/local-token/rotate` previously accepted **any** browser session. Org members and viewers could rotate the installation-wide `tc_local_` credential used by the CLI and channel workers (Telegram / Discord / WhatsApp), causing availability sabotage and violating least privilege.

This PR adds `requirePlatformAdmin(auth)` after the browser-session check, matching other sensitive install-scoped routes.

Closes #370

## Changes

- Guard `POST /v1/auth/local-token/rotate` with `requirePlatformAdmin`
- Keep existing constraints: must be browser-session auth (not bearer) + CSRF
- Regression test: org **member** browser session gets `403 Forbidden`; platform-admin setup session still rotates successfully

## Why it matters

The local token is a shared install credential (synthetic `local-client@nakama.internal`). Rotation invalidates every worker/CLI that still holds the old token until an admin re-syncs. That must not be reachable from the lowest-privilege dashboard roles.

## Test plan

- [x] `bun test apps/server/src/http/app.test.ts -t "local auth token"` — rotate as setup admin (200), reject bearer (403), reject member browser session (403)
- [x] Manual: sign in as platform admin → rotate local token succeeds; sign in as org member/viewer → rotate returns 403
- [x] Confirm CLI/workers that already hold the old token fail until they pick up the new one (expected after a successful admin rotate)

### DIFF
--- a/apps/server/src/http/app.test.ts
+++ b/apps/server/src/http/app.test.ts
@@ -549,6 +549,71 @@ describe("createHonoApp", () => {
     }
   });
 
+  test("rejects local auth token rotation for non-platform-admin browser sessions", async () => {
+    const configDir = await mkdtemp(
+      join(tmpdir(), "nakama-rotate-auth-member-")
+    );
+    process.env.NAKAMA_CONFIG_DIR = configDir;
+
+    try {
+      const options = createServerOptions();
+      const app = createHonoApp(options);
+      const setupResponse = await app.fetch(
+        new Request("http://localhost:4310/v1/auth/setup", {
+          body: JSON.stringify(
+            buildSetupAuthBody("admin@example.com", {
+              admin: { password: "secret123" },
+            })
+          ),
+          headers: { "Content-Type": "application/json" },
+          method: "POST",
+        })
+      );
+      expect(setupResponse.status).toBe(201);
+      const orgId = await seedOrgForUser(
+        options.databaseAdapter,
+        "admin@example.com"
+      );
+
+      const now = new Date().toISOString();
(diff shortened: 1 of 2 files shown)


ahmadrosid (approved):
nice

Issue #428, opened 2026-08-22
Security: SQLite DB and data directory are created world-readable (0644/0755)


vincentri:
## Context

Found in security & compliance review. Everything else under `~/.nakama` follows a strict private-permission convention — the database, which holds the most sensitive data, does not.

## Current behavior

- `ensureDatabaseDirectory()` uses bare `mkdirSync(dir, { recursive: true })` (`packages/db/src/database-url.ts:38`) → ancestor dirs land at umask default (~0755).
- `new Database(path, { create: true })` passes no file mode (`packages/db/src/adapters/sqlite.ts:407`) → the SQLite file is typically **0644 world-readable**.
- The project's own convention everywhere else is 0700 dirs / 0600 files, enforced on every write via `PRIVATE_DIR_MODE` / `PRIVATE_FILE_MODE` (`packages/core/src/fs.ts:13-14`). Config INIs, attachments, channel configs, WhatsApp session store all comply. The DB does not.

This is about restoring the perms baseline, not a substitute for encryption-at-rest (tracked separately in #364).

## Why it matters

The DB file contains plaintext MCP credentials (`mcp_servers.config`, `packages/db/sql/schema.sql:178`), all chat transcripts (`session_messages.payload`), org memory, and task outputs. Any other user or process on the host can read the entire installation. Docker's non-root uid 1000 mitigates somewhat; bare-metal and multi-user hosts are exposed.

## Proposed fix

- Apply 0700 to the data/sqlite directories and 0600 to the DB file (and any journal/WAL sidecar files) at creation, and idempotently re-tighten on startup — same pattern as `writeTextFile`, which re-chmods pre-existing loose files.

## Acceptance criteria

- [x] Fresh init produces 0600 `.sqlite` (+ sidecar files) and 0700 ancestor dirs on POSIX
- [x] Startup tightens pre-existing loose perms (test: create 0644 file, open adapter, assert 0600)
- [x] Perms assertions skipped/waived on Windows

fajarhide:
Ticked against #452, which is open and not merged.

Each box was checked on a real server start against a throwaway `NAKAMA_CONFIG_DIR`, not read off the diff:

- Fresh init: `-rw-------` on `nakama.sqlite`, `drwx------` on both `data` and `data/sqlite`.
- Pre-existing loose install: a config dir created by `8cc04d7b` at 0644 under 0755, started once on the branch, came back 0600 under 0700.
[...]

PR #438, opened 2026-08-22
fix(server): require an org-scoped automation id on the internal run endpoint

### LINKED ISSUE #434: fix(server): internal automation run endpoint executes any automation id with no org scope
## Summary

On `POST /v1/internal/automations/:automationId/run`, the only authorization is `auth.mode === "local-token"`. The automation is then fetched by id alone — `automationService.get(automationId)` — and run. Nothing validates that the caller may act on the org that owns that automation: a request carrying any live automation id and the install's local token executes it.

That is a single-layer defense for a cross-tenant action. Running an automation is not a read: it invokes that org's agent with its tools (bash, email, web), on demand. The public twin `POST /v1/automations/:id/run` resolves the automation through `automationService.get(automationId, orgId)`; the internal route omits the second argument even though the service has supported it since #321.

The local token is a single install-wide secret (`~/.nakama/local-auth-token`, or `NAKAMA_LOCAL_AUTH_TOKEN`), rotated only by platform admins through the dashboard (`auth.ts:619`). Browser-session callers get a 401 — so this is not reachable from the dashboard today. What the route is missing is the ownership assertion its public twin performs on every call. To be explicit about what that does and does not buy: it is a **fail-closed ownership assertion** — a bare id returns 400, a mismatched id/org pair returns 404 — not containment of a leaked token. The same token can read `GET /v1/internal/automations/schedules`, which returns id+orgId pairs for every schedule, and replay them into run. Containment under a leak lives in the file-permission work (#428/#429) and org-bound token principals — both follow-ups, not this fix.

Measured on `25c6bf24`.

## Symptoms

Browser-session callers are correctly rejected (this is the control, captured from a running dev server):

```
GET /v1/internal/automations/schedules
→ HTTP 401 {"error":"Authentication required"}
```

The handler has no org input at all — no query param, no body, no org claim on the token:

```ts
// apps/server/src/http/routes/internal-automations.ts:61-81 (@ 25c6bf24)
app.post("/v1/internal/automations/:automationId/run", async (c) => {
  const auth = c.get("auth");
  if (!auth || auth.mode !== "local-token") {
    return errorResponse("Authentication required", 401);
  }

  const automationId = decodeURIComponent(c.req.param("automationId"));
  const automation = await automationService.get(automationId);   // ← no orgId
  ...
  const result = await agent.runAutomation(automationId);
```

The one org-related check on this route — added upstream recently — reads the automation's own `orgId` and rejects archived orgs (`:74-79`). It validates the *target's* state, never the *caller's* scope, so it does not close the gap.

## Reproduction

Save as `apps/server/src/http/routes/internal-automation-org-scope.test.ts` and run `bun test apps/server/src/http/routes/internal-automation-org-scope.test.ts`:

```ts
import { loadLocalAuthToken } from "@nakama/core";
import { createInMemoryDatabaseAdapter } from "@nakama/db";
import { AuthService } from "../../services/auth-service";
import { AutomationService } from "../../services/automation-service";
import { OrgService } from "../../services/org-service";
import { createHonoApp } from "../app";
import { seedLocalClientUser } from "../test-org-helpers";

const ORG_A = "org_default";
const ORG_B = "org_other";

describe("internal automation org scope", () => {
  test("a local-token run under the wrong org id is refused", async () => {
    const runCalls: string[] = [];
    const agent = {
      providerConfigured: true,
      runAutomation: async (automationId: string) => {
        runCalls.push(automationId);
        return { skipped: false };
      },
    } as any;
    const automationService = new AutomationService(databaseAdapter, {
      getUserTimezone: async () => "UTC",
    });
    const orgService = new OrgService(
      databaseAdapter,
      new AuthService()
    );

    const now = new Date().toISOString();
    for (const [orgId, slug] of [
      [ORG_A, "org-a"],
      [ORG_B, "org-b"],
    ] as const) {
      await databaseAdapter.upsertOrganization({
        createdAt: now, id: orgId, name: slug, slug, updatedAt: now,
      });
      await databaseAdapter.upsertProfile({
        createdAt: now, id: `profile_${slug}`, isDefault: true, isSuper: false,
        model: null, name: slug, orgId, systemPrompt: "", updatedAt: now,
      });
    }
    await seedLocalClientUser(databaseAdapter);

    const orgBAutomation = await automationService.create(
      ORG_B,
      {
        description: "Ping",
        name: "Hourly",
        prompt: "Ping",
        trigger: { cron: "0 * * * *", timezone: "UTC", type: "schedule" },
      },
      "profile_org-b"
    );

    const app = createHonoApp({
      agent, authService: new AuthService(), automationService,
      databaseAdapter, mcpService: {} as any, orgService,
      systemStatus: {} as any, taskService: {} as any,
      webDistDir: null, workerManager: {} as any,
    });
    const token = await loadLocalAuthToken();

    const response = await app.fetch(
      new Request(
        `http://localhost:4310/v1/internal/automations/${encodeURIComponent(orgBAutomation.id)}/run?orgId=${ORG_A}`,
        { headers: { Authorization: `Bearer ${token}` }, method: "POST" }
      )
    );

    expect(response.status).toBe(404);
    expect(runCalls).toEqual([]);
  });
});
```

On `25c6bf24` both assertions fail: the request URL's `?orgId=` is simply ignored, the response is `204`, and `runCalls` contains org B's automation id. (Drop `?orgId=` from the URL and the same run succeeds on main with no org knowledge required at all.) Scratch file only — the landed tests extend `internal-automations.test.ts`, not a standalone file.

### Why that synthetic repro is the right shape

The endpoint's only inputs are the bearer token and the automation id; there is no provider or channel dependency to simulate. The in-memory adapter plus `seedLocalClientUser` is the same scaffolding the existing suite (`internal-automations.test.ts`) uses, and `automationService.create` is the service's own write path, so org B's automation is a real row with a real `orgId` — the run reaching `agent.runAutomation` means the server executed a cross-org action on the caller's bare assertion of an id.

## Root cause

| Layer | What it checks | Org bound to caller? |
|---|---|---|
| Auth middleware | Bearer token verifies against the install's single local token hash | No — one principal for the whole install (`local-auth.ts:132-157`) |
| Org middleware | Skipped — internal routes mount before it (`app.ts`, by design for the worker) | No |
| Route handler | `auth.mode === "local-token"`, then archived-org state of the *target* | No |
| Service | `get(id, orgId?)` — org check only when `orgId` is passed (`automation-service.ts:83-90`) | Available, unused here |

- The worker is install-wide and legitimately runs every org's schedules, so "give the worker an org" was never modeled; the route grew up as a trusted backdoor with no second layer.
- The public route proves the second layer exists and is one argument long: `automations.ts` resolves via `get(automationId, orgId)` and 404s on mismatch.
- The archived-org check answers "is this org alive", not "may this caller touch this org" — a guard checking the building's hours, never the visitor's floor pass.

## Why it went unnoticed

```
$ grep -n "requireNotViewerFromContext\|orgId" apps/server/src/http/routes/internal-automations.ts
74:    if (automation.orgId && orgService) {   // target-state check only
```

The RBAC regression matrix (`rbac-mutations.test.ts`, extended in #401/#423) covers *public* mutating routes where a viewer's session can be role-checked; internal routes are local-token-only by design and have no role to assert, so they sat outside every audit. The 401 for browser sessions also makes the route look guarded in casual probing — the missing layer only appears once a local token is in hand, which is exactly the scenario defense-in-depth exists for.

## What this is not

- Not exploitable from the dashboard — browser sessions get 401 (control above).
- Not the archived-org gap — that check exists and works; it just doesn't bind the caller.
- Not a reduction of a leaked token's blast radius — `GET /v1/internal/automations/schedules` hands the same token every id+orgId pair, so the run check is replayable. Containment under a leak is the file-permission work (#428/#429) and org-bound token principals (follow-up); rotation is already platform-admin + CSRF (`auth.ts:598-632`).
- Not about `GET /v1/internal/automations/schedules` — the install-wide worker legitimately needs the full list; worth a follow-up discussion (token org-claims) but separate from the run action.

## Proposed fix

Require the caller to assert the owning org and resolve through the existing org-scoped get — the worker already knows it from the schedules list:

```ts
const automationId = decodeURIComponent(c.req.param("automationId"));
const orgId = c.req.query("orgId")?.trim();
if (!orgId) {
  return errorResponse("orgId query parameter is required.", 400);
}

const automation = await automationService.get(automationId, orgId);
if (!automation) {
  return errorResponse("Automation not found", 404);
}
```

with the delegate chain carrying the org through: `AutomationSchedulerDelegate.runAutomation(id, orgId)` (`automation.orgId` is in scope at both call sites), and `client.runAutomationInternal(id, orgId)` appending `?orgId=`. This makes an id without its org a 400 and an id paired with the wrong org a 404; the install-wide worker keeps its behavior, with the delegate and client changes carried in the same change.

## Acceptance criteria

- `POST /v1/internal/automations/:id/run` without `?orgId=` returns 400.
- With a `?orgId=` that does not own the automation, it returns 404 and `agent.runAutomation` is never invoked.
- With the owning org, behavior is unchanged (204 / 409 / archived-org 404).
- The automation worker still runs its schedules end-to-end.

## Test plan

- Extend `internal-automations.test.ts`: missing-orgId → 400; cross-org id under org A → 404 with a spy asserting no run; existing happy-path/skip/archived tests updated to carry `?orgId=`.
- `automation-scheduler.test.ts` stubs keep compiling (fewer-parameter delegates remain assignable) and the delegate gains the org argument at both call sites.
- `client.test.ts` pins the new query string.

## Not verified

- The full worker e2e under PM2 (delegate and client are covered by unit tests; no live PM2 run was performed).
- Docker deployments using `NAKAMA_LOCAL_AUTH_TOKEN` (env-managed token path — hash comparison logic untouched, but not exercised here).

## References

- `apps/server/src/http/routes/internal-automations.ts:61-90`
- `apps/server/src/services/automation-service.ts:83-90`
- `apps/server/src/http/routes/automations.ts` (public twin, org-scoped)
- `packages/core/src/automation-scheduler.ts` (delegate)
- `packages/client/src/client.ts:1287-1294` (worker client call)
- `packages/core/src/local-auth.ts` (single install-wide token)

---
The fix and its tests are ready on a branch — I'll open the PR next.

cakobob:
## Summary

Run an internal automation → the caller must assert its owning org (`?orgId=`), and the route resolves through the org-scoped `automationService.get(id, orgId)` the public route already uses. Fixes #434.

**Before:** `POST /v1/internal/automations/:id/run` executed any live automation id with only the install-wide local-token check.
**After:** bare ids return 400, mismatched id/org pairs 404; the worker passes the org it already receives from `GET /schedules` (`AutomationSchedulerDelegate.runAutomation(id, orgId)`).

Why safe:
1. Fail-closed ownership assertion only — no containment claimed: a token holder can still replay id+orgId pairs from `/schedules` (containment follow-ups: #428/#429, org-bound principals)
2. Worker and server ship in the same install; the delegate/client changes ride in the same commit, so the 400 can't outrun the caller update
3. Owning-org behavior unchanged — runs still 204/409, archived orgs still 404 (archived-guard mutation-checked)

Only re-add / follow up if an out-of-repo caller of this internal route turns up needing the bare-id form — the 400 is intentional per the issue's acceptance criteria.

## Test plan
- [x] `bun test apps/server/src/http/routes/internal-automations.test.ts` (9 pass — 2 new: missing orgId → 400, cross-org id → 404 with a no-run spy; both fail on unfixed main)
- [x] `bun test` (2375 pass / 0 fail, 336 files)
- [x] `curl -X POST -H "Authorization: Bearer $(cat ~/.nakama/local-auth-token)" http://localhost:4310/v1/internal/automations/any-id/run` → `400 {"error":"orgId query parameter is required."}`


fajarhide:
@cakobob I checked this branch out and ran it. Reads ready to me.

The `?.trim()` and falsy check at `internal-automations.ts:69` is what makes the rest of it safe. `automationService.get` skips its own scope check when `orgId` is falsy (`automation-service.ts:85`, `orgId && !automationBelongsToOrg(...)`), so an empty `?orgId=` reaching the service would have restored the exact unscoped lookup this PR removes. Rejecting it at the route is what closes that.

To check the two new tests actually hold the guard up, I put main's route back under them:

```sh
git checkout upstream/main -- apps/server/src/http/routes/internal-automations.ts
bun test apps/server/src/http/routes/internal-automations.test.ts
```

```
173 |     expect(response.status).toBe(400);
error: expect(received).toBe(expected)
Expected: 400
Received: 204
(fail) internal automation routes > rejects a run without an orgId assertion [85.53ms]

233 |     expect(response.status).toBe(404);
error: expect(received).toBe(expected)
Expected: 404
Received: 204
(fail) internal automation routes > refuses to run an automation under another org's id [79.19ms]

 7 pass
 2 fail
```

204 on the cross-org case, so the run really executed under another org's id. That is #434 reproduced.

One gap, not blocking. The whole fix hangs on `automation.orgId` reaching the client from `automation-scheduler.ts:87,114`, and nothing asserts it. I pointed both call sites at a literal wrong org:

```sh
# packages/core/src/automation-scheduler.ts:87,114
#   .runAutomation(automation.id, automation.orgId)  ->  .runAutomation(automation.id, "org_WRONG")
bun test packages/core/src/automation-scheduler.test.ts \
         apps/platform/automation/src/scheduler.test.ts \
         packages/client/src/client.test.ts
```

```
 21 pass
 0 fail
```

`tsc --noEmit` stayed clean too, since a one-arg fake is assignable to the two-arg interface. On a real install a wrong org fails closed: every scheduled automation 404s and stops, with one worker log line to show for it. The fake at `automation-scheduler.test.ts:39` already receives the org from `schedule()`, so recording it is about a line.

One question. `/schedules` emits `orgId: automation.orgId ?? ""` (`internal-automations.ts:36,45`), so an automation with no org now gets `?orgId=` and a permanent 400 where it used to run. I could not build that row through any path: `create()` requires an org, `migrate.ts:801` backfills from the profile, and profiles with no org are deleted first. Do you know one that survives? If not, that `?? ""` could be a filter so the worker never asks.

Two I could not fault. `isLongHeldAutomationRunRequest` matches on `new URL(request.url).pathname` (`sse-idle-timeout.ts:39`), so the query string does not break the idle-timeout exemption for long runs. And `bun run test` on the branch gives 2375 pass, 0 fail, the number in your body.

I did not run a live worker against a real instance, only the suites and the route through `app.fetch`.

cakobob:
@fajarhide, thanks for running the branch.

Both suggestions taken in `e7e182de`.

- **Filter it is** — good idea. To your question: I couldn't find a path that creates an org-less row either (`automationService.create` always writes `orgId`; the nullable column is just pre-tenant legacy). `/schedules` now drops those rows, and the `?? ""` fallback is gone.
- **Delegate org test added.** Reproducing your experiment, wrong org hardcoded at both call sites, now fails the suite. The filter has its own route test too, and disabling it flips that one.

Full suite: 2377 pass / 0 fail. 

Verified the same way you did — test suites and app.fetch only, no live worker run.

PR #440, opened 2026-08-22
fix(automations): scope internal run to the owning org

ahmadrosid:
## Summary

Scheduled automation runs are scoped to the owning org. Wrong org → 404 and no run.

**Before:** `POST /v1/internal/automations/:id/run` looked up the job by ID only.
**After:** requires `?orgId=`; `automationService.get(id, orgId)`.

Why safe:
1. Scheduler already has `automation.orgId` on the schedule
2. Missing `orgId` → 400; wrong org → 404 and `runAutomation` is never called
3. Worker and client thread the same argument — no ID-only fallback

Only re-add ID-only lookup if a non-org caller still needs to fire a job (none in-tree).

## Test plan
- [x] `bun test packages/core/src/automation-scheduler.test.ts packages/client/src/client.test.ts apps/platform/automation/src/scheduler.test.ts apps/server/src/http/routes/internal-automations.test.ts` (31 pass)
- [ ] Trigger one scheduled automation in a real org — it still runs

### DIFF
--- a/apps/platform/automation/src/scheduler.test.ts
+++ b/apps/platform/automation/src/scheduler.test.ts
@@ -9,14 +9,14 @@ import { AutomationWorkerScheduler } from "./scheduler";
 function createMockClient(
   overrides: Partial<{
     listAutomationSchedules: () => Promise<AutomationSchedule[]>;
-    runAutomationInternal: (id: string) => Promise<void>;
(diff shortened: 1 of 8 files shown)


ahmadrosid:
Closing as a duplicate of #438, which landed first and covers the same org-scoped internal run change (Fixes #434).

Issue #441, opened 2026-08-23
feat: send error tracking to a Sentry-compatible error tracker


fajarhide:
## Summary

Let an operator point their self-hosted nakama at the error tracker they already run, by pasting a Sentry-compatible DSN into the Integrations tab. Nakama becomes a sender only: it stores no events, groups nothing, and ships no error UI of its own.

Today a crash in a self-hosted install exists as one line of stderr on whatever host it ran on, and nobody reads that until a user complains. There is no way to answer "is this install healthy" without shell access to four processes.

## Motivation

The product strategy makes error rate a P1 concern:

> **Behavior on tool timeout/error (P1).** Retry at most twice with backoff, then fall back to a controlled message to the user. Every failure is recorded in Agent Health Metrics for error rate monitoring.

"Recorded for error rate monitoring" needs somewhere to record to. Nakama has no error sink at all right now, so that requirement has no destination, and any Agent Health Metrics surface would have to invent one.

This issue builds that destination, and deliberately builds only the transport half. It does **not** implement the retry, the backoff, or the user-facing fallback message: those are agent-loop behaviour and belong in their own issue. What it does give that work is a place to send a failure once the retries are exhausted, on a protocol that already does grouping, rate limiting and alerting, so nakama does not have to grow its own.

Sentry-compatible was chosen over a bespoke endpoint because one protocol covers Sentry, GlitchTip, Bugsink and a self-hosted Sentry: they all accept the same store endpoint and `X-Sentry-Auth` header. A self-hoster who already runs one of them gets this for the cost of pasting a DSN.

### Phase 1: the sender (#442)

Core module: DSN parsing, the Sentry event shape, the scrubber, and an on-disk queue. Wired into nothing, so it changes no runtime behaviour on its own.

The queue is not belt and braces. Measured on Bun 1.3.14: with `uncaughtExceptionMonitor` the process dies before any async send finishes, and even the `exit` handler does not run, while a `writeFileSync` inside the monitor does complete. So a report is queued synchronously before the send and removed only once the sink confirms delivery. The same queue doubles as the retry for a send that simply failed.

### Phase 2: the four processes (#443)

The server and the Telegram, WhatsApp and Discord workers install the handlers. Only the server drains the queue at startup: the four share one config directory and the queue is a read-modify-write on one file, so draining from all of them would race.

### Phase 3: the Integrations form (#444)

[...]

PR #442, opened 2026-08-23
feat(error-tracking): deliver errors to a Sentry-compatible DSN

fajarhide:
## Summary

Phase 1 of #441: the sender. An operator points nakama at their own Sentry, GlitchTip or Bugsink by saving a DSN, and this PR is the module that knows how to get an error there. Wired into nothing yet, so it changes no runtime behaviour on its own.

**Before:** an error in a self-hosted install exists only as stderr on the host that ran it.
**After:** `reportError` queues the report to disk, sends it, and clears it from the queue only once the sink confirms delivery.

Measured against a local ingest, both arms:

```
BEFORE  no DSN configured (the default)
  queue on disk after the crash: no file
  next process: FLUSHED 0 report(s)
  ingest received: 0 event(s)

AFTER   operator saved their DSN
  queue on disk after the crash: 1
  next process: FLUSHED 1 report(s)
  ingest received: 1 event(s)
```

Why safe:
1. Off is the default and off is free. No DSN means no request, no queue file, and no work on the request path. `DO_NOT_TRACK=1` overrides a saved DSN.
2. The scrubber runs before anything leaves the process: home paths, secrets and quoted payloads. It is the one part of the closed #195 branch that came across unchanged.
3. Nothing imports this yet. `packages/core/src/index.ts` gains four exports and no call site.

Why it holds a disk queue, which is the part worth arguing about. With `uncaughtExceptionMonitor` the process dies before any async send finishes, measured on Bun 1.3.14, and even the `exit` handler does not run, while a `writeFileSync` inside the monitor does complete. The queue is therefore the only thing that survives an uncaught exception, and it doubles as the retry for a send that failed. The report id is reused as the Sentry event id so a queued copy and a live send collapse into one event at the ingest.

Scope is process crashes, so there is no `reportInvariant`, no breadcrumb trail and no AsyncLocalStorage request context: an error thrown in a Hono handler goes to `onError`, never to `uncaughtException`, so that machinery would have had no producer. It comes back with HTTP 5xx reporting if that is ever wanted.

## Test plan
- [x] `bun run test` across all workspaces: 0 fail. Two `apps/server` tests fail only under full parallel load and pass on a rerun and on `bun test apps/server` alone (967 pass, 0 fail), which is the known flaky family in that workspace.
- [x] `bun test packages/core/src/error-tracking`: 56 pass across 4 files.
- [x] End to end above, driven by a real uncaught exception rather than a synthetic call.

Refs #441. #443 installs the handlers, #444 adds the Integrations form.

### DIFF
--- a/packages/core/src/error-tracking-queue.ts
+++ b/packages/core/src/error-tracking-queue.ts
@@ -1,0 +1,62 @@
+import { mkdirSync, readFileSync, writeFileSync } from "node:fs";
+import { join } from "node:path";
+import type { ErrorReport } from "./error-tracking";
+import { getErrorTrackingConfigDir } from "./error-tracking-config";
+import { PRIVATE_DIR_MODE, PRIVATE_FILE_MODE } from "./fs";
+
+/**
+ * An uncaught exception kills the process before any async send can finish, so the
+ * report is written here synchronously and delivered by whichever process starts next.
+ * Measured, not assumed: with uncaughtExceptionMonitor even the "exit" handler does not
+ * run, while a writeFileSync inside the monitor completes.
+ *
+ * A send that fails for any other reason leaves its entry behind too, so the queue
+ * doubles as the retry.
+ */
+export const MAX_PENDING_ERROR_REPORTS = 5;
+
+export function getPendingErrorReportsPath(): string {
+  return join(getErrorTrackingConfigDir(), "pending.json");
+}
+
+export function readPendingErrorReports(): ErrorReport[] {
+  try {
+    const parsed: unknown = JSON.parse(
+      readFileSync(getPendingErrorReportsPath(), "utf8")
+    );
+    return Array.isArray(parsed) ? (parsed as ErrorReport[]) : [];
+  } catch {
+    // Missing, unreadable or corrupt all mean the same thing: nothing to deliver.
+    return [];
+  }
+}
+
+function write(reports: ErrorReport[]): void {
+  mkdirSync(getErrorTrackingConfigDir(), {
+    mode: PRIVATE_DIR_MODE,
+    recursive: true,
+  });
+  writeFileSync(getPendingErrorReportsPath(), JSON.stringify(reports), {
+    mode: PRIVATE_FILE_MODE,
+  });
+}
+
+/** Synchronous on purpose. See the note above. */
+export function appendPendingErrorReport(report: ErrorReport): void {
+  try {
+    write(
+      [...readPendingErrorReports(), report].slice(-MAX_PENDING_ERROR_REPORTS)
+    );
+  } catch {
+    // A full or read-only config dir must not turn one crash into two.
+  }
+}
+
+export function removePendingErrorReport(id: string): void {
+  try {
+    write(readPendingErrorReports().filter((report) => report.id !== id));
+  } catch {
+    // Worst case the report is delivered twice, and the event id collapses it at the
+    // ingest. That is strictly better than failing here.
+  }
+}
--- a/packages/core/src/error-tracking-queue.ts
+++ b/packages/core/src/error-tracking-queue.ts
@@ -47,8 +47,10 @@ export function appendPendingErrorReport(report: ErrorReport): void {
     write(
       [...readPendingErrorReports(), report].slice(-MAX_PENDING_ERROR_REPORTS)
     );
-  } catch {
-    // A full or read-only config dir must not turn one crash into two.
+  } catch (error) {
+    // A full or read-only config dir must not turn one crash into two, but swallowing it
+    // outright leaves an operator with tracking on and no reports and no reason why.
+    console.error("[nakama:error-tracking] cannot queue report", error);
   }
 }
 
--- a/packages/core/src/error-tracking-sentry.ts
+++ b/packages/core/src/error-tracking-sentry.ts
@@ -5,7 +5,7 @@ import {
   setErrorSink,
 } from "./error-tracking";
 import {
-  loadCachedErrorTrackingConfig,
+  loadErrorTrackingConfig,
   resolveErrorTrackingDsn,
 } from "./error-tracking-config";
 
@@ -113,7 +113,7 @@ export async function sendSentryEvent(
  */
 export function createErrorTrackingSink(): ErrorSink {
   return async (report) => {
-    const config = await loadCachedErrorTrackingConfig();
+    const config = await loadErrorTrackingConfig();
     const dsn = parseSentryDsn(resolveErrorTrackingDsn(config) ?? "");
 
     if (!dsn) {
--- a/packages/core/src/error-tracking-sentry.ts
+++ b/packages/core/src/error-tracking-sentry.ts
@@ -18,6 +18,34 @@ export interface SentryDsn {
 }
 
 /**
+ * The slice of the Sentry store payload nakama actually sends. Hand-rolled because no
+ * SDK is in use, and named so the envelope is a checked shape rather than a bag: a
+ * typo in a key here is silently ignored by the ingest and impossible to spot later.
+ */
+export interface SentryEvent {
+  contexts: {
+    os: { name: string };
+    runtime: { name: string; version: string };
+  };
+  event_id: string;
+  exception: { values: { type: string; value: string }[] };
+  extra?: { stack: string };
+  fingerprint: string[];
+  level: "error" | "info";
+  logger: string;
+  platform: string;
+  tags: {
+    api_version: string;
+    arch: string;
+    bun: string;
+    kind: string;
+    os: string;
+    source: string;
+  };
+  timestamp: string;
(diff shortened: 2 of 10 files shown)


packages/core/src/error-tracking-sentry.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  const prefix = segments.length > 0 ? `/${segments.join("/")}` : "";
+
+  return {
+    endpoint: `${url.protocol}//${url.host}${prefix}/api/${projectId}/store/`,
+    publicKey,
+  };
+}
+
+export function toSentryEvent(report: ErrorReport): Record<string, unknown> {
```
vincentri: **ts-taste.md:** `Record<string, unknown>` is on the avoid-list (the file says "Prefer SDK types and discriminated unions over Record<string, unknown>"). Defensible here because the Sentry envelope has no SDK in use and you hand-roll the shape, but if you ever grow a `SentryEvent` discriminated union (or pick up `@sentry/types` for the boundary), this is the spot that tightens up. Keeping as a judgement call rather than a hard violation since the function is the deliberate envelope boundary.

packages/core/src/error-tracking-queue.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/** Synchronous on purpose. See the note above. */
+export function appendPendingErrorReport(report: ErrorReport): void {
+  try {
+    write(
+      [...readPendingErrorReports(), report].slice(-MAX_PENDING_ERROR_REPORTS)
+    );
+  } catch {
+    // A full or read-only config dir must not turn one crash into two.
+  }
```
vincentri, line 54: **Smell — silent swallow.** The "don't turn one crash into two" comment is correct, but a permanently-broken config dir (full disk, permission revoked) now produces zero reports and zero signal. Consider writing a one-line `.last-error` marker the operator can read on `nakama doctor`, or at least surfacing it through `enabled` so a future health check picks it up. Not a Phase 1 blocker since this PR is explicitly "wired into nothing", but the swallow will bite later.

packages/core/src/error-tracking.ts
```diff
@@ -0,0 +1,274 @@
...
+    };
+  }
+
+  if (typeof error === "string") {
+    return { message: error, name: "NonError" };
+  }
+
+  try {
+    return {
+      message: JSON.stringify(error) ?? String(error),
```
vincentri, line 131: **Smell — circular fallback.** For a circular non-Error rejection `JSON.stringify(error)` returns `undefined` and the `?? String(error)` fallback then yields `"[object Object]"`, which dedups every circular rejection into one fingerprint. Acceptable for the rare case, but worth a one-line test that asserts a circular object still fingerprints distinctly from a plain "[object Object]" string, or a comment that calls out the trade.

vincentri (commented):
## Standards

This PR is on the lighter end of the standards axis. All 55 new tests pass and the code is well-commented with measured reasoning. No hard violations of `AGENTS.md`, `CONTRIBUTING.md`, `ts-taste.md`, or any smell baseline that tooling would not already enforce. A few judgement calls worth surfacing:

- **ts-taste.md borderline** — `toSentryEvent` returns `Record<string, unknown>` (`packages/core/src/error-tracking-sentry.ts:56`). `ts-taste.md` says prefer SDK types and discriminated unions. Defensible here because the Sentry envelope has no SDK in use and the shape is hand-rolled, but a `SentryEvent` discriminated union would tighten the boundary. Inline comment posted.
- **Smell — silent swallow** — `appendPendingErrorReport` catches and discards write errors (`packages/core/src/error-tracking-queue.ts:50-52`). The "don't turn one crash into two" rationale is correct, but a permanently broken config dir now produces zero reports and zero signal. Consider a one-line `.last-error` marker readable by a future health check. Inline comment posted.
- **Smell — circular fallback** — `errorToParts` falls back to `JSON.stringify(error) ?? String(error)` (`packages/core/src/error-tracking.ts:131`). Circular objects collapse to `"[object Object]"` and dedup into one fingerprint. Acceptable for the rare case; an inline test asserting the behaviour would make the trade explicit. Inline comment posted.
- **Smell — hidden singleton** — `enabled` is a module-level mutable (`packages/core/src/error-tracking.ts:38`), mutated by `refreshErrorTrackingEnabled`. Fine for the "Phase 1 wired into nothing" scope, but Phase 2 callers must remember to call `refreshErrorTrackingEnabled` after every `saveErrorTrackingDsn`. Worth a `@remarks` on `setErrorSink` / `installErrorTrackingSink` so the contract is documented, not assumed.

## Spec

Phase 1 of #441 calls for "DSN parsing, the Sentry event shape, the scrubber, and an on-disk queue. Wired into nothing, so it changes no runtime behaviour on its own." All four pieces are present and verified:

- DSN parsing — `parseSentryDsn` (`error-tracking-sentry.ts:25`), accepts Sentry / GlitchTip / Bugsink / subpath installs, rejects garbage.
- Sentry event shape — `toSentryEvent` (`error-tracking-sentry.ts:56`): own `fingerprint`, `event_id = report.id` so a queued copy and a live send collapse into one, no `server_name`, no `user`, `level: "info"` for `kind: "test"`.
- Scrubber — `scrubText` (`error-tracking-scrub.ts:58`): credentials (anthropic / openai / github / slack / aws / bearer / `apiKey=` / `ANTHROPIC_TOKEN=`), emails, home paths, double-quoted payloads, brackets/braces, 4000-char cap.
- On-disk queue — `error-tracking-queue.ts`: synchronous `writeFileSync` before the async send, `MAX_PENDING_ERROR_REPORTS = 5`, doubles as the retry path.

Verified acceptance items within Phase 1 scope:
- Empty DSN sends nothing, contacts no endpoint, writes no queue file — `error-tracking.test.ts:215`, `error-tracking-sentry.test.ts:163`.
- `DO_NOT_TRACK=1` overrides a saved DSN — `error-tracking-config.test.ts:68`, `error-tracking-sentry.test.ts:167`.
- Scrubbing runs before send — `error-tracking.ts:149,155`.
- An uncaught exception reaches the sink via the synchronous queue — `error-tracking.ts:206-208`.
- Config read per send, not at startup — `error-tracking-sentry.ts:116-117`.
- "Wired into nothing" — `grep -r 'reportError\|installErrorHandlers\|installErrorTrackingSink\|flushPendingErrorReports' apps packages` confirms no call sites outside the new module and its tests.

No missing or partial requirements, no scope creep into Phase 2/3/4 (handlers, Integrations form, tool-failure routing), no wrong-looking implementations in Phase 1.

## Summary
- Standards: 4 findings, all judgement calls (worst: `ts-taste.md` borderline on `Record<string, unknown>` at the Sentry envelope boundary).
- Spec: 0 findings.

packages/core/src/error-tracking-sentry.ts
fajarhide: Done in 62a4ccaa. `SentryEvent` names the six tags and the exception shape, and `sendSentryEvent` takes it too, so a typo in a key now fails the build instead of being a field the ingest silently drops.

Two tests moved with it: `event.user` became `expect(event).not.toHaveProperty("user")`, and the down-ingest case sends a real event instead of `{}`.

packages/core/src/error-tracking-queue.ts
fajarhide: Signal added, marker file skipped. The catch stays, since one crash must not become two, but it now writes `[nakama:error-tracking] cannot queue report` to stderr with the underlying error.

A `.last-error` file would need the same directory that just refused the write, so it goes quiet in exactly the case worth reporting. There is no `nakama doctor` yet either; when something does read health, stderr is where this already lands.

packages/core/src/error-tracking.ts
fajarhide: Comment plus a test, no behaviour change. Small correction on the mechanism: on a cycle `JSON.stringify` throws rather than returning `undefined`, so it lands in the `catch` and `String(error)` flattens it. The `?? String(error)` on the line above is for `undefined` and symbols.

The test now asserts two different circular objects produce the same fingerprint, so the collapse is recorded rather than implied. Splitting them needs an object walk, and a rejection carrying a circular non-Error is rare enough not to pay for that.

fajarhide:
Thanks @vincentri, all four were worth acting on. Three are fixed in 62a4ccaa and answered in the threads.

On the `enabled` singleton: the remark went on `saveErrorTrackingDsn` rather than `setErrorSink`, since saving the DSN is what creates the obligation and is where a Phase 2 caller is looking. #444 already pairs them (`agent-service.ts`, save then refresh), so this documents a contract that is being kept, at the point it could be forgotten.

#443 and #444 are rebased on this.

```
bunx ultracite check    clean
bun run knip    clean
```

ahmadrosid (approved):
lets just get this merged i believe in you @fajarhide 😄

PR #443, opened 2026-08-23
feat(error-tracking): install the handlers in the server and the workers

fajarhide:
## Summary

> Stacked on #442, and out of draft because #442's feature commit is already on `main`: @ahmadrosid landed it with his cleanup in #453. What is still shared is the work on top of it, the scrub test refactor and the review fixes, 8 files under `packages/core`. The four entrypoints under `apps/` are what this PR owns. Merging #442 first keeps that split; merging this one first carries those 8 files along and leaves #442 empty.

Phase 2 of #441: the four processes that can die now report it. The server and the Telegram, WhatsApp and Discord workers install the handlers from #442, and the server drains the queue at startup. Stacked on #442.

**Before:** a crash in any of the four exists only as stderr on the host.
**After:** the process queues the report synchronously, and the server delivers it to the operator's DSN.

Measured against a real `bun apps/server/src/index.ts`, with one report left in the queue by a previous process:

```
BEFORE  no config.ini at all (feature off)
  queue before: 1
  server up: 200
  ingest received: 0 event
  queue after: 1        <- left alone, not dropped

AFTER   operator saved their DSN
  queue before: 1
  server up: 200
  ingest received: 1 event
  queue after: 0
```

Why safe:
1. With no DSN nothing changes: no request, nothing written, and a queued report is left alone rather than discarded.
2. `installErrorHandlers` uses `uncaughtExceptionMonitor`, which observes the error and leaves Bun's own crash and exit intact. Exit codes are untouched. It does not reach a throw during module init of an imported package: ESM evaluates every import before the first statement runs, so the handler is not registered yet. The comment in `index.ts` says so.
3. Only the server drains. The four share one config directory and the queue is a read-modify-write on a single file, so draining from all of them would race. The workers are spawned by the server, so it is the one that is always up. The code carries the upgrade path if a worker ever runs alone.

Rebased on #442 after @ahmadrosid's cleanup landed on `main`. One line changed here while doing it: the "server is the only drainer" comment had been pasted into all three worker entrypoints, and the server file already explains the whole decision, so the three copies are gone.

There is no unit test here worth writing. The change is four call sites; what needed proving is that a real server drains at startup, which is the run above. The sender and the queue are covered by #442.

## Test plan
- [x] `bun run test` across all workspaces: 0 fail.
- [x] Both arms above, driven by the real server entrypoint rather than a harness.

Refs #441. #444 adds the Integrations form so the DSN can be saved from the UI.

### DIFF
--- a/apps/server/src/index.ts
+++ b/apps/server/src/index.ts
@@ -1,9 +1,24 @@
 import { dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
+import {
+  flushPendingErrorReports,
+  installErrorHandlers,
+  installErrorTrackingSink,
+} from "@nakama/core";
 import type { Server } from "bun";
 import { ensureProcessPath } from "./lib/ensure-process-path";
 
 ensureProcessPath();
+installErrorHandlers("server");
+
+/**
+ * Only the server drains the queue. Four processes share one config dir, and the
+ * queue is a read-modify-write on a single file, so draining from all of them would
+ * race. The workers are spawned by the server, so it is the one that is always up.
+ * ponytail: single drainer, give the queue per-process files if a worker ever runs
+ * without a server.
+ */
+void installErrorTrackingSink().then(() => flushPendingErrorReports());
 
 import { generateSkillConsolidateMarkdown } from "@nakama/agent";
 import {
--- a/apps/server/src/index.ts
+++ b/apps/server/src/index.ts
@@ -9,6 +9,8 @@ import type { Server } from "bun";
 import { ensureProcessPath } from "./lib/ensure-process-path";
 
 ensureProcessPath();
+// Position is cosmetic: ESM evaluates every import above before this line runs, so a throw
+// inside @nakama/db or @nakama/agent module init is already past. Everything after is covered.
 installErrorHandlers("server");
 
 /**
--- a/apps/platform/discord/src/index.ts
+++ b/apps/platform/discord/src/index.ts
@@ -1,4 +1,5 @@
 import { createClient } from "@nakama/client";
+import { installErrorHandlers, installErrorTrackingSink } from "@nakama/core";
 import {
   ChannelOrgStore,
   getChannelOrgSelectionPath,
@@ -21,6 +22,9 @@ import { createBot } from "./bot";
 import { loadConfig } from "./config";
 import { SessionStore } from "./session-store";
 import { ThreadStore } from "./thread-store";
+
+installErrorHandlers("worker:discord");
+void installErrorTrackingSink();
 
 let spawnedChild: Bun.Subprocess | null = null;
 let clientStop: (() => void) | null = null;
--- a/apps/platform/telegram/src/index.ts
+++ b/apps/platform/telegram/src/index.ts
@@ -1,4 +1,5 @@
 import { createClient } from "@nakama/client";
+import { installErrorHandlers, installErrorTrackingSink } from "@nakama/core";
 import {
   ChannelOrgStore,
   getChannelOrgSelectionPath,
@@ -20,6 +21,9 @@ import { TelegramAuthStore } from "./auth-store";
 import { createBot } from "./bot";
 import { loadConfig } from "./config";
 import { SessionStore } from "./session-store";
+
+installErrorHandlers("worker:telegram");
+void installErrorTrackingSink();
 
 let spawnedChild: Bun.Subprocess | null = null;
 let botStop: (() => void) | null = null;
--- a/apps/platform/whatsapp/src/index.ts
+++ b/apps/platform/whatsapp/src/index.ts
@@ -1,4 +1,5 @@
 import { createClient } from "@nakama/client";
+import { installErrorHandlers, installErrorTrackingSink } from "@nakama/core";
 import {
   ChannelOrgStore,
   getChannelOrgSelectionPath,
@@ -23,6 +24,9 @@ import { loadConfig } from "./config";
 import { startWhatsAppOutboundServer } from "./outbound-server";
 import { SessionStore } from "./session-store";
 import { createWhatsAppSocket } from "./socket";
+
+installErrorHandlers("worker:whatsapp");
+void installErrorTrackingSink();
 
 let spawnedChild: Bun.Subprocess | null = null;
 let socketHandle: {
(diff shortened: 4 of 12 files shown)


apps/server/src/index.ts
```diff
@@ -1,9 +1,24 @@
...
+import {
+  flushPendingErrorReports,
+  installErrorHandlers,
+  installErrorTrackingSink,
+} from "@nakama/core";
 import type { Server } from "bun";
 import { ensureProcessPath } from "./lib/ensure-process-path";
 
 ensureProcessPath();
+installErrorHandlers("server");
```
vincentri, line 14: Style nit: `installErrorHandlers("server")` and the drain chain sit between two import groups (top batch lines 1-9, then this code, then the rest of the imports start at line 23). ESM hoists all imports before any user code runs, so the runtime order is unaffected — the source placement is cosmetic. Either way, a short comment near the top noting that handlers are intentionally installed before later module loads (so a throw during `@nakama/agent` / `@nakama/db` init still gets reported) would document the intent more clearly than the current arrangement does.

apps/server/src/index.ts
```diff
@@ -1,9 +1,24 @@
...
+
+/**
+ * Only the server drains the queue. Four processes share one config dir, and the
+ * queue is a read-modify-write on a single file, so draining from all of them would
+ * race. The workers are spawned by the server, so it is the one that is always up.
+ * ponytail: single drainer, give the queue per-process files if a worker ever runs
+ * without a server.
+ */
+void installErrorTrackingSink().then(() => flushPendingErrorReports());
```
vincentri, line 23: Soft observation: the PR description claims #442's commit shows up in this diff, but the diff stat (`7 files, +55/-35`) only contains the wiring + the test refactor. `git log origin/main -- packages/core/src/error-tracking.ts` shows `b2fe8483 feat(error-tracking): deliver errors to a Sentry-compatible DSN` (the #442 head) already in main. Worth trimming the lead paragraph before flipping this out of draft.

vincentri (commented):
## Standards

Small, focused diff. Conventional commit title, ADHD PR description shape followed. 25/25 tests pass (`bun test packages/core/src/error-tracking-config.test.ts packages/core/src/error-tracking-scrub.test.ts packages/core/src/error-tracking.test.ts`); `bun x ultracite check` clean on the touched files.

Findings (all soft):

- **`apps/server/src/index.ts:11-21`** — `installErrorHandlers("server")` and the drain chain are placed between the first import group (lines 1-9) and the second (`@nakama/agent`, `@nakama/db`, etc., starting at line 23). ESM hoists all imports before any user code runs, so the source position is cosmetic; the intent ("handlers first") is signalled by the placement but not by a comment. Either move the calls after the import block (and rely on import hoisting to keep the runtime order), or add a one-line comment near the top noting the handlers are intentionally installed before later module loads so a throw inside `@nakama/agent`/`@nakama/db` init still gets reported.

- **PR description stale** — the lead paragraph claims #442's commit shows up in this diff, but `git log origin/main -- packages/core/src/error-tracking.ts` already shows `b2fe8483 feat(error-tracking): deliver errors to a Sentry-compatible DSN` on main. The diff stat (7 files, +55/-35) only carries the wiring and the test refactor. Trim the lead before flipping out of draft.

- The test refactor (`packages/core/src/error-tracking-scrub.test.ts`) is a strict improvement and aligns with the AGENTS.md "assert behavior, not prompt/description/error copy" rule — each case now carries its own secret and asserts both "input really contains it" and "output does not", so neutering `scrubText` to the identity function now fails 16 of the file's 19 tests rather than 8. Good cleanup.

## Spec

Phase 2 of issue #441 (line 27-29):

> The server and the Telegram, WhatsApp and Discord workers install the handlers. Only the server drains the queue at startup: the four share one config directory and the queue is a read-modify-write on one file, so draining from all of them would race.

Implementation matches:

- Server: `installErrorHandlers("server")` plus `void installErrorTrackingSink().then(() => flushPendingErrorReports())` — `apps/server/src/index.ts:12,21`.
- Telegram / WhatsApp / Discord workers: `installErrorHandlers("worker:...")` plus `void installErrorTrackingSink()` (no drain) — `apps/platform/{telegram,whatsapp,discord}/src/index.ts:25-26, 28-29, 26-27`. Without the sink install, the worker handlers would queue nothing, so installing both is required to satisfy the AC, not scope creep.

No missing requirements. No behaviour the issue did not ask for. The "Uncaught exception delivered via the synchronous queue and the next process to start" full-issue AC is satisfied end-to-end by the wiring (`installErrorHandlers` → `reportError` queues synchronously → `flushPendingErrorReports` drains on the next server start). Hard-to-find ACs from the full issue (scrubbing before egress, exit code preserved on send failure, `DO_NOT_TRACK` precedence, empty DSN means off) are covered by #442 and unchanged here.

## Summary

- Standards: 2 soft findings (style placement; PR-description staleness).
- Spec: no spec findings (PR implements the phase 2 AC cleanly).

apps/server/src/index.ts
fajarhide: You are right that the position is cosmetic, and it is worse than cosmetic: the arrangement implies protection it cannot give. ESM evaluates every import in the file before the first statement, so a throw during `@nakama/db` or `@nakama/agent` module init is already past by the time the handler registers.

Comment added in 9c69997f, saying that rather than saying "handlers first". The placement stays, since the split block and the `ensureProcessPath()` in it are already on `main` and moving only our two calls out would make the file less consistent, not more.

apps/server/src/index.ts
fajarhide: Confirmed, and the reason is worth recording: `b2fe8483` is an ancestor of `main` because @ahmadrosid landed it with his cleanup in #453 as a **merge**, not the squash the paragraph predicted. So the diff shrank on its own, exactly the thing that paragraph said would not happen.

Body rewritten. What this still shares with #442 is the work on top of that commit, the scrub test refactor and the review fixes, 8 files under `packages/core`. The four entrypoints under `apps/` are what this PR owns, so it stays a draft until that follow-up lands rather than until the whole of #442 does.

fajarhide:
Thanks @vincentri, both landed, and the second one was more than a stale sentence.

Chasing it turned up that `b2fe8483` is on `main` because @ahmadrosid merged it with his cleanup in #453 rather than squashing it, which is the opposite of what that paragraph predicted. So the draft rationale here was resting on a wrong assumption, not just stale wording. Body rewritten to say what is actually shared with #442 now: 8 files under `packages/core`, the scrub test refactor and the review fixes, not the feature commit.

The placement note is in 9c69997f as a comment rather than a move, and it states the limit the arrangement was quietly implying away: a throw during module init of an imported package is not covered, because ESM has already evaluated it.

```
bunx ultracite check    clean
```

ahmadrosid (approved):

PR #444, opened 2026-08-23
feat(error-tracking): configure the DSN from the Integrations tab

fajarhide:
## Summary

> **Draft until #443 merges.** Built on #443, which is built on #442, so their commits appear here too. The settings routes call into the core module from #442, so this cannot sit on `main` by itself. Fork PRs here land as squashes, so those commits will not become ancestors of `main` and this diff will not shrink on its own. Once the two below it land I will rebuild this branch from the new `main` with a cherry-pick of its own two commits, then mark it ready.

Phase 3 of #441: the operator can finally do this without a text editor. Adds an "Error tracking" section to the Integrations tab, so the DSN is saved from the UI instead of a hand-edited `~/.nakama/error-tracking/config.ini`. Stacked on #442 and #443.

**Before:** eight integration sections, no way to reach error tracking from the dashboard.
**After:** a ninth section with a DSN field, a status badge, and a test event button.

| Before | After |
| --- | --- |
|  |  |

Same instance, same 1440x1000 viewport at 2x. The before shows the default section, since the point is that "Error tracking" is not in the list.

Saving a DSN and pressing "Send test event", against a local ingest that logged exactly one received event:

The badge went from "Off" to "Sending", the field shows the saved DSN masked, and the message is the real result of the send rather than an optimistic string.

Why safe:
1. Writes are org admin or platform admin only. The DSN is workspace-global so there is no org to scope it to, which makes the role guard the only thing between a viewer and the whole install's error routing. Removing the guard turns the member and viewer tests red, checked both ways.
2. The GET response carries a masked DSN. A test asserts the raw value appears nowhere in the response body.
3. A malformed DSN is rejected at save rather than accepted and then silently never delivering.
4. "Send test event" sends one event directly instead of going through `reportError`, so a failed test does not leave a fake crash sitting in the delivery queue. It is disabled until a DSN is saved, and the event carries kind `test` at a lower level so it never sits in a triage queue next to a real crash.

Trimmed after @ahmadrosid's cleanup landed on #442. `ErrorTrackingSettingsCard` carried an `embedded` prop that nothing passed, so the header fork, the second status badge and the two padding variables sat on a path that never ran. The card is 17 lines shorter and renders identical bytes, so the screenshots above are still current.

## Test plan
- [x] `bun run test` across all workspaces: 0 fail.
- [x] `bun test apps/server/src/http/routes/error-tracking.test.ts`: 5 pass.
- [x] Role guard removed on purpose: the member and viewer tests fail, then pass again with it restored.
- [x] The flow above driven through a real browser against a throwaway instance, not the shared one.
- [x] Card screenshotted before and after the trim on the same instance: identical sha256, 0 differing pixels. A one-word control edit moved 1663 pixels, so the comparison is not measuring nothing.

Refs #441. Phase 4, routing tool failures past their retry budget into the same tracker, is the part that closes the Agent Health Metrics requirement and is not started.

### DIFF
--- a/apps/server/src/index.ts
+++ b/apps/server/src/index.ts
@@ -1,9 +1,24 @@
 import { dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
+import {
+  flushPendingErrorReports,
+  installErrorHandlers,
+  installErrorTrackingSink,
+} from "@nakama/core";
 import type { Server } from "bun";
 import { ensureProcessPath } from "./lib/ensure-process-path";
 
 ensureProcessPath();
+installErrorHandlers("server");
+
+/**
+ * Only the server drains the queue. Four processes share one config dir, and the
+ * queue is a read-modify-write on a single file, so draining from all of them would
+ * race. The workers are spawned by the server, so it is the one that is always up.
+ * ponytail: single drainer, give the queue per-process files if a worker ever runs
+ * without a server.
+ */
+void installErrorTrackingSink().then(() => flushPendingErrorReports());
 
 import { generateSkillConsolidateMarkdown } from "@nakama/agent";
 import {
--- a/apps/server/src/index.ts
+++ b/apps/server/src/index.ts
@@ -9,6 +9,8 @@ import type { Server } from "bun";
 import { ensureProcessPath } from "./lib/ensure-process-path";
 
 ensureProcessPath();
+// Position is cosmetic: ESM evaluates every import above before this line runs, so a throw
+// inside @nakama/db or @nakama/agent module init is already past. Everything after is covered.
 installErrorHandlers("server");
 
 /**
--- a/apps/web/src/components/ErrorTrackingSettingsCard.tsx
+++ b/apps/web/src/components/ErrorTrackingSettingsCard.tsx
@@ -1,6 +1,9 @@
 import { Link01Icon, ViewIcon, ViewOffIcon } from "hugeicons-react";
 import { useState } from "react";
-import { IntegrationCardShell } from "@/components/integration-settings.shared";
+import {
+  IntegrationCardShell,
+  IntegrationStatusHeader,
+} from "@/components/integration-settings.shared";
 import { Button } from "@/components/ui/button";
 import {
   InputGroup,
@@ -15,27 +18,6 @@ import {
   useSendErrorTrackingTest,
 } from "@/hooks/use-error-tracking";
 import { formatError } from "@/lib/client";
-
-function StatusBadge({ configured }: { configured: boolean }) {
-  if (configured) {
-    return (
-      <span className="inline-flex shrink-0 items-center gap-1.5 rounded-full border border-emerald-200 bg-emerald-50 px-2.5 py-0.5 font-medium text-emerald-800 text-xs dark:border-emerald-800/60 dark:bg-emerald-950/40 dark:text-emerald-200">
-        <span aria-hidden className="size-1.5 rounded-full bg-emerald-500" />
-        Sending
-      </span>
-    );
-  }
-
-  return (
-    <span className="inline-flex shrink-0 items-center gap-1.5 rounded-full border border-border bg-background px-2.5 py-0.5 font-medium text-muted-foreground text-xs">
-      <span
-        aria-hidden
-        className="size-1.5 rounded-full bg-muted-foreground/60"
-      />
-      Off
-    </span>
-  );
-}
 
 export function ErrorTrackingSettingsCard() {
   const {
@@ -90,18 +72,13 @@ export function ErrorTrackingSettingsCard() {
 
   return (
     <IntegrationCardShell>
-      <div className="flex items-start justify-between gap-4 p-5 pb-4">
-        <div className="min-w-0 space-y-1">
-          <h2 className="font-semibold text-base text-foreground leading-tight [text-wrap:balance]">
-            Error tracking
-          </h2>
-          <p className="text-muted-foreground text-sm leading-snug [text-wrap:pretty]">
-            Send this instance's own errors to the error tracker you already
-            run.
-          </p>
-        </div>
-        <StatusBadge configured={configured} />
-      </div>
+      {/* One state, not the channels' three: a saved DSN means it is sending. */}
+      <IntegrationStatusHeader
+        configured={configured}
+        connected={configured}
+        statusBadge={configured ? "Sending" : "Off"}
+        title="Error tracking"
+      />
 
       <div className="border-border border-t" />
 
--- a/apps/web/src/components/ErrorTrackingSettingsCard.tsx
+++ b/apps/web/src/components/ErrorTrackingSettingsCard.tsx
@@ -1,0 +1,229 @@
+import { Link01Icon, ViewIcon, ViewOffIcon } from "hugeicons-react";
+import { useState } from "react";
+import { IntegrationCardShell } from "@/components/integration-settings.shared";
+import { Button } from "@/components/ui/button";
+import {
+  InputGroup,
+  InputGroupAddon,
+  InputGroupButton,
+  InputGroupInput,
+} from "@/components/ui/input-group";
+import { Spinner } from "@/components/ui/spinner";
+import {
+  useErrorTrackingSettings,
+  useSaveErrorTrackingSettings,
+  useSendErrorTrackingTest,
+} from "@/hooks/use-error-tracking";
+import { formatError } from "@/lib/client";
+import { cn } from "@/lib/utils";
+
+function StatusBadge({ configured }: { configured: boolean }) {
+  if (configured) {
+    return (
+      <span className="inline-flex shrink-0 items-center gap-1.5 rounded-full border border-emerald-200 bg-emerald-50 px-2.5 py-0.5 font-medium text-emerald-800 text-xs dark:border-emerald-800/60 dark:bg-emerald-950/40 dark:text-emerald-200">
+        <span aria-hidden className="size-1.5 rounded-full bg-emerald-500" />
+        Sending
+      </span>
+    );
+  }
+
+  return (
+    <span className="inline-flex shrink-0 items-center gap-1.5 rounded-full border border-border bg-background px-2.5 py-0.5 font-medium text-muted-foreground text-xs">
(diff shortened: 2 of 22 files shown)


apps/server/src/index.ts
```diff
@@ -1,9 +1,24 @@
...
 import { ensureProcessPath } from "./lib/ensure-process-path";
 
 ensureProcessPath();
+installErrorHandlers("server");
+
+/**
+ * Only the server drains the queue. Four processes share one config dir, and the
+ * queue is a read-modify-write on a single file, so draining from all of them would
+ * race. The workers are spawned by the server, so it is the one that is always up.
+ * ponytail: single drainer, give the queue per-process files if a worker ever runs
```
vincentri, line 20: The  prefix reads as a self-authored TODO marker (the rest of the comment reads cleanly on its own). Either drop the prefix or, if the per-process-queue refactor is actually planned, surface it as a tracking comment on  rather than inside the server entry file.

apps/web/src/components/ErrorTrackingSettingsCard.tsx
```diff
@@ -0,0 +1,212 @@
...
+} from "@/components/ui/input-group";
+import { Spinner } from "@/components/ui/spinner";
+import {
+  useErrorTrackingSettings,
+  useSaveErrorTrackingSettings,
+  useSendErrorTrackingTest,
+} from "@/hooks/use-error-tracking";
+import { formatError } from "@/lib/client";
+
+function StatusBadge({ configured }: { configured: boolean }) {
```
vincentri: The dashboard already carries `IntegrationStatusHeader` in `apps/web/src/components/integration-settings.shared.tsx:111` for a title + subtitle + status badge. Building a one-off `StatusBadge` here forks that helper rather than extending it; a third card with a third copy of the badge is now the next divergence waiting to happen. If the shared helper does not quite fit (it has 3 states where this card only needs 2), adjusting that helper is cheaper than carving out the alternates every time.

apps/web/src/components/ErrorTrackingSettingsCard.tsx
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  return (
+    <IntegrationCardShell>
+      <div className="flex items-start justify-between gap-4 p-5 pb-4">
+        <div className="min-w-0 space-y-1">
+          <h2 className="font-semibold text-base text-foreground leading-tight [text-wrap:balance]">
+            Error tracking
+          </h2>
+          <p className="text-muted-foreground text-sm leading-snug [text-wrap:pretty]">
```
vincentri: AGENTS.md (React guideline) says do not add descriptive copy beneath headings by default and never restate the heading. The line below restates "Error tracking" ("send this instance's own errors to the error tracker you already run"). The other two sub-blocks (DSN compatibility list and the "Find your DSN" footer) are defensible because they prevent misuse, but this one is editorial — the heading plus the visibility of the section already tells the operator what it is.

vincentri (commented):
## Standards

3 inline comments — see the file/line anchors above.

- **`apps/server/src/index.ts:18`** — `ponytail:` prefix in the queue-drain comment is a personal TODO marker; the rest of the comment is useful but the prefix should go.
- **`apps/web/src/components/ErrorTrackingSettingsCard.tsx:19`** — local `StatusBadge` forks `IntegrationStatusHeader` from `apps/web/src/components/integration-settings.shared.tsx:111`. Divergent change, third card invites the next fork.
- **`apps/web/src/components/ErrorTrackingSettingsCard.tsx:98`** — first subtitle restates the heading "Error tracking". AGENTS.md's React guideline says no editorial copy beneath headings; the DSN-compatibility line and the footer link line are defensible (they prevent misuse), this one is not.

Other notes that do not rise to inline:

- `setErrorTrackingSettings` reads the config twice (save, then `getErrorTrackingSettings` → `loadErrorTrackingSettingsPublic` re-reads the same file). Functionally fine, just slightly wasteful; minor.
- No client-side hook test (`use-error-tracking.ts`) and no `*.test.tsx` for the card. The hook logic is thin and the page renders identical bytes post-cleanup (per the PR test plan), so I do not block on this, but a one-shot hook test like `apps/web/src/hooks/use-thinking-settings.test.ts` would be in keeping with the surrounding pattern.
- `bun x ultracite check` and `bun run knip` both run clean on the modified files in this worktree.

## Spec

Spec source: issue #441 Phase 3, refs the PR.

Phase 3 acceptance criteria coverage:

- "DSN is saved from the Integrations tab" → implemented (PUT `/v1/settings/error-tracking`, UI in `IntegrationsPage`).
- "masked in the API response" → `loadErrorTrackingSettingsPublic()` masks via `maskSecret`; `error-tracking.test.ts:84-86` asserts no `publickey` and no full DSN leak.
- "writable by an org admin only" → routes use `requireOrgAdminOrPlatformAdminFromContext`. The implementation is **broader than the AC text**: platform admin can also write. This matches the dominant pattern across `routes/models.ts` (telegram, discord, composio all do the same), and per the AGENTS.md role table platform admin is a strictly stronger role, so I do not flag it as a violation — but it is worth being explicit about in the Phase 3 AC list or the route comment.
- "Saving takes effect without restarting the server" → `refreshErrorTrackingEnabled()` runs after `saveErrorTrackingDsn` (`agent-service.ts:1153`); the file-level test that previously proved this directly was removed (now covered at the route level).
- "Send test event" button reports success or failure → implemented; `error-tracking.test.ts:133-146` covers the 400-before-DSN case.
- "An empty DSN sends nothing, contacts no endpoint, and writes no queue file" — covered at the core layer (`error-tracking-config.test.ts:58` and `error-tracking-sentry.test.ts:167`), but **not** at the route layer. A `setErrorTrackingSettings({ dsn: "" })` → empty-body response from the same PUT path would round out Phase 3 AC coverage and is one test long.

Out of scope (correctly deferred to #449): Phase 4 ("a tool failure past its retry budget reaches the same tracker, tagged by tool") — confirmed not started.

Docs gap (soft): `docs/website/content/docs/integrations.mdx` lists 6 of the 9 sections visible in the Integrations tab; the new "Error tracking" entry is not added. Per AGENTS.md's docs section this is a user-facing feature and would earn a one-liner ("Error tracking: paste a Sentry-compatible DSN so crash reports from this instance ship to your own tracker"). Not blocking — the page itself is a curated overview rather than an exhaustive list — but the PR description markets this as the "operator can finally do this without a text editor" moment, so the page should catch up.

## Summary

- Standards: 3 findings (worst: local `StatusBadge` forks the shared `IntegrationStatusHeader` — small Divergent Change, third card waits).
- Spec: 0 hard violations (worst: route-level test doesn't cover the empty-DSN clear path that the Phase 3 AC explicitly calls out, and platform-admin write is broader than the AC text but matches existing repo conventions).
- Verdict: comment. Draft + on top of an unmerged stack, so a `--comment` is the right shape; lift to `--request-changes` only if the platform-admin write widening is decided to be a spec violation.

apps/server/src/index.ts
fajarhide: Keeping this one. `ponytail:` is not a self-authored marker here, it is already the repo's convention for a deliberate ceiling plus its upgrade path, with six on `main` today:

```
apps/platform/telegram/src/chat-handler.ts:874
apps/platform/whatsapp/src/socket.ts:203
packages/core/src/attachments/store.ts:56
packages/core/src/automation-scheduler.ts:5
packages/core/src/whatsapp-config.ts:347
packages/db/src/migrate.ts:866
```

A `TODO` says somebody meant to come back. This says the opposite: the single drainer is the decision, and the per-process queue is what to do if a worker ever runs without a server. Dropping the prefix would leave that comment looking like an unexplained constraint rather than a priced one.

Worth noting the line lives in #443, not here. It shows up in this diff because this PR is stacked on that one.

apps/web/src/components/ErrorTrackingSettingsCard.tsx
fajarhide: Fixed in d4c9a8a5. The local `StatusBadge` and the hand-rolled title block are gone; the card renders through `IntegrationStatusHeader` like Discord, Telegram and WhatsApp do.

You were right that adjusting the helper was the cheaper half. The three states cover this card's two by treating a saved DSN as connected, so the only change the helper needed was `subtitle?: string` and a conditional render, which no existing caller notices. Net `-36/+47` across the card, the helper and a new route test.

Both screenshots were recaptured against a live instance, since the header they showed no longer ships.

apps/web/src/components/ErrorTrackingSettingsCard.tsx
fajarhide: Gone in d4c9a8a5. The heading plus a badge reading Off or Sending is the whole header now.

The DSN compatibility line and the footer link stay, for the reason you gave: one of them stops an operator assuming Sentry SaaS only, the other stops them hunting for the DSN in the wrong place.

fajarhide:
Thanks @vincentri. Two of the three inline ones landed, the `ponytail:` one I pushed back on in its thread with the six precedents already on `main`.

The non-inline notes:

**Route test for the empty DSN.** Added. You were right that it was the one Phase 3 line proven at the config layer and never through the PUT an operator actually uses. It saves a DSN, clears it with `""`, and asserts `configured: false, dsnMasked: null`.

**Docs entry.** Already there, in #445: `integrations.mdx` gains an Error tracking bullet. That PR also gained the docs hub link and a scrub table that matches the scrubber, from your review on it.

**Double read in `setErrorTrackingSettings`.** Leaving it. Collapsing it means either `loadErrorTrackingSettingsPublic` also owning the cached flag, or `refreshErrorTrackingEnabled` returning the masked view, and both couple two jobs into one function to save a file read that happens when an operator pastes a DSN. Not a path worth the shape.

**No hook or card test.** Declining, and the reason is the maintainer's own recent pass: #458 removed a batch of low-value tests, and a test asserting a thin `useQuery` wrapper calls the client is exactly that shape. The card's behaviour that matters is at the route, which is now covered.

**Platform admin write.** Left as is, for the reason you gave: `models.ts` does the same for Telegram, Discord and Composio, and the role is strictly stronger.

Both screenshots in the body are recaptured, since the header changed under them.

```
bunx ultracite check    clean
bun run knip    clean
```

PR #445, opened 2026-08-23
docs: page for error tracking

fajarhide:
## Summary

Documentation for #441, as a page under Deploy. One commit off `main`, five files, no code: it describes the feature as #442, #443 and #444 leave it, but does not depend on them and can merge in any order.

**Before:** the only description of error tracking is the PR bodies and the source.
**After:** `/error-tracking`, linked from Integrations and listed under Deploy next to backup and restore.

Two sections carry most of the weight, because a self-hoster asks both questions before pasting a DSN and neither answer should live in the source:

- **What gets sent.** The four processes, and an explicit list of what is not sent: no 4xx or 5xx, no tool failures, no user input.
- **What is stripped before it leaves.** A table of the scrubber's real rules, from home paths through to the quoted payload a `JSON.parse` failure prints. It also says the local log is deliberately not scrubbed, and why.

The queue gets a paragraph under the mechanics. "My crash only showed up after I restarted" reads as a bug until you know the process dies before an asynchronous send can finish.

Listed under **Deploy** rather than Extend. It is about operating an install, not giving an agent a new capability.

## Test plan
- [x] `bun run build:docs`: exit 0, `/error-tracking.md` prerendered at 7.4 KB, both screenshots copied into the output.
- [x] Anchor and link sweep, since the build resolves routes and not fragments: every `](/page)` in the new page resolves, and the page uses no `#fragment` at all. The one hit across the whole site, `quickstart.mdx` linking `/docs/`, predates this branch.
- [x] Screenshots are 1280x889 at 92 and 98 KB, inside the 27 to 102 KB range the other pages sit in.
- [x] The new page is in `meta.json`, so it is reachable from the sidebar rather than being an orphan route.

Refs #441.


docs/website/content/docs/error-tracking.mdx
```diff
@@ -0,0 +1,106 @@
...
+
+**Not sent:** HTTP 4xx or 5xx responses, tool failures, failed agent turns, and anything a user typed. A rejected request is not a defect, and reporting every bad password would bury the real crashes.
+
+## What is stripped before it leaves
+
+Stack traces from your machine quote your machine. Every report is scrubbed first:
+
+| Stripped | Replaced with |
+| --- | --- |
+| Home directory paths, on Unix and Windows | `~` |
```
vincentri, line 51: Scrub table is missing two real rules from the implementation. `scrubText` (packages/core/src/error-tracking-scrub.ts) also collapses any `[A-Za-z0-9_-]{32,}` token to `<redacted-token>` (the catch-all for opaque credentials the named patterns miss) and redacts `name@host.tld` to `<email>`. Email is arguably the most sensitive PII on the list — the page telling operators what is and isn't safe to send should name it.

docs/website/content/docs/error-tracking.mdx
```diff
@@ -0,0 +1,106 @@
...
+| Bearer tokens | `Bearer <redacted>` |
+| API keys shaped like `sk-`, `ghp_`, `xox...`, and AWS access keys | `<redacted-key>` |
+| Anything after a variable whose name contains `TOKEN`, `KEY`, `SECRET`, or `PASSWORD` | `<redacted>` |
```
vincentri: The variable-name list understates the actual rule. The regex in `error-tracking-scrub.ts` matches `apikey` / `api_key` and `authorization` / `credential` / `passwd` in addition to the four names called out here. An operator reading this row to predict what stays on the wire will be wrong about `apiKey=` and `Authorization: Bearer …` (the bare header — the `Bearer <token>` form is caught by the row above, but `Authorization: <opaque>` is caught by this row, not the Bearer row).

vincentri (commented):
## Standards

Docs-only PR. Five files, no code. The page is registered in `meta.json` under `---Deploy---`, cross-linked from `integrations.mdx`, both screenshots match the prose, and the build plan in the body is sensible. Style matches neighbouring pages (`backup-restore`, `tailscale`) — task-oriented headings, role tables, screenshot alt-text in the same voice, no extras under headings or labels.

A few soft conventions worth flagging:

- **Not linked from the docs hub.** `CONTRIBUTING.md` (and the in-repo `contributing.mdx`) say a docs-only change "register[s] new MDX pages in `meta.json`, link[s] them from the Documentation hub, then run `bun run build:docs`." The hub (`docs/website/content/docs/docs.mdx`) lists `backup-restore` under its Deploy section but not `error-tracking`. The sidebar still picks it up, so it isn't an orphan, but it isn't a Discoverability win either.
- **No `capture-error-tracking-screenshots.sh`.** `AGENTS.md` says "screenshots under `docs/website/public/screenshots/`, capture scripts in `docs/website/scripts/capture-*.sh`." Existing screenshots are mixed (chat has one, most don't), so this is a judgement call rather than a hard miss.
- **Description frontmatter.** The description says "Sentry, GlitchTip, or Bugsink" while the body lists four (adds "self-hosted Sentry"). Trivial, but the meta description is what search results and SEO snippets pull from.

## Spec

Issue #441 is the spec. Its acceptance criteria are all covered:

- DSN saved from Integrations tab, masked, org admin only → `Setup`, `Who can do what`, and `How it works underneath` all say so. Verified against `error-tracking-config.ts#loadErrorTrackingSettingsPublic` (`maskSecret`) and the `requireOrgAdminOrPlatformAdminFromContext` guard on PUT and on POST `/test`.
- Takes effect without restart → "The DSN is read per send, not at startup."
- Empty DSN sends nothing → "What you need first" + "Turning it off" (plus the env-var empty-means-off detail under the mechanics).
- `DO_NOT_TRACK=1` beats a saved DSN → covered; verified against `resolveErrorTrackingDsn`.
- Scrubbing → table at lines 49–55. See inline comments — the table understates the actual scrub rules (see below).
- Send test event button → step 3 + screenshot.
- Failed send does not change exit code → "Delivery failure is swallowed. … exit codes are never changed by it."
- Uncaught exception delivered via the queue → line 96, accurate down to the "Bun 1.3.14: not even an `exit` handler runs" measurement claim.

Out-of-scope items from #441 are also respected (no error UI inside Nakama, no HTTP 5xx events — the doc explicitly calls out "Not sent: HTTP 4xx or 5xx responses").

Two accuracy gaps against the implementation rather than against the AC list:

- **Scrub table is missing two real rules.** `scrubText` in `packages/core/src/error-tracking-scrub.ts` also redacts `name@host.tld` → `<email>` and any opaque `[A-Za-z0-9_-]{32,}` token → `<redacted-token>`. The doc covers neither. See inline comment on line 51.
- **Variable-name row is narrower than the regex.** The pattern matches `apikey` / `api_key`, `authorization`, `credential`, and `passwd` in addition to the four names listed. An operator trying to predict what survives from `apiKey=…` or `Authorization: <opaque>` will read this row wrong. See inline comment on line 54.

These are accuracy refinements, not lost spec coverage, so a `--comment` verdict is appropriate rather than `--request-changes`.

## Summary
- Standards: 3 findings (worst: docs hub link missing per the documented docs-only-PR convention).
- Spec: 2 findings (worst: scrub table understates the actual rules — see inline comments).

docs/website/content/docs/error-tracking.mdx
fajarhide: Both added in 66524bc2. Verified against `SECRET_PATTERNS` rather than from memory: the email rule is line 17 and the 32-character catch-all is line 20, and the table listed five of the seven.

You are right about which one matters most. Email was the only rule in the scrubber that removes personal data rather than a credential, and it was the one missing from the page telling operators what is safe to send.

The catch-all row says what it is for, so a reader understands why an ordinary-looking id in a stack frame can come back as `<redacted-token>`.

docs/website/content/docs/error-tracking.mdx
fajarhide: Fixed in 66524bc2. The row now lists all nine the regex matches: `TOKEN`, `KEY`, `API_KEY`, `APIKEY`, `SECRET`, `PASSWORD`, `PASSWD`, `AUTHORIZATION`, `CREDENTIAL`.

Your split of the two `Authorization` cases is right and is why this mattered: `Bearer <token>` is caught by the Bearer row, and a bare `Authorization: <opaque>` by this one. An operator reading the old four names would have predicted neither.

fajarhide:
Thanks @vincentri, all five acted on.

Both scrub findings are fixed and answered in their threads. The other three:

**Docs hub.** Added under Deploy, next to Backup and restore. `meta.json` alone only feeds the sidebar, which is what made this easy to miss.

**Capture script.** Written and checked in as `capture-error-tracking-screenshots.sh`, and it earned itself immediately: the review on #444 removed the card's hand-rolled status badge and its subtitle, so both screenshots were showing a header that no longer ships. Recaptured with it, against a throwaway `NAKAMA_CONFIG_DIR` and a local ingest that accepts the event, with the test event driven by clicking the button because the result line is component state. It needs #444 to run, since the card it photographs is not on `main` yet.

It also records the trap that cost two runs here: with no provider configured the `SetupGuard` redirects every route to the setup wizard, so the capture quietly photographs step 3 of the wizard instead of the Integrations tab.

**Frontmatter.** Widened to the four platforms the body lists.

`bun run build:docs` passes.

PR #454, opened 2026-08-23
feat(server): retry custom tool failures at most twice with backoff

### LINKED ISSUE #449: Retry custom tool failures at most twice with backoff, the prerequisite for #441 phase 4
### Summary

Custom tool execution has no retry. Add at most two retries with backoff around the handler call, then return the controlled error that already exists.

#441 names this as the prerequisite for its phase 4: "Route a tool failure that has exhausted its retries into the same sink ... depends on the retry and backoff work existing first." There is no retry budget to be past yet, so phase 4 cannot start.

### Why it matters

A single transient failure is handed to the user as a final answer today. Verified on a throwaway instance at 827d2b9d: a Python tool that exceeds its budget returns immediately, with no second attempt.

```json
{"ok":false,"error":"Python tool timed out after 8000ms (exit code null): (no stderr)"}
```

The controlled-message half of the behaviour is already right. Only the retry is missing, and it is the half that turns a flaky network call inside a tool into a visible failure.

`apps/server/src/services/custom-tool-handlers.ts` is the seam. Both loaders resolve through `getCustomToolHandler`, so a wrapper there covers JavaScript and Python in one place instead of once per loader.

`context.signal` must keep aborting immediately. A cancelled turn that then retries twice would hold the session open past the cancel, which is the opposite of what the signal is for.

### Acceptance criteria

- [ ] A handler that throws is retried at most twice, with a backoff between attempts
- [ ] A handler that succeeds on attempt two or three returns `ok: true` and the caller sees no error
- [ ] The final failure keeps its current message shape, so existing callers are unaffected
- [ ] An aborted `context.signal` stops immediately and is never retried
- [ ] Retry applies to both handler types through the shared seam, not once per loader
- [ ] A test asserts the attempt count, not just the final result

### One open question, which is why this is `stage: later`

Does a **timeout** count as a retryable failure? The strategy sentence behind #441 says "on tool timeout/error", which reads as yes. But the Python timeout is a wall-clock budget, so retrying twice makes the worst case three times the configured budget. A tool with `NAKAMA_PYTHON_TOOL_TIMEOUT_MS` at 30s would hold a turn for 90s before answering.

Three ways out, and the answer changes the shape enough that it should be settled before anyone sizes this: retry errors but never timeouts, retry timeouts against one shared deadline rather than per attempt, or keep the per-attempt budget and lower the default. I have no strong preference and it is the maintainer's call.

puspoaditya:
Closes #449

## What

Wraps custom tool execution (both JavaScript and Python) in a retry policy at the shared `getCustomToolHandler` seam: at most **two retries** with exponential backoff (500ms, then 1s) around the handler `run`.

- **Thrown errors and timeouts are retried**; the final failure is re-thrown unchanged, so existing callers keep the current error message shape.
- **`context.signal` aborts stop immediately** and are never retried — including mid-backoff (the backoff sleep is abort-aware).
- **Error tools are naturally not retried**: config/load failures produce a `createErrorTool` whose `run` returns `{ error }` instead of throwing.
- The seam wraps `load`, so both handler types are covered in one place (no per-loader duplication).

## Decision on the open question (timeout retryability)

The #441 strategy sentence reads "on tool timeout/error", and the acceptance criteria do not distinguish, so I implemented **retrying both errors and timeouts** (2 retries, 3 attempts total — worst case 3× the configured budget). If a shared-deadline approach is preferred instead, the wrapper is a single function (`withToolRetries`) and the change is small.

## Tests (assert behavior, not copy)

`apps/server/src/services/custom-tool-handlers.test.ts` — 8 tests:
- success on first attempt (no extra calls)
- retries at most twice, succeeds on attempt 3 (**attempt count asserted**)
- stops after two retries and re-throws the last error unchanged
- abort during run → stops immediately, never retried
- already-aborted signal → run never starts
- abort during backoff → retry cancelled
- seam applies to both `javascript` and `python` handlers
- unknown handler type → null

```
8 pass / 0 fail
```

Verified locally: `bun test` (8/8), `ultracite check` clean.

### DIFF
--- a/apps/server/src/services/custom-tool-handlers.test.ts
+++ b/apps/server/src/services/custom-tool-handlers.test.ts
@@ -1,0 +1,120 @@
+import { describe, expect, test } from "bun:test";
+import type { ToolContext } from "@nakama/core";
+import {
+  CUSTOM_TOOL_HANDLERS,
+  getCustomToolHandler,
+  TOOL_RETRY_LIMIT,
+  withToolRetries,
+} from "./custom-tool-handlers";
+
+function ctx(signal?: AbortSignal): ToolContext {
+  return { signal };
+}
+
+describe("withToolRetries", () => {
+  test("succeeds on the first attempt without extra calls", async () => {
+    let attempts = 0;
+    const run = async () => {
+      attempts += 1;
+      return { ok: true };
+    };
+
+    const result = await withToolRetries(run)({}, ctx());
+
+    expect(result).toEqual({ ok: true });
+    expect(attempts).toBe(1);
+  });
+
+  test("retries at most twice and returns success when a later attempt succeeds", async () => {
+    let attempts = 0;
+    const run = async () => {
+      attempts += 1;
+      if (attempts < 3) {
+        throw new Error("transient");
+      }
+      return { ok: true };
+    };
+
+    const result = await withToolRetries(run)({}, ctx());
+
+    expect(result).toEqual({ ok: true });
+    expect(attempts).toBe(3);
+  });
+
+  test("stops after two retries and re-throws the last error unchanged", async () => {
+    let attempts = 0;
+    const message =
+      "Python tool timed out after 8000ms (exit code null): (no stderr)";
+    const run = async () => {
+      attempts += 1;
+      throw new Error(message);
+    };
+
+    await expect(withToolRetries(run)({}, ctx())).rejects.toThrow(message);
+    expect(attempts).toBe(TOOL_RETRY_LIMIT + 1);
+  });
+
+  test("an aborted signal during the run stops immediately and is never retried", async () => {
+    let attempts = 0;
+    const controller = new AbortController();
+    const run = async () => {
+      attempts += 1;
+      controller.abort(new Error("cancelled"));
+      throw new Error("boom");
+    };
+
+    await expect(
+      withToolRetries(run)({}, ctx(controller.signal))
+    ).rejects.toThrow("boom");
+    expect(attempts).toBe(1);
+  });
+
+  test("an already-aborted signal never starts the run", async () => {
+    const controller = new AbortController();
+    controller.abort(new Error("cancelled"));
+    let attempts = 0;
+    const run = async () => {
+      attempts += 1;
+      return { ok: true };
+    };
+
+    await expect(
+      withToolRetries(run)({}, ctx(controller.signal))
+    ).rejects.toThrow("cancelled");
+    expect(attempts).toBe(0);
+  });
+
+  test("aborting during the backoff cancels the retry", async () => {
+    let attempts = 0;
+    const controller = new AbortController();
+    const run = async () => {
+      attempts += 1;
+      if (attempts === 1) {
+        throw new Error("transient");
+      }
+      return { ok: true };
+    };
+
+    const pending = withToolRetries(run)({}, ctx(controller.signal));
+    setTimeout(() => controller.abort(), 25);
+
+    await expect(pending).rejects.toThrow();
+    // Never reached the second attempt.
+    expect(attempts).toBe(1);
+  });
+});
+
+describe("getCustomToolHandler seam", () => {
+  test("applies the retry wrapper to both javascript and python handlers", () => {
+    for (const type of ["javascript", "python"] as const) {
+      const viaSeam = getCustomToolHandler(type);
+      expect(viaSeam).not.toBeNull();
+      // The seam returns a wrapped load, distinct from the raw loader.
+      expect(viaSeam!.load).not.toBe(CUSTOM_TOOL_HANDLERS[type].load);
+    }
+  });
+
+  test("returns null for unknown handler types", () => {
+    expect(getCustomToolHandler("ruby")).toBeNull();
+  });
+});
--- a/apps/server/src/services/custom-tool-handlers.test.ts
+++ b/apps/server/src/services/custom-tool-handlers.test.ts
@@ -1,5 +1,9 @@
 import { describe, expect, test } from "bun:test";
+import { mkdir, mkdtemp, rm, writeFile } from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
 import type { ToolContext } from "@nakama/core";
+import type { StoredToolRecord } from "@nakama/db";
 import {
   CUSTOM_TOOL_HANDLERS,
   getCustomToolHandler,
@@ -9,6 +13,23 @@ import {
 
 function ctx(signal?: AbortSignal): ToolContext {
   return { signal };
+}
+
+const originalConfigDir = process.env.NAKAMA_CONFIG_DIR;
+
+function makeRecord(
+  overrides: Partial<StoredToolRecord> = {}
+): StoredToolRecord {
+  return {
+    createdAt: new Date().toISOString(),
+    description: "Retry probe",
+    handlerConfig: { modulePath: "flaky.py" },
+    handlerType: "python",
+    id: "tool_flaky",
+    name: "flaky",
+    updatedAt: new Date().toISOString(),
+    ...overrides,
+  };
 }
 
 describe("withToolRetries", () => {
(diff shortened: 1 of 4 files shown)


fajarhide:
@puspoaditya The seam choice is right. `getCustomToolHandler` is the one place both loaders
resolve through, so wrapping `load` there costs nothing per loader, and the
abort handling is real rather than just claimed in the description.

What blocks: Python never retries. `runPythonTool` catches everything and
returns `{ error: message }` (`python-tool-loader.ts:118-124`), so nothing ever
throws for `withToolRetries` to catch. I checked out the branch and ran a
python tool that appends a line to a file, then sleeps past an 800ms budget,
alongside a JS tool that throws:

```
python  -> {"error":"Python tool timed out after 800ms (exit code null): (no stderr)"}
         spawns=1 elapsed=813ms
js      -> threw boom
         calls=3 elapsed=1506ms
```

One spawn, which is the exact failure #449 was filed on. The seam test passes
because it asserts `load` was wrapped, not that a python failure gets a second
attempt.

The cheapest fix I found is a deletion: drop the try/catch in `runPythonTool`
and let `spawnAndParse` reject. `executeToolCall` already turns a throw into
`{ error: error.message }` (`packages/agent/src/tool-loop.ts:45`), the same
expression, so callers see the same shape. With it removed the probe reads:

```
python  -> {"error":"Python tool timed out after 800ms (exit code null): (no stderr)"}
         spawns=3 elapsed=3954ms
```

It costs four assertions in `python-tool-loader.test.ts` that expect a returned
error and would expect a throw instead. Retrying on a returned
`{ error: string }` is the other route, but that also retries the
`createErrorTool` config failures, which are deterministic, so I would take the
deletion.

Second, 3954ms for an 800ms budget is the open question the issue carries
`stage: later` for. At the default 30s timeout, a retried timeout holds a turn
for over 90 seconds. Your reading is defensible and you flagged it in the
description, but it is @ahmadrosid's call, so I would rather have it answered
than assumed.

Smaller, not blocking: `abortableDelay` and `abortReason` are
`node:timers/promises`. `await delay(ms, undefined, { signal })` rejects on
abort and on an already-aborted signal, verified on Bun 1.3.14, and drops about
25 lines. What you lose is `signal.reason` passthrough, which only your own
test reads.

What I did not check: no live instance, just the loaders in a worktree, and
nothing about how a retried tool interacts with turn cancel beyond the signal
test you already wrote.

puspoaditya:
Thanks for the detailed review — the empirical python-vs-js probe was exactly the right call. All three points addressed:

**1. Python now actually retries (blocking fix)** — removed the try/catch in `runPythonTool` so `spawnAndParse` rejects; `executeToolCall` (tool-loop.ts:45) already converts the throw into `{ error: message }`, so callers see the same shape. Updated the 4 loader assertions that expected a returned error to expect a throw instead.

**2. Added the test you asked for** — a functional test through the seam: a python module persists an attempt counter in the workspace, fails (exit 3) until count >= 3, then succeeds. It asserts `attempts === 3`, proving a python failure gets two real re-spawns, not just a wrapped `load`:

```
(pass) getCustomToolHandler seam > python handler failures are actually retried through the seam [1627ms]
```

**3. `abortableDelay` → `node:timers/promises`** — dropped the custom sleep. One note: on Bun 1.3.14 the export is `setTimeout` (there is no named `delay` export), so it's `import { setTimeout as delay } from "node:timers/promises"` with the `{ signal }` option.

**Open question for @ahmadrosid:** retrying a timeout at the default 30s budget means a worst-case ~90s turn hold (3× budget). Options: (a) keep retrying timeouts as today, (b) retry errors only, never timeouts, (c) retry timeouts against one shared deadline. Happy to switch the wrapper to whichever you prefer — it's a small change in `withToolRetries`.

apps/server/src/services/custom-tool-handlers.ts
```diff
@@ -42,13 +47,86 @@ export const CUSTOM_TOOL_HANDLERS = {
...
+ * (up to 3 attempts total) before the last error is surfaced unchanged.
+ */
+export const TOOL_RETRY_LIMIT = 2;
+
+/**
+ * Base backoff between attempts; each retry doubles it: 500ms, then 1s.
+ */
+const TOOL_RETRY_BASE_DELAY_MS = 500;
+
+function abortReason(signal: AbortSignal): unknown {
```
vincentri, line 61: `abortReason` is declared twice in the same module scope — once here at line 61 and again immediately below at line 67 with an identical body. Biome flags this as `lint/suspicious/noRedeclare`, so `bun x ultracite check` (which AGENTS.md says to run before committing) reports an error and CI fails. The second copy is dead code from a paste; delete one of them. The same file also trips `assist/source/organizeImports` and the formatter — `bun x ultracite fix` will clean those up automatically.

apps/server/src/services/custom-tool-handlers.test.ts
```diff
@@ -0,0 +1,201 @@
...
+  });
+});
+
+describe("getCustomToolHandler seam", () => {
+  test("applies the retry wrapper to both javascript and python handlers", () => {
+    for (const type of ["javascript", "python"] as const) {
+      const viaSeam = getCustomToolHandler(type);
+      expect(viaSeam).not.toBeNull();
+      // The seam returns a wrapped load, distinct from the raw loader.
+      expect(viaSeam!.load).not.toBe(CUSTOM_TOOL_HANDLERS[type].load);
```
vincentri, line 134: The acceptance criterion "Retry applies to both handler types through the shared seam, not once per loader" only has end-to-end coverage for Python. The structural check on line 134 just compares `viaSeam.load` to the raw loader by identity — it doesn't prove a JavaScript handler actually retries through the seam. A symmetric end-to-end test (a `flaky.js` module that succeeds on attempt 3, asserting the retry count) would close the asymmetry.

vincentri (changes requested):
## Standards

Two documented-standard violations in `apps/server/src/services/custom-tool-handlers.ts` (also cited inline on line 61):

1. **`lint/suspicious/noRedeclare` — duplicate `abortReason` declaration.** The helper is declared twice in the same module scope (lines 61 and 67) with identical bodies. Biome flags it as a redeclaration, so `bun x ultracite check` (which AGENTS.md says to run before committing) reports an error and CI fails. The second copy is dead code from a paste — delete one. Verified by running `bun x ultracite check apps/server/src/services/custom-tool-handlers.ts` against the PR head, which fails with three errors.
2. **Unorganised imports + format drift on the same file.** `assist/source/organizeImports` and the Biome formatter also flag this file (auto-fixable via `bun x ultracite fix`, but the PR claims `ultracite check` clean, so it wasn't run).

Soft notes (smells, not violations):
- `abortReason` returns `unknown`, but the body is `Error | signal.reason` after the `instanceof` narrowing — the return type could be `Error` rather than `unknown`. Minor.
- The "aborting during the backoff" test uses a `setTimeout(..., 25)` magic number. Robust against the 500ms first backoff, but a named constant or a deterministic signal would read better.

## Spec

Issue #449 acceptance criteria, mapped to the PR:

- ✅ A handler that throws is retried at most twice, with a backoff between attempts.
- ✅ A handler that succeeds on attempt two or three returns `ok: true`; caller sees no error.
- ✅ The final failure keeps its current message shape (`executeToolCall` in `packages/agent/src/tool-loop.ts` still wraps thrown errors as `{ error: message }`).
- ✅ An aborted `context.signal` stops immediately and is never retried, including mid-backoff.
- ⚠️ **Partial:** "Retry applies to both handler types through the shared seam, not once per loader." Python has an end-to-end retry test (`python handler failures are actually retried through the seam`, asserts the attempt count from the spawned child). JavaScript only has a structural assertion that `viaSeam.load !== CUSTOM_TOOL_HANDLERS.javascript.load` — it does not prove retries actually reach the JS loader. A symmetrical JS end-to-end test (a `flaky.js` that succeeds on attempt 3) would close the gap. Cited inline on `custom-tool-handlers.test.ts:134`.
- ✅ A test asserts the attempt count, not just the final result (Python end-to-end + `withToolRetries` unit).

Scope is otherwise tight — no extra public API beyond `TOOL_RETRY_LIMIT` and `withToolRetries`, and the error-tool behaviour is preserved correctly because `createErrorTool` returns (not throws) and the wrapper passes the value through.

## Summary
- Standards: 2 findings (worst: duplicate `abortReason` declaration — `lint/suspicious/noRedeclare`, CI-blocking)
- Spec: 1 finding (worst: AC #5 only has Python end-to-end retry coverage; JS coverage is structural)

puspoaditya:
Thanks for the review, @vincentri — both points are addressed in follow-up PR #472: the duplicate `abortReason` is removed (ultracite check is clean again) and a symmetric `flaky.js` end-to-end retry test now covers the javascript handler through the seam, matching the python one (10/10 tests passing).

PR #632, opened 2026-08-26
fix(settings): confirm before removing LLM provider

### LINKED ISSUE #625: settings(web): removing an LLM provider needs confirmation (data loss on sole provider)
## Summary

`Remove` on an LLM provider row in Settings (`provider-instance-card.tsx`, handler `handleDelete` in `use-provider-instance-card.ts`) deletes the provider instantly with **no confirmation dialog**, even when it is the only/default provider.

Deleting the sole provider drops its stored API key from user config with zero friction; recovery requires editing `~/.nakama/config.ini` manually.

**Measured on:** `cf3d6ac6`

## Reproduction

1. Open `/settings` → LLM providers
2. Click **Remove** on the only provider row
3. Row disappears immediately — no prompt

Observed during scripted QA: the sole OpenRouter provider (with real API key) was removed by a single click.

## Expected vs actual

Expected (actual output captured at the time):

```
click Remove → row deleted immediately, no dialog, key unrecoverable from UI
```

Actual:

```
expected: confirm dialog; refuse or warn when provider is last/default
got: instant deletion
```

Note: automation delete already confirms ("Delete automation? … permanently") — providers are the inconsistent case, and they carry credentials.

## Root cause / proposed fix

`handleDelete()` calls `onDelete(instance.id)` directly. Add a `window.confirm` guard (pattern exists in `McpTab.tsx`), with an extra warning line when the target is the default/last provider.

Security: [x] touches credential handling (stored provider API keys)

itw-code:
## Summary

Remove an LLM provider from Settings → confirm dialog first; warn extra when it is your only/default provider. Fixes #625

**Before:** one click on Remove deleted the provider instantly — sole-provider removal silently dropped its stored API key.
**After:** `handleDelete()` in `use-provider-instance-card.ts` guards with a confirm that names label/type and prepends "This is your only/default LLM provider." when applicable.

Why safe:
1. Same pattern already ships in `McpTab.tsx` deletes
2. UI-only — server `DELETE /v1/providers/:id` and all auth paths untouched (3 files, +26 LOC)
3. Dismiss/cancel keeps the row exactly as before, verified live

Only re-add / follow up if product wants an in-app dialog instead of `window.confirm` for a11y consistency.

## Test plan
- [x] `bun x tsc --noEmit -p apps/web` clean; `bun run knip` no findings
- [x] Playwright: Remove → dialog appears → dismiss → row intact; Update-key & Manage-models dialogs unaffected (`qa-evidence/F-011/verify/`)
- [ ] Settings → Remove on any provider → Cancel keeps it; message warns when it is the last one

---

HARNESS

### DIFF
--- a/apps/web/src/components/settings/ProviderSettingsCard.tsx
+++ b/apps/web/src/components/settings/ProviderSettingsCard.tsx
@@ -104,6 +104,10 @@ export function ProviderSettingsCard({
                     <ProviderInstanceCard
                       catalog={catalog}
                       instance={instance}
+                      isDefaultProvider={
+                        providersResponse?.defaultProviderId === instance.id
+                      }
+                      isLastProvider={providers.length === 1}
                       key={instance.id}
                       onDelete={async (providerId) => {
                         await deleteProviderMutation.mutateAsync(providerId);
--- a/apps/web/src/components/settings/use-provider-instance-card.ts
+++ b/apps/web/src/components/settings/use-provider-instance-card.ts
@@ -35,6 +35,8 @@ export function useProviderInstanceCard({
   onUpdate,
   onDelete,
   onError,
+  isDefaultProvider = false,
+  isLastProvider = false,
 }: {
   instance: ProviderInstanceSummary;
   catalog: ProviderModelOption[];
@@ -44,6 +46,8 @@ export function useProviderInstanceCard({
   ) => Promise<void>;
   onDelete: (providerId: string) => Promise<void>;
   onError: (error: string | null) => void;
+  isDefaultProvider?: boolean;
+  isLastProvider?: boolean;
 }) {
   const [replaceKeyOpen, setReplaceKeyOpen] = useState(false);
   const [editOpen, setEditOpen] = useState(false);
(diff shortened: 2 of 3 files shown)


apps/web/src/components/settings/use-provider-instance-card.ts
```diff
@@ -157,6 +161,18 @@ export function useProviderInstanceCard({
...
+      isLastProvider || isDefaultProvider
+        ? "This is your only/default LLM provider. "
+        : "";
+    const confirmed = window.confirm(
+      `${soleWarning}Remove ${instance.label} (${instance.type})? Models using this provider will stop working. This cannot be undone.`
+    );
+
+    if (!confirmed) {
+      return;
+    }
```
ahmadrosid, line 174: `shrink:` `soleWarning` + `confirmed` locals. Same as `McpTab.tsx`:

```ts
if (
  !window.confirm(
    `${isLastProvider || isDefaultProvider ? "This is your only/default LLM provider. " : ""}Remove ${instance.label} (${instance.type})? Models using this provider will stop working. This cannot be undone.`
  )
) {
  return;
}
```

apps/web/src/components/settings/ProviderSettingsCard.tsx
```diff
@@ -104,6 +104,10 @@ export function ProviderSettingsCard({
                     <ProviderInstanceCard
                       catalog={catalog}
                       instance={instance}
+                      isDefaultProvider={
+                        providersResponse?.defaultProviderId === instance.id
+                      }
+                      isLastProvider={providers.length === 1}
```
ahmadrosid, line 110: `yagni:` `isDefaultProvider` and `isLastProvider` are threaded through 3 files, then OR’d into one prefix. One prop is enough:

```tsx
isSole={providers.length === 1 || providersResponse?.defaultProviderId === instance.id}
```

ahmadrosid (commented):
Ponytail review (complexity only).

Two flags + two locals for one confirm prefix. Collapse to one prop and `if (!window.confirm(...)) return;` like `McpTab.tsx`.

`net: -10 lines possible.`

PR #643, opened 2026-08-26
fix(cli): strip cursor probes in a single pass

### LINKED ISSUE #623: [cli] consumeTerminalInput cursor-probe regex order is non-deterministic
## Bug
The cursor-position report regex runs on `this.pending` then `replace(/\x1b\[\d+;\d+R/g, "")`. If multiple reports arrive in one chunk, the second report may be removed by the first replace before the second match fires (but JavaScript runs the first `replace` synchronously, then `consumeTerminalInput` runs on the cleaned string).

## Files
- `apps/cli/src/terminal-input.ts:166-176`

## Fix
Strip the entire matched sequence and the prefix in one pass; or use a separate buffer for cursor probes.

## Severity
Medium — terminal input.

akmalsyrf:
## Summary

Multiple cursor-position reports in one stdin chunk are stripped in a single pass; waiters still get the first row.

**Before:** `match` then a separate global `replace` could drop later CPRs without a deterministic strip/extract pairing.
**After:** `stripCursorPositionReports()` removes every `\x1b[row;colR` in one `replace` and returns the first row for `cursorWaiters`.

Why safe:
1. First-row semantics match the previous `match()` behavior
2. Key/paste/mouse consumption still runs on the cleaned buffer
3. Unit tests cover multi-CPR + no-CPR (closes #623)

Only revisit if a waiter ever needs the *latest* CPR in a burst instead of the first.

## Test plan
- [x] `bun test apps/cli/src/terminal-screen.test.ts` (7 pass)
- [ ] In CLI chat, trigger layout that probes cursor (resize / redraw) — input still accepts keys

Closes #623

### DIFF
--- a/apps/cli/src/terminal-input.ts
+++ b/apps/cli/src/terminal-input.ts
@@ -1,5 +1,24 @@
 const CURSOR_POSITION_REPORT = /^\x1b\[(\d+);(\d+)R$/;
+const CURSOR_POSITION_REPORT_GLOBAL = /\x1b\[(\d+);(\d+)R/g;
 const MOUSE_EVENT_REPORT = /^\x1b\[<\d+;\d+;\d+[mM]$/;
+
+/** Strip all CPR sequences in one pass; return the first report's row. */
+export function stripCursorPositionReports(pending: string): {
+  pending: string;
+  row: number | null;
+} {
+  let row: number | null = null;
+  const cleaned = pending.replace(
+    CURSOR_POSITION_REPORT_GLOBAL,
+    (_match, rowText: string) => {
+      if (row === null) {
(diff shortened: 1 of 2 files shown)


akmalsyrf:
Heads-up after review: #642 now includes this CPR strip plus a corrected stdin encoding restore (the original `setEncoding(null)` path is a no-op in Node/Bun). Please keep this draft and prefer #642 to avoid a `terminal-input.ts` conflict on `main`.

akmalsyrf:
Closing as superseded by #642 (CPR strip + real stdin encoding restore). Keeps `terminal-input.ts` from conflicting on `main`.

PR #647, opened 2026-08-26
fix(cli): bail early on oversized clipboard images

### LINKED ISSUE #610: [cli] clipboard-image.ts no early size bailout for huge clipboard data
## Bug
`getImageBinary()` returns the full bytes; only afterwards is `validateImageAttachments` called which estimates decoded length. A 50 MB clipboard image is fully read into memory, base64-encoded into a ~67 MB string, and only then rejected.

## Files
- `apps/cli/src/clipboard-image.ts:8-21`

## Fix
Check the byte length before `Buffer.from(...).toString("base64")`; bail early with an error.

## Severity
Medium — memory spike.

akmalsyrf:
## Summary

Clipboard paste rejects oversized images on raw bytes before base64 encoding.

**Before:** full clipboard buffer → base64 string → `validateImageAttachments` reject (large memory spike).
**After:** `attachmentFromClipboardBytes` checks `bytes.length > MAX_IMAGE_BYTES` first, then encodes.

Why safe:
1. Same `MAX_IMAGE_BYTES` (5 MB) as file/`validateImageAttachments` paths
2. Small images still validate via existing `validateImageAttachments`
3. Unit tests cover under/over limit (closes #610)

Only revisit if clipboard APIs expose size without loading full bytes (further optimization).

## Test plan
- [x] `bun test apps/cli/src/clipboard-image.test.ts` (2 pass)
- [ ] Paste a normal PNG in CLI chat; oversized paste errors without hanging

Closes #610

### DIFF
--- a/apps/cli/src/clipboard-image.test.ts
+++ b/apps/cli/src/clipboard-image.test.ts
@@ -1,0 +1,21 @@
+import { describe, expect, test } from "bun:test";
+import { MAX_IMAGE_BYTES } from "@nakama/core";
+import { attachmentFromClipboardBytes } from "./clipboard-image";
+
+describe("attachmentFromClipboardBytes", () => {
+  test("encodes small images without base64 ballooning first", () => {
+    const bytes = Buffer.from([0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a]);
+    const attachment = attachmentFromClipboardBytes(bytes);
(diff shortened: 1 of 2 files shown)


akmalsyrf:
Closing as superseded by #649, which includes the early size bailout **and** magic-byte media typing on the same path. Landing both against `main` would conflict on `clipboard-image.ts`.

PR #663, opened 2026-08-27
feat(bash): MicroSandbox backend with 1:1 profile isolation

### LINKED ISSUE #100: feat: add MicroSandbox-backed backend for bash tool execution
## Summary

Add an opt-in MicroSandbox backend for the `bash` tool so each profile runs in its own isolated microVM instead of on the host.

## Isolation model

**1 profile → 1 sandbox** (warm reuse across turns). Profile A cannot access profile B’s filesystem or processes.

## Behavior

- Keep the existing `bash` tool contract (`stdout` / `stderr` / `exitCode` / `timedOut`).
- Backend via config: `host` (default) or `microsandbox`.
- Bind-mount the profile workspace into that profile’s sandbox.
- Deny public network by default (`off` | `public`).
- If `microsandbox` is selected but unavailable → **fail closed** (no host fallback).
- `codingAgent: true` is **unsupported** under `microsandbox` for v1 (clear error). Use `host` for coding-agent until a follow-up.

## Out of scope

- Sandboxing other tool runtimes
- Coding-agent inside the sandbox
- Shared/pooled sandboxes across profiles
- Operator UI for sandbox lifecycle

## Acceptance

- [ ] `bash` works on MicroSandbox with the same I/O contract
- [ ] One warm sandbox per profile; cross-profile isolation holds
- [ ] Workspace writes persist for that profile only
- [ ] Timeout cancels the command; sandbox can stay warm
- [ ] Deployments can choose `host` | `microsandbox`
- [ ] Fail closed when microsandbox is misconfigured
- [ ] Network defaults to deny; `codingAgent` fails clearly under microsandbox
- [ ] Tests cover success, timeout, cwd, env, backend selection, fail-closed, coding-agent gate, cross-profile isolation, network deny

## Plan

`docs/plans/2026-08-27-002-feat-microsandbox-bash-backend-plan.md`

ahmadrosid:
## Summary

Ops can set `NAKAMA_BASH_BACKEND=microsandbox` → each profile runs `bash` in its own warm microVM (1:1 isolation). Closes #100.

**Before:** `bash` always spawned `/bin/bash` on the host inside a path-guarded workspace.
**After:** opt-in MicroSandbox backend with bind-mounted soul dir, deny-network default, fail-closed probe; `codingAgent` stays on `host` only.

Why safe:
1. Default remains `host` — existing installs unchanged
2. Misconfigured microsandbox fails closed (no silent host fallback)
3. Unit tests inject a fake runtime — CI never needs a live microVM

Only re-add / follow up if coding-agent must run inside the sandbox (image + egress), or Docker needs `/dev/kvm` for production microsandbox.

## Test plan
- [x] `bun test apps/server/src/tools/bash.test.ts apps/server/src/tools/bash-sandbox.test.ts` (28 pass)
- [ ] With MicroSandbox installed: set `NAKAMA_BASH_BACKEND=microsandbox`, run ordinary bash; confirm `codingAgent: true` errors clearly

## Post-Deploy Monitoring & Validation

No additional operational monitoring required for default `host` deploys. If enabling microsandbox: watch for fail-closed bash tool errors (`MicroSandbox backend unavailable`) and confirm hypervisor/`/dev/kvm` is present before flipping the env.


fajarhide (commented):
One profile to one sandbox, keyed on a fingerprint so a network or image change recreates it, is the right shape, and injecting the runtime so CI never needs a live microVM is the right call. Two things break the same-I/O-contract criterion from #100, and the env builder does more than it has to.

I checked the branch out, wrote two throwaway probes for the timeout and warm-claim cases, and ran the server suite: 1096 pass, 0 fail with the manager fix below applied. No MicroSandbox installed here, so the test-plan box for a real microVM is still open.

@ahmadrosid the shell one is a product call rather than a bug.

apps/server/src/tools/bash-microsandbox-runtime.ts
```diff
@@ -0,0 +1,121 @@
...
+    try {
+      const result = await handle.collect();
+      return {
+        exitCode: result.code,
+        stderr: truncateOutput(result.stderr()),
+        stdout: truncateOutput(result.stdout()),
+        timedOut: false,
+      };
+    } catch (error) {
+      if (error instanceof ExecTimeoutError) {
```
fajarhide, line 119: On timeout this returns empty stdout and stderr, so a command that printed for 29 minutes shows nothing. The host path keeps it. A `runBash` probe at 800ms gave `{"exitCode":null,"stdout":"before-the-timeout\n","timedOut":true}`. If `handle` still exposes what it buffered before the deadline, returning that keeps the contract.

apps/server/src/tools/profile-sandbox-manager.ts
```diff
@@ -0,0 +1,118 @@
...
+      this.ensured.set(name, fingerprint);
+    }
+
+    const guestCwd = toGuestCwd({
+      guestWorkspace,
+      hostCwd: args.hostCwd,
+      hostWorkspace: args.hostWorkspace,
+    });
+
+    return this.runtime.exec({
```
fajarhide: If the sandbox dies out of band (msb restart, host reboot, a manual remove) `ensured` still holds the fingerprint, so every later call skips `ensure` and fails on connect until the server restarts. My probe: `ensure` ran once across repeated failures. Wrapping this in a try/catch that does `this.ensured.delete(name)` before rethrowing recovers on the next call, and the suite stays at 1096 pass.

apps/server/src/tools/bash-microsandbox-runtime.ts
```diff
@@ -0,0 +1,121 @@
...
+      throw new Error(
+        "MicroSandbox backend unavailable: runtime is not installed. Set NAKAMA_BASH_BACKEND=host or install MicroSandbox. No host fallback."
+      );
+    }
+
+    let builder = Sandbox.builder(args.name)
+      .image(args.image)
+      .detached(true)
+      .workdir(args.guestWorkspace)
+      .shell("/bin/sh")
```
fajarhide, line 48: Host spawns `/bin/bash -lc` at `bash.ts:222`, this runs `/bin/sh -lc` on `alpine`, so busybox ash. `[[ ]]`, arrays and `set -o pipefail` work on one backend and fail on the other. `bashToolDescription()` warns about network and codingAgent but not this, so the model gets a busybox error with no hint.

apps/server/src/tools/bash-sandbox-env.ts
```diff
@@ -0,0 +1,41 @@
...
+export function buildBashSandboxEnv(args: {
+  hostEnv?: NodeJS.ProcessEnv;
+  overrides?: Record<string, string>;
+  workspaceRoot?: string;
+}): Record<string, string> {
+  const host = args.hostEnv ?? process.env;
+  const env: Record<string, string> = {};
+
+  if (host.PATH) {
+    env.PATH = host.PATH;
```
fajarhide: This copies the host `PATH` into the guest. The runtime stage is `oven/bun:1.3-slim` and dev machines are macOS, so the guest inherits directories that do not exist in it and loses alpine's own `/sbin` and `/usr/sbin`. Dropping these two lines lets the image keep its default.

apps/server/src/tools/bash-sandbox-env.ts
```diff
@@ -0,0 +1,41 @@
+const SECRET_ENV_KEY =
+  /(?:^|_)(KEY|TOKEN|SECRET|PASSWORD|PWD|API[_-]?KEY|ACCESS[_-]?KEY[_-]?ID|CREDENTIALS?)$/i;
+
+function isSecretEnvKey(key: string): boolean {
+  return SECRET_ENV_KEY.test(key) || /_URL$/i.test(key);
```
fajarhide, line 5: The base above is already an allowlist of PATH, LANG, HOME and NAKAMA_WORKSPACE_ROOT, so no host env ever reaches this filter. It only sees the caller's own overrides, which means `DATABASE_URL` and `API_URL` are dropped silently here and passed through on host. Same input, different env, no error.

apps/server/src/tools/bash.ts
```diff
@@ -53,9 +68,43 @@ interface ShellRunOptions {
...
+    "./bash-microsandbox-runtime"
+  );
+  sharedSandboxManager = new ProfileSandboxManager(
+    new MicrosandboxBashRuntime()
+  );
+  return sharedSandboxManager;
+}
+
+/** Test helper to clear the process-wide warm-sandbox manager. */
+export function resetBashSandboxManagerForTests(): void {
```
fajarhide, line 89: 9bb459cc says it dropped the test reset, but this is still exported, and `ensuredNamesForTests` at `profile-sandbox-manager.ts:115` is the same shape. The fake runtime already records its own `ensure` calls, which is what the assertion at `bash-sandbox.test.ts:239` is actually after.

apps/server/src/tools/bash-microsandbox-runtime.ts
ahmadrosid: > On timeout this returns empty stdout and stderr, so a command that printed for 29 minutes shows nothing.

Addressed in e6514eec — `exec` now streams events and returns whatever stdout/stderr was buffered before the deadline when `ExecTimeoutError` fires.

apps/server/src/tools/profile-sandbox-manager.ts
ahmadrosid: > If the sandbox dies out of band … every later call skips `ensure` and fails on connect until the server restarts.

Addressed in e6514eec — `run` clears the warm claim on exec failure so the next call re-runs `ensure`. Added a fake-runtime regression test for the recover path.

apps/server/src/tools/bash-microsandbox-runtime.ts
ahmadrosid: > `bashToolDescription()` warns about network and codingAgent but not this, so the model gets a busybox error with no hint.

Addressed differently in e6514eec — kept `/bin/sh` on alpine for now, and added an explicit ash/POSIX note to the microsandbox tool description so the model gets the constraint up front. Switching the guest to real bash (image + `/bin/bash -lc`) is still a product call if we want full host-shell parity.

apps/server/src/tools/bash-sandbox-env.ts
ahmadrosid: > This copies the host `PATH` into the guest. … Dropping these two lines lets the image keep its default.

Addressed in e6514eec — stopped copying host `PATH` into the sandbox env so the guest image keeps its own default.

apps/server/src/tools/bash-sandbox-env.ts
ahmadrosid: > `DATABASE_URL` and `API_URL` are dropped silently here and passed through on host. Same input, different env, no error.

Good catch on the host/sandbox mismatch. Plan KTD7 deliberately strips secret-shaped override keys on the microsandbox path only (and the tests assert that). I want a quick call before changing it — leave this open.

apps/server/src/tools/bash.ts
ahmadrosid: > this is still exported, and `ensuredNamesForTests` … The fake runtime already records its own `ensure` calls

Addressed in e6514eec — removed `ensuredNamesForTests` and assert via `fake.ensures` instead. Kept `resetBashSandboxManagerForTests` because `afterEach` still clears the process-wide singleton so a later path that hits `getSandboxManager()` cannot leak warm state.

ahmadrosid:
> Two things break the same-I/O-contract criterion from #100, and the env builder does more than it has to.

Landed in e6514eec: timeout now returns buffered output, warm claims clear after exec failure, host PATH is no longer copied into the guest, and the microsandbox tool description warns about `/bin/sh`/ash. Left the secret-env override strip open — that one is a deliberate KTD7 choice vs same-I/O parity and needs a quick product call.

fajarhide:
@ahmadrosid LGTM 🚀

apps/server/src/tools/bash-sandbox-env.ts
fajarhide: On this path `overrides` is only what the model passed, and `command` is unfiltered, so `OPENAI_API_KEY=xxx cmd` still works. The strip does not close exfil.

Cheapest fix: drop `/_URL$/i`, keep the KEY/TOKEN/SECRET/PASSWORD regex. `DATABASE_URL` reads as a credential, `API_URL` does not. If it stays as is, name the dropped key in stderr so the model gets a reason.

Fine either way if the call says microsandbox is a restricted backend rather than host parity. Then it's a description change, not a code one. Can you paste KTD7 first? It isn't in the branch.

PR #677, opened 2026-08-27
fix(org-memory): tolerate malformed history entries

### LINKED ISSUE #587: [agent/core] JSON.parse without try/catch on disk-loaded org-memory history entries
## Bug
`JSON.parse(raw) as OrgMemoryChangeLogEntry` will throw if any history file is corrupted, which crashes the entire `listOrgMemoryHistory` / `getOrgMemoryHistoryEntry` calls and prevents the org memory UI from loading.

## Files
- `packages/core/src/soul/org-memory-history.ts:88, 112`

## Fix
Wrap each parse in try/catch, log a warning, and skip the bad entry.

## Severity
Medium — robustness.

mamenesia:
## Summary

Malformed org-memory snapshots no longer break history or Undo.

**Before:** one corrupt metadata file crashed list/direct history reads and prevented the memory timeline from loading.
**After:** malformed metadata is skipped with an ID-only warning, while timeline and Undo resolve the nearest readable state in true chronological order.

Why safe:
1. Only JSON parse failures are isolated; filesystem errors still surface and corrupt files remain available for recovery.
2. Live-memory comparison prevents a skipped snapshot from producing a false Current badge or a no-op Undo.

Only follow up if malformed snapshots should be quarantined automatically instead of preserved.

Closes #587

## Test plan
- [x] `bun test --timeout 120000 <org-memory suites>` (59 pass, 144 assertions)
- [x] Scoped Ultracite + diff check; core/server/web builds and current web type-check
- [ ] Corrupt the newest history `.json`, reload System ΓåÆ Organization, then click Undo latest


ahmadrosid:
Quick ponytail pass — mostly lean.

One thing I’d drop: `org-memory-history-panel.shared.ts`. It’s only used once, and the helper is basically “trim both sides and compare.” Feels cleaner to just do that inline in `OrgMemoryHistoryPanel` and delete the shared file + its test.

Something like:

```ts
const live = memoryData?.content;
const snap = latestRevision?.content;
const latestRevisionIsCurrent =
  live !== undefined && snap !== undefined && live.trim() === snap.trim();
const canUndo =
  live !== undefined &&
  snap !== undefined &&
  changes.length >= (latestRevisionIsCurrent ? 2 : 1);
```

The server-side skip/malformed handling looks good as-is.

ahmadrosid (approved):

PR #700, opened 2026-08-28
fix(server): reject malformed optional JSON

### LINKED ISSUE #551: [server] Multiple routes swallow JSON parse errors with .catch(() => ({}))
## Bug
Each silently returns an empty body when JSON is malformed, bypassing `readJson`'s 400 "Invalid JSON in request body" and letting malformed requests be treated as empty.

## Files
- `apps/server/src/http/routes/profiles.ts:961-963` (clone profile)
- `apps/server/src/http/routes/org-curator.ts:99-103`
- `apps/server/src/http/routes/org-memory.ts:761-763`
- `apps/server/src/http/routes/composio.ts:155-157`

## Fix
Drop the `.catch` so JSON errors bubble as 400 like everywhere else.

## Severity
Medium — error handling.

isnaenihidayat:
## Summary

Send malformed optional JSON → receive 400 before any route mutation; empty optional bodies still use their defaults.

**Before:** Six routes caught every parse failure and silently treated malformed JSON as `{}`.
**After:** `readOptionalJson` falls back only for an empty body and preserves `readJson`'s 400 behavior otherwise.

Why safe:
1. Existing optional-body defaults remain unchanged across clone, curator, memory, Composio, compaction, and email-test routes.
2. Regression coverage proves malformed clone input writes nothing while an empty body still clones successfully.

Only follow up if an optional endpoint intentionally needs to accept non-JSON content.

## Test plan
- [x] `bun test apps/server/src/http/shared.test.ts apps/server/src/http/routes/profiles-clone.test.ts` (6 pass)
- [x] `bun test` (2,612 pass); `bun run check`; `bun run knip`; `bun run build`
- [ ] POST `{` to an affected route and confirm HTTP 400 with no mutation

Closes #551

### DIFF
--- a/apps/server/src/http/routes/composio.ts
+++ b/apps/server/src/http/routes/composio.ts
@@ -14,7 +14,7 @@ import {
   requireNotViewerFromContext,
   requireOrgAdminFromContext,
 } from "../org-guards";
-import { errorResponse, json, readJson } from "../shared";
+import { errorResponse, json, readJson, readOptionalJson } from "../shared";
 import type { HonoApp } from "../types";
 
 export function registerComposioOAuthRoutes(
@@ -152,8 +152,9 @@ export function registerComposioRoutes(
     const auth = requireNotViewerFromContext(c);
 
     try {
-      const body = await readJson<ComposioConnectRequest>(c.req.raw).catch(
-        () => ({})
+      const body = await readOptionalJson<ComposioConnectRequest>(
+        c.req.raw,
+        {}
       );
       return json<ComposioConnectResponse>(
         await service.connectToolkit(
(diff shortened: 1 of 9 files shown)


ahmadrosid (commented):
Small ponytail pass — overall this is lean and the `readOptionalJson` extraction earns its keep with the six call sites.

Two easy trims if you feel like it:

1. **`readOptionalJson` catch block** — `JSON.parse` only throws `SyntaxError` here, so the `instanceof` check + rethrow is just mirroring `readJson` for no gain. Can collapse to a plain `catch { throw new NakamaApiError(...) }`.

2. **`models.ts` import** — the multi-line import sprawl for adding one symbol is noisier than the one-liners in the other route files. Worth keeping consistent.

Nothing else jumped out. Ship when ready.

isnaenihidayat:
Thanks — I applied the readOptionalJson catch simplification in 97ff5d6b.

I tried the one-line models.ts import too, but the repository formatter rejects that line length and rewrites it to multiline, so I kept Ultracite’s canonical output.

The branch is also rebased onto the latest main. Fresh local verification:
- bun test: 2,612 pass, 0 fail
- bun run check: clean
- bun run knip: clean
- bun run build: success

PR #702, opened 2026-08-28
fix(server): scope profile avatar reads to the caller's org

### LINKED ISSUE #508: [security] GET /v1/profiles/:profileId/avatar exposes every profile avatar across orgs (IDOR)
## Bug
`agent.getProfileAvatarByProfileId(profileId)` calls `db.getProfile(profileId)` (no org check) and returns the avatar bytes to anyone. While `getProfile(profileId)` requires `profile.orgId` to be set, there is no check that the caller is a member of that org. Anyone who learns or guesses a profile id can enumerate and fetch every avatar across all orgs.

## Files
- `apps/server/src/http/routes/profiles.ts:872-878`
- `apps/server/src/http/public-routes.ts:28`

## Fix
Require `requireNotViewerFromContext(c)` (or at least authentication) and verify `profile.orgId === auth.activeOrgId` before serving.

## Severity
High — IDOR / info disclosure.

fajarhide:
Closes #508.

`GET /v1/profiles/:profileId/avatar` matched the public-route regex and served bytes through `getProfileAvatarByProfileId`, which read the org off the profile row instead of the caller. No login, no org check: guess a profile id and you get the avatar out of any org.

The route now resolves the active org from the request and calls the already-scoped `getProfileAvatar`. The by-id variant is deleted rather than fixed, so nothing can reach the unscoped read again.

The web UI renders avatars in an `` tag, which sends the session cookie and no `X-Org-Id` header. `resolveOrgId` falls back to the session's `activeOrgId`, so that path still works and the third test pins it.

Before, on `main`:

```
(fail) rejects anonymous callers
  Expected: 401  Received: 200
(fail) hides the avatar from another org
  Expected: 404  Received: 200
1 pass  2 fail
```

After:

```
3 pass  0 fail
```

Full suite: `bun run test`, exit 0, 2603 pass / 0 fail across 11 workspaces.

### DIFF
--- a/apps/server/src/http/app.test.ts
+++ b/apps/server/src/http/app.test.ts
@@ -118,10 +118,6 @@ function createServerOptions() {
         bytes: new Uint8Array([1, 2, 3]),
         mediaType: "image/png",
       }),
-      getProfileAvatarByProfileId: async (_profileId: string) => ({
-        bytes: new Uint8Array([1, 2, 3]),
-        mediaType: "image/png",
-      }),
       getProfileSoulStack: async (_profileId: string) => ({
         stack: ["SOUL.md"],
       }),
--- a/apps/server/src/http/public-routes.ts
+++ b/apps/server/src/http/public-routes.ts
@@ -25,7 +25,6 @@ export function isPublicRouteRequest(
   return (
     PUBLIC_ROUTES.has(pathname) ||
     /^\/v1\/notify\/[^/]+$/.test(pathname) ||
-    (method === "GET" && /^\/v1\/profiles\/[^/]+\/avatar$/.test(pathname)) ||
     (method === "GET" &&
       /^\/v1\/public\/artifact-shares\/[^/]+$/.test(pathname))
   );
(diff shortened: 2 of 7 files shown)


fajarhide:
Reopened and rebased onto main (235f8642). The route is still in the public-route match there, and still reads the org off the profile row rather than off the caller, so the IDOR is live today.

Copied the new test file onto current main to confirm:

```
(fail) rejects anonymous callers          Expected 401, received 200
(fail) hides the avatar from another org  Expected 404, received 200
1 pass  2 fail
```

On this branch, 3 pass 0 fail. Full suite `bun run test` exits 0 with 2630 pass, 0 fail across 11 workspaces.

I also checked whether anything still needs the route public before reopening. `PublicArtifactSharePage.tsx` renders no avatar, and the in-app `` keeps working because `resolveOrgId` falls back to the session's `activeOrgId`. The third test pins that with a cookie and no `X-Org-Id` header.

You closed it without a comment, so I may have missed your reason. If you want it done differently, or not at all, say so and I will rework it or close it myself.

fajarhide:
Closing this again on my side. You closed it once without a comment, so I am taking that as a no rather than pushing it a second time.

The bug itself is still live on main: `GET /v1/profiles/:profileId/avatar` matches the public-route regex and reads the org off the profile row, so any profile id returns an avatar from any org with no session. I left #508 open with that detail. The branch stays on my fork if you ever want the diff.

Issue #596, opened 2026-08-26
[agent/core] mergeCodingAgentSpawnEnv missing dangerous-key denylist


vincentri:
## Bug
The credential-key filter only matches `CODING_AGENT_CREDENTIAL_ENV_KEYS`. `BASH_ENV`/`LD_PRELOAD` are not in the list, so a coding-agent spawn will happily inherit any `BASH_ENV` from `process.env`.

## Files
- `apps/server/src/services/coding-agent-spawn-env.ts:312-337`

## Fix
Always-on denylist for `BASH_ENV`, `ENV`, `BASH_FUNC_*`, `LD_*`, `NODE_OPTIONS`, `PYTHON*`, `IFS`, `PS4`.

## Severity
Medium — sandbox bypass amplification.

fajarhide:
Same finding as #528, same file and same missing denylist. #528 carries more: it also names the bash tool passing `envOverrides` straight through, which is #480's entry point.

Labelled both, but they are one fix in `coding-agent-spawn-env.ts`. Worth closing one of the two so nobody writes the denylist twice.

fajarhide:
Fixed on main by dc5773d2 (#708).

`mergeCodingAgentSpawnEnv` now drops `BASH_ENV`, `IFS`, `PS4`, `SHELLOPTS`, `BASHOPTS`, `NODE_OPTIONS`, `BASH_FUNC_*`, `LD_*`, `DYLD_*` and `PYTHON*` from both `spawnEnv` and `callerEnv`. The filter is always on rather than opt-in, since no caller has a reason to set those. Regression tests live in `apps/server/src/tools/bash.test.ts`.

```
$ bun test src/tools/bash.test.ts
 11 pass  0 fail
```

Two deliberate gaps, so they are on the record rather than forgotten:

- Bare `ENV` is left out. Bash sources it only in posix mode and it is a common application setting.
- Keys already present in the server's own `process.env` still reach the child. That is the operator's environment, not a caller-controlled one.

Issue #528, opened 2026-08-26
[security] mergeCodingAgentSpawnEnv missing dangerous-key denylist


vincentri:
## Bug
The credential-key filter only matches `CODING_AGENT_CREDENTIAL_ENV_KEYS`. The bash tool passes `envOverrides` from the LLM without that filter; even when the flag is set, `BASH_ENV`/`LD_PRELOAD` aren't in the list — a coding-agent spawn will happily inherit any `BASH_ENV` from `process.env`.

## Files
- `apps/server/src/services/coding-agent-spawn-env.ts:312-337`

## Fix
Always-on denylist for `BASH_ENV`, `ENV`, `BASH_FUNC_*`, `LD_*`, `NODE_OPTIONS`, `PYTHON*`, `IFS`, `PS4`.

## Severity
High — sandbox bypass amplification.

fajarhide:
Duplicate of #596, and fixed on main by dc5773d2 (#708). The denylist covers `BASH_ENV`, `IFS`, `PS4`, `SHELLOPTS`, `BASHOPTS`, `NODE_OPTIONS`, `BASH_FUNC_*`, `LD_*`, `DYLD_*` and `PYTHON*`, applied to both `spawnEnv` and `callerEnv`. See #596 for the two gaps left in on purpose.

PR #708, opened 2026-08-28
fix(server): strip shell-hijacking keys from spawn env

### LINKED ISSUE #480: [security] Bash tool env parameter allows arbitrary env injection (RCE)
## Bug

The `bash` tool's `env` parameter is merged into the spawned `/bin/bash -lc` environment with no allowlist. Because the command runs under a login shell, the agent can set:

- `BASH_ENV=/tmp/evil.sh` — sourced before the command runs
- `LD_PRELOAD=…`
- `PATH=…`
- `NODE_OPTIONS=--require=/tmp/evil.js`
- `PYTHONPATH=…`
- `IFS`, `PS4`, etc.

Combined with a one-line `command`, this is full arbitrary-code-execution under the server uid regardless of any other guard on the command itself.

## Files

- `apps/server/src/tools/bash.ts:131-153`

## Impact

- Full server compromise via any prompt-injected tool call
- Steals `NAKAMA_*` secrets, provider API keys, DB session, all org data

## Fix

Strip dangerous keys before merging (`BASH_ENV`, `ENV`, `LD_*`, `NODE_OPTIONS`, `PYTHON*`, `IFS`, `PS4`, `BASH_FUNC_*`). Also call `mergeCodingAgentSpawnEnv(..., { protectCredentialKeys: true })`.

## Severity

Critical — security, RCE.

fajarhide:
Closes #480. Also covers #528 and #596, which are the same missing denylist seen from the coding-agent side.

`runBash` passes the model's `env` map into `mergeCodingAgentSpawnEnv`, which merged it over `process.env` and spawned `/bin/bash -lc`. A login shell sources `BASH_ENV` before it runs the command, so the map decided what executed first. `LD_PRELOAD`, `DYLD_INSERT_LIBRARIES`, `NODE_OPTIONS` and the `PYTHON` keys get there through the loader and the interpreters instead.

The filter lives in `mergeCodingAgentSpawnEnv`, not in `bash.ts`, because `coding-agent-bash-env.ts` routes the same model-supplied map through `callerEnv`. It is always on. The credential filter is untouched: still opt-in, still `callerEnv` only, so the harness bundle keeps setting its own `ANTHROPIC_` and `OPENAI_` values.

Bare `ENV` is left out of the list. Bash sources it only in posix mode, and it is a common application setting, so stripping it costs more than it buys.

Scope, stated plainly: the bash tool runs arbitrary commands by design, so this does not add a boundary against a caller who already reaches the tool. What it closes is the gap where an env key rewrites what a *later* spawn runs, including the coding-agent harness, without appearing in the command string an operator reads.

Before, on `main`:

```
(fail) drops env keys that hijack the shell before the command runs
  Expected to not contain: "HIJACKED"
  Received: "HIJACKED\nkeep=kept preload=/tmp/evil.so node=--require=/tmp/evil.js\n"
```

After:

```
keep=kept preload= node=
24 pass  0 fail
```

The passthrough half of that assertion is deliberate: `KEEP_ME` still reaches the shell, so the filter is not just emptying the map.

Full suite: `bun run test`, exit 0, 2604 pass / 0 fail across 11 workspaces.

### DIFF
--- a/apps/server/src/services/coding-agent-spawn-env.ts
+++ b/apps/server/src/services/coding-agent-spawn-env.ts
@@ -309,6 +309,19 @@ export async function buildSpawnEnvForHarness(
   return buildOpenCodeSpawnEnv(routing, providerType);
 }
 
+/**
+ * Env keys that run code before the spawned command does, so a caller that can
+ * name one owns the process: `bash -lc` sources BASH_ENV, the loader reads
+ * the LD_ and DYLD_ preload keys, node reads NODE_OPTIONS (--require), python
+ * reads the PYTHON keys (sitecustomize), and BASH_FUNC_ smuggles in shell
+ * functions. The bash tool
+ * takes this map straight from the model, so the filter is always on rather
+ * than opt-in. Bare `ENV` is left alone: bash only sources it in posix mode,
+ * and it is a common application setting.
+ */
+const HIJACKING_ENV_KEY =
+  /^(BASH_ENV|IFS|PS4|SHELLOPTS|BASHOPTS|NODE_OPTIONS|BASH_FUNC_.*|LD_.*|DYLD_.*|PYTHON.*)$/;
+
 export function mergeCodingAgentSpawnEnv(
   baseEnv: NodeJS.ProcessEnv,
   spawnEnv: Record<string, string>,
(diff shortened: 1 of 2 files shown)


ahmadrosid:
nice catch

PR #713, opened 2026-08-28
fix(automation): prevent overlapping scheduler polls

### LINKED ISSUE #514: [platform workers] Automation scheduler can run concurrent ticks if previous tick stalls
## Bug
`setInterval` fires the next tick while a previous `await this.scheduler.reload()` is still in flight — JS doesn't wait. Two `reload()` calls can run concurrently, each calling `runAutomationInternal(...)`, which can trigger duplicate automations for the same schedule slot.

## Files
- `apps/platform/automation/src/scheduler.ts:42-54`

## Fix
Track an in-flight flag (or chain via `setTimeout` instead of `setInterval`) and skip overlapping runs.

## Severity
High — duplicate automations.

mamenesia:
## Summary

Automation workers now skip overlapping poll ticks, so one complete reload/curator/status cycle runs at a time.

Closes #514.

**Before:** An async `setInterval` callback could start another reload while the prior poll was stalled, causing duplicate production dispatch attempts.
**After:** New ticks return until the complete in-flight poll cycle settles.

Why safe:
1. Admission is synchronous before the first `await`, and release happens in `finally` after the full cycle.
2. A deterministic test covers blocked reload and curator phases, both failure paths, and successful recovery.

Only follow up if worker stop/restart must cancel an already-running poll; this change only serializes scheduled poll ticks.

## Test plan
- [x] `bun run --filter @nakama/automation test` (3 pass)
- [x] `bun run --filter @nakama/automation build` and `bun x ultracite check apps/platform/automation/src/scheduler.ts apps/platform/automation/src/scheduler.test.ts` (build passed; 2 files checked)
- [ ] Stall a poll beyond its interval and confirm the next tick does not issue a second schedules request


apps/platform/automation/src/scheduler.test.ts
```diff
@@ -62,4 +65,110 @@ describe("AutomationWorkerScheduler", () => {
...
     scheduler.stop();
   });
+
+  test("serializes complete poll cycles and resumes after failures", async () => {
+    const releaseFirstReload = Promise.withResolvers<void>();
+    const curatorStarted = Promise.withResolvers<void>();
+    const failingCurator =
+      Promise.withResolvers<
+        Awaited<ReturnType<NakamaClient["listSkillCuratorOrgs"]>>
+      >();
```
ahmadrosid: Reload hang + reload reject already prove admission before the first await and release in `finally` across the whole cycle. The curator-phase overlap / failure path doesn't add much on top of that — you could drop `curatorStarted`, `failingCurator`, and `curatorCalls` and keep the same confidence with a shorter test.

(Also that `Awaited<ReturnType<...>>` goes away with it.)

apps/platform/automation/src/scheduler.test.ts
```diff
@@ -1,8 +1,10 @@
-import { describe, expect, test } from "bun:test";
+import { describe, expect, spyOn, test } from "bun:test";
 import type { NakamaClient } from "@nakama/client";
 import type { AutomationSchedule } from "@nakama/core";
 import { AutomationWorkerScheduler } from "./scheduler";
 
+const POLL_INTERVAL_MS = 1000;
+
```
ahmadrosid: Only used once — fine to inline `1000` at `beginPolling`.

ahmadrosid (commented):
Nice, small fix — the in-flight flag is exactly what this needed.

Test covers the right behaviors; a couple spots look like they could drop without losing confidence.

apps/platform/automation/src/scheduler.test.ts
mamenesia: Updated in `ffa1598a`: the curator rejection/type/counter plumbing is removed, and `1000` is inlined.

One successful curator stall remains intentionally so the test still proves the guard spans the complete reload + curator cycle, rather than reload alone. As a focused mutation check, releasing `pollInFlight` immediately after reload made this reduced test fail (`listCalls` became 3 instead of 2); the real implementation passes. Reload rejection remains the single failure/recovery case, so the test is shorter without losing that boundary coverage.

ahmadrosid (approved):
Lets resolve git conflict and merge this PR

PR #717, opened 2026-08-28
fix(cli): restore stdin encoding when TerminalInput stops

### LINKED ISSUE #621: [cli] cleanupChat removes listeners but never restores terminal encoding
## Bug
`terminalInput.stop()` resets raw mode but encoding (`utf8`) is left set. Subsequent readline operations (e.g. error fallback) inherit the wrong encoding if the caller expected ASCII.

## Files
- `apps/cli/src/chat.ts:818-822`

## Fix
Reset encoding to default in `stop()`.

## Severity
Medium — terminal hygiene.

ahmadrosid:
## Summary

Leaving interactive chat restores stdin encoding so later readline/error fallbacks keep the prior encoding.

**Before:** `TerminalInput.stop()` (via `cleanupChat`) cleared raw mode/listeners but left `utf8` set.
**After:** `start()` saves `readableEncoding`; `stop()` restores it (including `null` Buffer mode).

Why safe:
1. Only touches encoding restore on the existing start/stop path
2. Preserves whatever encoding was active before chat (`ascii`, `null`, etc.)
3. Regression tests cover ascii and null restore

Only re-open if a TTY path still needs encoding after stop for a reason we missed.

## Test plan
- [x] `bun test apps/cli/src/terminal-input.test.ts apps/cli/src/terminal-screen.test.ts` (7 pass)
- [x] `bun run check -- apps/cli/src/terminal-input.ts apps/cli/src/terminal-input.test.ts`
- [ ] Exit CLI chat then trigger a readline/error fallback — encoding matches pre-chat

Closes #621

### DIFF
--- a/apps/cli/src/terminal-input.test.ts
+++ b/apps/cli/src/terminal-input.test.ts
@@ -1,0 +1,99 @@
+import { afterEach, describe, expect, mock, test } from "bun:test";
+import { TerminalInput } from "./terminal-input";
+
+type StdinStub = NodeJS.ReadStream & {
+  setRawMode: (mode: boolean) => NodeJS.ReadStream;
+};
+
+describe("TerminalInput encoding", () => {
+  const restores: Array<() => void> = [];
+
+  afterEach(() => {
+    for (const restore of restores.splice(0).reverse()) {
+      restore();
+    }
+  });
+
+  function stubStdio() {
+    const stdin = process.stdin as StdinStub;
+    const setEncoding = mock((_encoding?: BufferEncoding | null) => stdin);
+    const setRawMode = mock((_mode: boolean) => stdin);
+    const resume = mock(() => stdin);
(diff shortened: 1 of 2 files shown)


ahmadrosid:
Closing as duplicate — this fix is already on `main`.

Same change landed on main in these commits:
- `a6d25509` fix(cli): restore stdin encoding when TerminalInput stops
- `f124b9f0` fix(cli): actually restore stdin encoding after TerminalInput
- `219c3b9a` fix(cli): strip cursor probes in a single pass
- `77b33cef` merge: include cursor-probe strip with encoding restore

`main` now has `restoreReadableEncoding()`, which handles the `setEncoding(null)` case (nodejs/node#51083) by clearing `_readableState` — this PR's `setEncoding(previousEncoding)` path doesn't. Tests are covered by `apps/cli/src/terminal-encoding.test.ts`.

PR #718, opened 2026-08-28
fix(cli): bail early on oversized clipboard images

### LINKED ISSUE #610: [cli] clipboard-image.ts no early size bailout for huge clipboard data

ahmadrosid:
## Summary

CLI clipboard paste rejects huge images before Base64 encoding — no more ~67 MB string spike.

**Before:** full clipboard bytes → Base64 → then `validateImageAttachments` rejects
**After:** `bytes.length > MAX_IMAGE_BYTES` throws the same `NakamaApiError` first

Why safe:
1. Same message/status as `validateImageAttachments` size path
2. Valid images still encode + validate as before
3. Only adds an early length check

Only re-add post-encode size validation alone if a path bypasses this bailout.

## Test plan
- [x] `bun test apps/cli/src/clipboard-image.test.ts` (3 pass)
- [x] `bun test apps/cli/src/image-input.test.ts packages/core/src/message-content.test.ts` (related; all pass)
- [x] `bun x ultracite check apps/cli/src/clipboard-image.ts apps/cli/src/clipboard-image.test.ts`
- [ ] `/paste` with a normal screenshot still attaches; oversized clipboard still shows "Each image must be at most 5 MB."

Closes #610

### DIFF
--- a/apps/cli/src/clipboard-image.test.ts
+++ b/apps/cli/src/clipboard-image.test.ts
@@ -1,0 +1,47 @@
+import { describe, expect, mock, test } from "bun:test";
+import { MAX_IMAGE_BYTES, type NakamaApiError } from "@nakama/core";
+
+const tinyPng = Buffer.from(
+  "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8z8BQDwAEhQGAhKmMIQAAAABJRU5ErkJggg==",
+  "base64"
+);
+
+let hasImageResult = false;
+let imageBytes: Uint8Array | null = null;
+
+mock.module("@crosscopy/clipboard", () => ({
+  getImageBinary: async () => imageBytes,
+  hasImage: () => hasImageResult,
+}));
+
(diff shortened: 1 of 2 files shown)


ahmadrosid:
Closing as duplicate — #649 already fixed #610 on main.

`attachmentFromClipboardBytes` now does the early `bytes.length > MAX_IMAGE_BYTES` bailout before Base64 encoding, which is the exact behavior this PR adds. This branch is based on the pre-#649 version of `clipboard-image.ts`, so it would just conflict or duplicate the check.

Only difference is the error type (this PR throws `NakamaApiError` vs the merged plain `Error`), which is cosmetic. If matching the `NakamaApiError` message matters, that's a one-line follow-up on main, not a separate PR.

PR #719, opened 2026-08-28
fix(db): bump sessions.updated_at on replaceMessagesForSession

### LINKED ISSUE #606: [db] replaceMessagesForSession does not update sessions.updated_at
## Bug
`appendMessagesForSession`, `updateSessionTitle`, `updateSessionModel` all touch `sessions` (or not). `replaceMessagesForSession` deletes and re-inserts messages but leaves `sessions.updated_at` untouched, so the session-summary view (`MAX(m.created_at) AS updated_at`) will show a timestamp that ignores the most recent edit.

## Files
- `packages/db/src/adapters/sqlite.ts:2590-2602` vs. `:612-622`

## Fix
Update `sessions.updated_at = ?` in the same statement that deletes messages, or derive it from the latest `message.createdAt`.

## Severity
Medium — inconsistency.

ahmadrosid:
## Summary

Replace/compaction of session messages advances the session-list timestamp even when message `createdAt` values stay stale — Fixes #606.

**Before:** Summaries used `MAX(m.created_at)` only, so `replaceMessagesForSession` with old message times left the sidebar stamp stuck.
**After:** `sessions.updated_at` (migrated + schema); replace bumps it to at least now inside one SQLite transaction; summaries take `max(message times, sessions.updated_at)`.

Why safe:
1. Migration backfills from existing message max / `created_at`
2. SQLite + in-memory stay aligned; append does not rewrite the column
3. Regression test covers stale replace and append-after-replace

Only re-check if a caller needs append itself to bump `sessions.updated_at` independently of message timestamps.

## Test plan
- [x] `bun test packages/db/src/replace-messages-updated-at.test.ts packages/db/src/migrate.test.ts packages/db/src/session-model.test.ts` (23 pass)
- [x] `bun test packages/db/src/` (81 pass)
- [ ] Compact or regenerate a chat whose message stamps stay old — sidebar “updated” moves to now

### DIFF
--- a/packages/db/sql/schema.sql
+++ b/packages/db/sql/schema.sql
@@ -44,6 +44,7 @@ CREATE TABLE IF NOT EXISTS sessions (
   channel TEXT NOT NULL,
   user_id TEXT,
   created_at TEXT NOT NULL,
+  updated_at TEXT NOT NULL,
   title TEXT,
   model TEXT,
   agent_todos TEXT DEFAULT '[]' NOT NULL,
--- a/packages/db/src/adapters/in-memory.ts
+++ b/packages/db/src/adapters/in-memory.ts
@@ -64,6 +64,7 @@ export function createInMemoryDatabaseAdapter(): DatabaseAdapter {
   const profileSkills = new Map<string, Set<string>>();
   const skillUsage = new Map<string, StoredSkillUsageRecord>();
   const sessions = new Map<string, StoredSessionRecord>();
+  const sessionUpdatedAt = new Map<string, string>();
   const sessionMessages = new Map<string, StoredSessionMessageRecord[]>();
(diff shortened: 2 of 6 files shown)


ahmadrosid:
Closing as duplicate of #675 — same fix for #606, same title, same file set, both authored within the org.

Delta worth noting: #675 landed first (Aug 27). The only real difference is the `sqlite.ts` path — #719 bumps `sessions.updated_at` inside a single transaction (+33/-13 vs #675's +18/-3) and adds an append-after-replace regression assertion. If that transaction nuance matters, fold it into #675 rather than keeping two PRs open.

Keeping #675 as the canonical PR.

ahmadrosid:
Closed: duplicate of #675.

PR #730, opened 2026-08-28
fix(providers): reject malformed API keys on create/update

### LINKED ISSUE #628: server(api): provider creation accepts malformed API keys silently (HTTP 2xx)
## Summary

`POST /v1/providers` accepts an obviously invalid API key (`sk-junk-qa-123`) plus a nonexistent model (`gpt-junk`) and answers 2xx — a broken provider is persisted with no client- or server-side sanity check.

**Measured on:** `cf3d6ac6`

## Reproduction

```bash
curl -X POST /v1/providers \n  -d '{"type":"openai","apiKey":"sk-junk-qa-123","model":"gpt-junk"}'
# → 2xx, provider persisted
```

Same via UI: Settings → Add provider.

## Expected vs actual

Expected:

```
key-format sanity check at minimum (prefix/length per provider type);
ideally optional verify call with clear error on failure
```

Actual:

```
2xx + broken provider stored; failure only discovered later at chat time
```

Note: the Composio integration already validates keys server-side and rejects invalid ones with 400 — that pattern could be reused here.

idsoke:
## Summary
- `POST /v1/providers` and `PATCH /v1/providers/:id` accepted API keys with no format sanity check, so obviously fake keys (e.g. `sk-junk-qa-123`) were persisted and only surfaced as a failure later at chat time.
- Add `validateProviderApiKeyFormat` (`packages/core/src/user-config.ts`) checking prefix/length per provider for the providers with a well-documented key format (openai, anthropic, gemini, openrouter, deepseek, cerebras). Providers with opaque/self-hosted formats (ollama, openai_compatible, cloudflare, fireworks, opencode_go) are intentionally left unchecked to avoid false rejections.
- Wire the validator into both create (`buildProviderInstanceFromCreateRequest`) and update (`applyProviderInstanceUpdate`) paths.
- `POST /v1/providers` and `PATCH /v1/providers/:id` previously had no try/catch, so thrown errors fell through to the global handler as a 500. Wrapped them to return 400 on validation errors, matching the pattern already used elsewhere in this file (e.g. Composio key validation).

Fixes #628

## Test plan
- [x] `bun test apps/server/src/services/provider-instance-helpers.test.ts` — 14 pass
- [x] `bun test packages/core/src/user-config.test.ts` — 12 pass
- [x] `bun test apps/server/src/http/routes/settings-rbac.test.ts` — 36 pass (no auth/routing regressions)
- [x] Manually reproduced the exact issue payload (`type: openai`, `apiKey: sk-junk-qa-123`, `model: gpt-junk`) against `buildProviderInstanceFromCreateRequest` before/after: before → persisted silently; after → rejected with `That doesn't look like a valid OpenAI API key (expected at least 40 characters).`

### DIFF
--- a/apps/server/src/services/provider-instance-helpers.test.ts
+++ b/apps/server/src/services/provider-instance-helpers.test.ts
@@ -208,18 +208,6 @@ describe("resolveProfileProviderSelection", () => {
 });
 
 describe("applyProviderInstanceUpdate", () => {
-  test("rejects an obviously malformed API key on update", () => {
-    const current: ProviderInstance = createProviderInstance({
-      id: "openai-1",
-      label: "OpenAI",
-      type: "openai",
-    });
-
-    expect(() =>
-      applyProviderInstanceUpdate(current, { apiKey: "sk-junk-qa-123" })
-    ).toThrow(/valid OpenAI API key/i);
-  });
-
   test("stores wireApi only for a recognised value on a compatible instance", () => {
     const instance = createProviderInstance({
       baseUrl: "https://endpoint.test/v1",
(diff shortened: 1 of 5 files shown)


apps/server/src/services/provider-instance-helpers.test.ts
```diff
@@ -388,4 +400,40 @@ describe("buildProviderInstanceFromCreateRequest", () => {
       }
     }
   });
+
+  test("rejects an obviously malformed OpenAI API key instead of persisting it", () => {
```
ahmadrosid, line 392: `validateProviderApiKeyFormat` already covers short / wrong-prefix / accept / opaque in `user-config.test.ts`. One create reject here is enough to prove the helpers call into it — the update reject, wrong-prefix, accept, and opaque cases below don’t add much on top of that.

ahmadrosid (commented):
Validator + route wiring look good. One place the helpers tests re-cover ground the unit suite already owns.

apps/server/src/services/provider-instance-helpers.test.ts
idsoke: Good point, trimmed it down. Kept just the one create-reject test to prove the wiring works, dropped the rest since they're already covered in user-config.test.ts. Pushed in 82fe769b.

Issue #531, opened 2026-08-26
[db] replaceMessagesForSession does DELETE + N inserts without transaction


vincentri:
## Bug
Function runs `DELETE FROM session_messages WHERE session_id = ?` followed by N `INSERT`s outside any transaction. If a crash / unique-violation occurs after the DELETE but before the inserts complete, the session is wiped to zero messages.

## Files
- `packages/db/src/adapters/sqlite.ts:2590-2602`

## Fix
Wrap the delete + insert loop in `db.transaction(() => { ... })()`.

## Severity
High — data loss.

mamenesia:
I would like to work on this. I reproduced the partial-replacement failure locally and have a transaction-based fix with deterministic rollback coverage. Could you assign #531 to me? I will open a draft PR for review.

PR #735, opened 2026-08-29
fix(db): make message appends atomic

### LINKED ISSUE #532: [db] appendMessagesForSession writes N rows without a transaction
## Bug
Each `appendMessageStmt.run()` is committed independently. A failed insert in the middle leaves a half-written session with gaps in `seq`. The in-memory adapter is atomic (single Map set); SQLite is not.

## Files
- `packages/db/src/adapters/sqlite.ts:1574-1584`

## Fix
Wrap the for-loop in a single `db.transaction`.

## Severity
High — data integrity.

mamenesia:
## Summary

Failing SQLite message batches now roll back completely, preserving the session state that existed before the append.

Closes #532.

**Before:** A later insert failure left earlier messages from the same batch committed, while the caller still treated persistence as failed.
**After:** `appendMessagesForSession` executes its existing insert loop in one SQLite transaction.

Why safe:
1. The transaction changes only the SQLite append path; the API, payloads, and error propagation remain unchanged.
2. The regression first preserves an existing message, then proves a later duplicate-key failure does not retain the batch prefix.
3. Removing the transaction makes the regression fail with the extra inserted message.

Only follow up for replace semantics under #531; `replaceMessagesForSession` is intentionally unchanged.

## Test plan
- [x] `bun test packages/db/src/replace-messages-updated-at.test.ts` (4 pass on Bun 1.3.0, 1.3.10, and 1.3.14)
- [x] DB tests excluding the existing Windows-incompatible `reopen.test.ts` (82 pass, 2 platform skips)
- [x] `bun run --filter @nakama/db build` and targeted `bun x ultracite check` (pass)
- [x] GitHub Ubuntu CI on pinned Bun 1.3.14 (test, knip, react-doctor pass)

### DIFF
--- a/packages/db/src/adapters/sqlite.ts
+++ b/packages/db/src/adapters/sqlite.ts
@@ -653,6 +653,19 @@ function createSqliteDatabaseAdapter(db: Database): DatabaseAdapter {
     INSERT INTO session_messages (id, session_id, seq, payload, created_at)
     VALUES (?, ?, ?, ?, ?)
   `);
+  const appendMessagesTransaction = db.transaction(
+    (sessionId: string, messages: StoredSessionMessageRecord[]) => {
+      for (const message of messages) {
+        appendMessageStmt.run(
+          message.id,
+          sessionId,
+          message.seq,
+          JSON.stringify(message.payload),
+          message.createdAt
+        );
+      }
+    }
+  );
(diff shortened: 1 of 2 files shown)


ahmadrosid (commented):
Looks lean — ship. Transaction around the existing insert loop, plus one regression that fails without it.

PR #739, opened 2026-08-29
fix(db): make message replacements atomic

### LINKED ISSUE #531: [db] replaceMessagesForSession does DELETE + N inserts without transaction

mamenesia:
## Summary

Failed SQLite message replacements preserve the previous session history instead of leaving a partial replacement.

**Before:** `replaceMessagesForSession` deleted existing rows and inserted replacements without a transaction, so a later insert failure lost the prior set and retained partial rows.
**After:** deletion, all inserts, and `sessions.updated_at` commit or roll back within one SQLite transaction.

Why safe:
1. The transaction uses the same Bun SQLite wrapper already used for atomic message appends.
2. A deterministic duplicate-ID regression proves the previous message remains after an insert failure.

Only follow up if replacement gains another database write that must join this transaction boundary.

Closes #531

## Test plan
- [x] Exact no-transaction mutation reproduces the partial row; restored fix passes on Bun 1.3.0, 1.3.10, and 1.3.14 (5 pass each)
- [x] Impacted DB tests, DB build, and targeted Ultracite (83 pass, 2 platform skips; build and lint pass)
- [ ] GitHub CI runs the repository checks


ahmadrosid (commented):
Looks lean — ship. Shared insert + one transaction around delete/insert/updated_at is the right size for the fix.

Issue #571, opened 2026-08-26
[frontend] storeChatDraft collides on rapid clicks


vincentri:
## Bug
Draft key is `d${Date.now()}` — if the user clicks twice in the same millisecond (e.g. double-click on "Send"), two draft entries share the key and the second overwrites the first.

## Files
- `apps/web/src/lib/chat-history.ts:98-102`

## Fix
Use `createClientId()` for the draft key.

## Severity
Medium — data loss on rapid input.

fajarhide:
Re-read on `60ce1bbf`. Fixed. `chat-history.ts:101` is `const key = createClientId();`, exactly the fix this asked for. Landed in PR #743, which wrote `Fixes` once before a list of five numbers, so only the first closed. Closing.

Issue #570, opened 2026-08-26
[frontend] chat-history.ts URL cache buster always changes


vincentri:
## Bug
`buildNewChatPath` adds `_: ${Date.now()}` to bust cache on every render. Bloats `window.history` with a new entry per click and triggers `useEffect` for `useLayoutEffect` even when the user clicks "New chat" twice in a row with no semantic change.

## Files
- `apps/web/src/lib/chat-history.ts:36`

## Fix
Use a `useRef` counter inside the page that increments only when the user actually initiates a new chat; drop `Date.now()` from the URL.

## Severity
Medium — UX / history pollution.

fajarhide:
Re-read on `60ce1bbf`. Fixed. `buildNewChatPath` at `chat-history.ts:37` builds `new URLSearchParams({ new: "1" })` and `Date.now()` is gone from the path. The `useRef` counter the issue also asked for was not added and is not needed: with the parameter gone there is no per-click history entry to bloat. Landed in PR #743, which wrote `Fixes` once before a list, so only the first number closed. Closing.

Issue #519, opened 2026-08-26
[frontend] mcpArgKey uses arg value as React key — wrong DOM reuse


vincentri:
## Bug
When two args have the same value, `priorMatches` increments so the key becomes `arg:2`, `arg:3`… but if the user removes the duplicate at index 1 then re-adds the same arg at the end, the keys can collide with a still-mounted sibling. React will reuse the wrong DOM node, which in the chip UI causes the wrong "remove" handler to fire.

## Files
- `apps/web/src/components/soul-tools/mcp-tab/McpFormEditors.tsx:12-17, 133`

## Fix
Use a stable id (e.g. `createClientId()` on add) as the key.

## Severity
High — state corruption in MCP form editor.

fajarhide:
Re-read on `60ce1bbf`. Fixed. `McpFormEditors.tsx:129` is `key={argKeysRef.current[index]}`, the keys come from `createClientId()` at `:245` and `syncRowKeys`, and `mcpArgKey` and `priorMatches` have zero hits anywhere under `apps/web`. Landed in PR #743, which wrote `Fixes` once before a list of five numbers, so only the first closed. Closing.

Issue #516, opened 2026-08-26
[frontend] handleProfileSwitch reads stale busy/profileId from closure


vincentri:
## Bug
`use-chat-page.ts:564-573` exposes `handleProfileSwitch` as a callback that uses `profileId` and `busy` from the closure. Meanwhile the handler-registered version at `:575-589` correctly reads `profileIdRef.current`/`busRef.current`. Closure values can lag the actual store, allowing a profile switch mid-turn that races against the in-flight `AbortController` and message queue in `streamAbortRef`.

## Files
- `apps/web/src/pages/chat/use-chat-page.ts:564-573`
- `apps/web/src/pages/chat/use-chat-page.ts:575-589`

## Fix
Use the same `profileIdRef`/`busyRef` pattern in `handleProfileSwitch`, or derive the check from refs at call time.

## Severity
High — race condition in chat core flow.

fajarhide:
Re-read on `60ce1bbf`. Fixed. `use-chat-page.ts:634-645` now reads `profileIdRef.current` and `busyRef.current` inside `handleProfileSwitch`, so the closure copies are gone. Landed in PR #743, which wrote `Fixes` once before a list, so only the first number closed. Closing.

PR #743, opened 2026-08-29
fix(web): clear stuck chat stream state and related races

### LINKED ISSUE #518: [frontend] finalizeStreamingMessages only clears streaming on the last assistant
## Bug
When a turn is aborted mid-stream and the in-flight assistant was already finalized by an earlier tool end, the loop returns without flipping `streaming: false`. Subsequent renders show a "still streaming" UI for a message that will never update.

## Files
- `apps/web/src/lib/chat-stream.ts:417-431`

## Fix
Flip `streaming`/`thinkingStreaming` to `false` on any assistant that was being acted on, or track the streaming assistant id from the start.

## Severity
High — UX, stuck UI state.

ahmadrosid:
## Summary

Abort, profile switch, MCP args, chat drafts, and New chat URLs no longer leave stuck or colliding UI state. Fixes #518, #516, #519, #571, and #570.

**Before:** Abort left earlier assistants marked streaming. Profile switch used stale busy/profile values. MCP arg keys collided after remove-and-readd. Rapid Send reused draft keys. New chat always changed `_` in the URL.
**After:** `finalizeStreamingMessages` clears every streaming assistant. Profile switch reads `profileIdRef`/`busyRef`. MCP args use `syncRowKeys`. Drafts use `createClientId`. New chat path drops `_`.

Why safe:
1. Finalize only flips flags on assistants that still stream or think-stream
2. Profile switch guard matches the registered switch handler
3. New chat still carries `new=1` and `profile`

Only re-check if a second New chat click must re-run while the URL still has `?new=1` (no `_` cache buster). No hook unit test covers #516.

## Test plan
- [x] `bun test apps/web/src/lib/chat-stream-finalize.test.ts apps/web/src/lib/chat-history-route.test.ts apps/web/src/lib/client-id.test.ts` (16 pass)
- [x] `bun run --cwd apps/web build` (tsc + vite ok)
- [x] `bun x ultracite check` on changed files (ok)
- [ ] Abort mid-tool turn: no assistant stays on streaming UI
- [ ] Switch profile while busy is blocked; switch when idle works

### DIFF
--- a/apps/web/src/components/soul-tools/mcp-tab/McpFormEditors.tsx
+++ b/apps/web/src/components/soul-tools/mcp-tab/McpFormEditors.tsx
@@ -9,13 +9,6 @@ import { Input } from "@/components/ui/input";
 import { createClientId, syncRowKeys } from "@/lib/client-id";
 import { cn } from "@/lib/utils";
 
-function mcpArgKey(arg: string, index: number, args: string[]): string {
-  const priorMatches = args
-    .slice(0, index)
-    .filter((value) => value === arg).length;
-  return priorMatches === 0 ? arg : `${arg}:${priorMatches + 1}`;
-}
-
 export function McpFormField({
   label,
   htmlFor,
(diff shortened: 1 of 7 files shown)


ahmadrosid (commented):
Trimmed the finalize fixture setup and the redundant buildNewChatPath stability test. Rest looks lean — ship.

Issue #611, opened 2026-08-26
[security] CLI: parseIni allows key/value injection on disk-loaded INI


vincentri:
## Bug
If `~/.nakama/cli.ini` is writable by another user (rare, but possible on shared hosts), that user can inject arbitrary keys/values into the parsed dict. Subsequent saves will re-emit them, persisting injected data.

## Files
- `apps/cli/src/cli-config.ts:70-91`

## Fix
Lock down file permissions (already 0o600 via `writeTextFile`), and validate keys against a known set.

## Severity
Medium — data injection on shared hosts.

fajarhide:
Re-read on `60ce1bbf`. Fixed. `cli-config.ts:87-89` checks every parsed key against `CLI_CONFIG_KEYS` and skips anything not in the set, so an injected key is dropped at parse time and cannot be re-emitted on the next save. That is the allowlist this issue asked for. Closing.

PR #746, opened 2026-08-29
fix(cli): reject unknown keys in cli.ini parse

ahmadrosid:
## Summary

`parseIni` for `~/.nakama/cli.ini` only accepts `org_id` / `profile_id` — unknown keys are dropped on load (#611).

**Before:** any `key=value` line entered the parsed dict and could ride along on later saves.
**After:** `CLI_CONFIG_KEYS` allowlist in `parseIni`; unknown keys skipped.

Why safe:
1. Write path already only emits `org_id` / `profile_id`
2. File mode still 0o600 via `writeTextFile`
3. Known keys still load/save unchanged

Only re-add keys if a new CLI setting is intentionally stored in `cli.ini`.

## Test plan
- [x] `bun test apps/cli/src/cli-config.test.ts` (2 pass)
- [ ] Write a junk key into `cli.ini`, run a CLI command that saves config — junk key gone, profile/org intact

---

### DIFF
--- a/apps/cli/src/cli-config.test.ts
+++ b/apps/cli/src/cli-config.test.ts
@@ -1,31 +1,69 @@
 import { describe, expect, test } from "bun:test";
-import { mkdtemp, readFile } from "node:fs/promises";
+import { mkdtemp, readFile, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import {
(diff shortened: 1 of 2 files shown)


ahmadrosid (commented):
Looks lean — ship. Allowlist in parse + one load/save roundtrip test covers the fix without extra machinery.

Issue #564, opened 2026-08-26
[platform] withChatLock chain .then() not awaited for unhandled-rejection safety


vincentri:
## Bug
`const chain = previous.then(() => current)` is stored but `chain` is never awaited by the next caller; only `chatLocks.get(chatId)` is used. On racey error paths the chain's `.catch` may propagate to an unhandled rejection.

## Files
- `apps/platform/whatsapp/src/chat-handler.ts:671`
- `apps/platform/telegram/src/chat-handler.ts:913`

## Fix
Use `previous.then(…, …)` with a catch, or store `chain` and have subsequent callers `await chatLocks.get(chatId)`.

## Severity
Medium — unhandled-rejection hazard.

Issue #562, opened 2026-08-26
[platform] Typing interval / AbortController cleared too late in chat-handler error paths


vincentri:
## Bug
`signal = registerActiveStream(conversationKey)` at line 453 — if anything between 453 and 458 throws, the AbortController stays in `activeByChat` and never gets `clearActiveStream`'d. Stale abort controllers accumulate per conversation key. Same shape across telegram/discord/whatsapp.

## Files
- `apps/platform/telegram/src/chat-handler.ts:451-509`
- `apps/platform/discord/src/chat-handler.ts` (analogous)
- `apps/platform/whatsapp/src/chat-handler.ts` (analogous)

## Fix
Move `registerActiveStream` and `clearActiveStream` into a `try/finally` that wraps the whole body.

## Severity
Medium — latent leak.

xietsunzao:
I'd like to pick this up.

Telegram and WhatsApp are already fixed by #762 (`registerActiveStream` moved right before `try`, `typingLoop.start()` moved inside `try`, both cleaned up in one `finally`) — tests confirm `hasActiveStreams()` goes false on an error path for both.

Discord's `chat-handler.ts` wasn't touched by that PR and still has the gap: `registerActiveStream`, `createTypingLoop`, two status message constructors, and `typingLoop.start()` all run before `try`, so a throw in any of those leaves the AbortController registered with nothing to clear it. There's also no test covering discord's stream/typing cleanup at all right now.

Scoping this PR to discord: apply the same fix shape as telegram/whatsapp, plus a regression test.

ahmadrosid:
assigned to you @xietsunzao

PR #762, opened 2026-08-29
fix(platform): harden Telegram/WhatsApp stream safety

### LINKED ISSUE #560: [platform] Telegram download fails open on oversized file (no streaming)
## Bug
`downloadTelegramFile` checks `file.file_size` then downloads the file anyway and re-checks after `arrayBuffer()`. A malicious sender could send a 2 GB file with `file_size` lying, or omit `file_size` so the first check is skipped, exhausting worker memory before the post-download check fires.

## Files
- `apps/platform/telegram/src/attachments.ts:39-73`

## Fix
Stream the response with `response.body` and abort as soon as `bytesReceived > maxBytes`.

## Severity
Medium — DoS amplification.

ahmadrosid:
## Summary

Telegram and WhatsApp stream paths stop oversized downloads early and clean abort/typing state on every error path.

**Before:** Telegram could load a whole file into memory before the size check; a throw after `registerActiveStream` could leave a stale abort controller; chat-lock `.then` chains could surface unhandled rejections.
**After:** Telegram streams with a hard byte cap (`readResponseBodyCapped`); Telegram/WhatsApp clear stream + typing in an outer `finally`; `withChatLock` keeps the stored chain rejection-safe.

Why safe:
1. Scope is Telegram/WhatsApp only — no Discord edits
2. Cap still throws `OversizedTelegramFileError` (same user reply path)
3. Lock map still deletes the chain in `finally`

Only re-add Discord cleanup if the same leak shows up there in production.

Fixes #560, #562, #564.

## Test plan
- [x] `bun test apps/platform/telegram/src/attachments.test.ts apps/platform/telegram/src/chat-handler.test.ts apps/platform/whatsapp/src/chat-handler.test.ts` (96 pass)
- [x] `bun x ultracite check` on the 6 changed files (clean)
- [x] `bun run --filter @nakama/telegram build` (exit 0)
- [x] `bun run --filter @nakama/whatsapp build` (exit 0)
- [ ] Send an oversized Telegram document with a lying/missing `file_size` — bot rejects without OOM

### DIFF
--- a/apps/platform/telegram/src/attachments.test.ts
+++ b/apps/platform/telegram/src/attachments.test.ts
@@ -204,27 +204,4 @@ describe("downloadTelegramFile", () => {
     );
     expect(pulls).toBeLessThanOrEqual(3);
   });
-
-  test("rejects oversized body when file_size is omitted", async () => {
-    const maxBytes = 8;
-    fetchSpy = spyOn(globalThis, "fetch").mockResolvedValue(
-      new Response(new Uint8Array(maxBytes + 1).fill(66), {
-        headers: { "content-type": "application/pdf" },
-        status: 200,
-      })
-    );
-
-    const ctx = {
-      api: {
-        getFile: async () => ({
-          file_path: "documents/liar.pdf",
-        }),
-        token: "test-token",
-      },
-    } as unknown as Context;
-
-    await expect(downloadTelegramFile(ctx, "file-1", maxBytes)).rejects.toThrow(
-      "File is too large."
-    );
-  });
 });
--- a/apps/platform/telegram/src/attachments.test.ts
+++ b/apps/platform/telegram/src/attachments.test.ts
@@ -145,6 +145,12 @@ describe("buildTelegramDocumentInput", () => {
 });
 
 describe("downloadTelegramFile", () => {
+  let fetchSpy: ReturnType<typeof spyOn> | undefined;
+
+  afterEach(() => {
+    fetchSpy?.mockRestore();
+  });
+
   test("surfaces download failures to caller", async () => {
     const ctx = {
       api: {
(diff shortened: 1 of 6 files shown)


apps/platform/telegram/src/attachments.ts
```diff
@@ -72,6 +68,48 @@ export async function downloadTelegramFile(
   };
 }
 
+async function readResponseBodyCapped(
+  response: Response,
+  maxBytes: number
+): Promise<ArrayBuffer> {
+  if (!response.body) {
```
ahmadrosid: The `!response.body` branch loads the whole payload before checking size — same hazard this helper is meant to close. Fetch responses for this download always have a body; you can drop the branch and go straight to `getReader()`.

apps/platform/telegram/src/attachments.test.ts
```diff
@@ -159,4 +165,66 @@ describe("downloadTelegramFile", () => {
...
+      },
+    } as unknown as Context;
+
+    await expect(downloadTelegramFile(ctx, "file-1", maxBytes)).rejects.toThrow(
+      "File is too large."
+    );
+    expect(pulls).toBeLessThanOrEqual(3);
+  });
+
+  test("rejects oversized body when file_size is omitted", async () => {
```
ahmadrosid: This case already omits `file_size` in the streaming test above, and both assert the same oversized reject. The stream test covers the interesting part (stop mid-body); this second one can go.

apps/platform/telegram/src/chat-handler.ts
```diff
@@ -453,59 +453,61 @@ export function createChatHandler(deps: ChatHandlerDeps) {
...
-            void todoStatus.update(todos);
-          },
-          onToolEnd: () => {
-            typingLoop.ping();
-          },
-          onToolStart: () => {
-            typingLoop.ping();
+      typingLoop.start();
+
+      try {
```
ahmadrosid: Outer `try`/`finally` already runs cleanup when catch returns — the inner `try`/`catch` is just nesting. One `try`/`catch`/`finally` with `typingLoop.start()` at the top of the try is enough. Same shape in the WhatsApp handler.

ahmadrosid (commented):
Stream cap + lock rejection safety look solid. A few spots can flatten without changing behavior.

Issue #620, opened 2026-08-26
[cli] index.ts registers signal handlers that exit before cleanup runs in some paths


vincentri:
## Bug
`process.on(signal, () => { cleanup(); process.exit(0); })` — `process.exit(0)` exits immediately even if `cleanup` returns a Promise. The `try { ... } finally { stopSpawnedServer(spawnedChild); }` at `:131-133` doesn't run because `process.exit` interrupts the microtask queue.

## Files
- `apps/cli/src/index.ts:137-143`

## Fix
Await `cleanup()`; or call `stopSpawnedServer` from the signal handler.

## Severity
Medium — cleanup fragility.

fajarhide:
Fixed on main by e7a8166a (#763), and it took both halves of the Fix section.

The signal handler is now `void runCleanupThenExit(cleanup)`, which awaits `cleanup()` inside a try and only then calls `process.exit(0)` in the finally. And `cleanup` itself is `async () => { abortController.abort(); stopSpawnedServer(spawnedChild); }`, so the spawned server is stopped from the signal path rather than relying on the top-level `finally`.

`chat.test.ts` covers "awaits cleanup before exit" and "exits even when cleanup throws".

Issue #614, opened 2026-08-26
[cli] drainQueue recursion on errors may leak isStreaming


vincentri:
## Bug
If `startSend` throws inside the `try`, the `finally` runs `await drainQueue()`. If that throws (e.g. another `startSend` fails), the error is unhandled and the function exits with `isStreaming = true` leaked. The post-stream error block (lines 280-284) is skipped because of the throw.

## Files
- `apps/cli/src/chat.ts:275`

## Fix
Wrap `drainQueue` in try/catch and reset `isStreaming` on throw.

## Severity
Medium — state leak.

fajarhide:
Stale for the same reason as #547: `drainQueue` no longer exists. e7a8166a (#763) removed it earlier today and replaced it with a serialized queue plus a drain loop, so there is no recursion left to leak `isStreaming` on error. `grep -rn drainQueue apps packages` returns nothing on main.

Closing as fixed.

Issue #612, opened 2026-08-26
[cli] isStreaming flag set/read across async boundaries without serialization


vincentri:
## Bug
`isStreaming` is mutated by `startSend` and read by `drainQueue`, `handleChatMessage`, and slash command handlers. Single-threaded JS makes this safe in the same tick, but `await` boundaries (e.g. `await refreshModelsCache()` at `:155-160` or `await options.client.getModels()` at `:564`) yield control. A user message during that window could see stale `isStreaming = false` and start a parallel `startSend`.

## Files
- `apps/cli/src/chat.ts:145, 238, 254, 271, 318`

## Fix
Use a re-entrant guard and re-check after every `await`; or wrap `startSend` so all entry points funnel through the same async gate.

## Severity
Medium — message queue race.

fajarhide:
Fixed on main by e7a8166a (#763), via the second option in the Fix section.

`drainQueue` is gone. Every send now funnels through one async gate: `startSend`'s whole body is `await sendQueue.enqueue(...)`, and `sendQueue` is `createSerializedQueue`, which chains each task onto the previous one's settlement. A second `startSend` entered during an `await` no longer runs in parallel, it queues behind the first.

Issue #544, opened 2026-08-26
[cli] CLI config (cli.ini) and local-auth-token writes are not atomic


vincentri:
## Bug
`writeTextFile` uses `writeFile()` directly — no temp file, no `fsync`, no `rename`. A crash mid-write (or a Ctrl+C during `runBlockingChat`'s `await saveCliProfileId`) leaves the INI truncated or the auth-token file partially written. For the auth token this can permanently brick local auth.

## Files
- `packages/core/src/fs.ts:59-77`
- `apps/cli/src/cli-config.ts:65`
- `packages/core/src/local-auth.ts:65`

## Fix
Write to `${path}.tmp` then `fsync` and `rename(tmp, path)`.

## Severity
High — auth state corruption on crash.

fajarhide:
Fixed on main by e7a8166a (#763), exactly as described here.

`writeTextFile` now writes `${path}.tmp`, opens it and calls `handle.sync()`, then `rename(tempPath, path)`. Both call sites named in the report go through it: `cli-config.ts:67` for `cli.ini` and `local-auth.ts:66` for the auth token.

PR #763, opened 2026-08-29
fix(cli): harden input, send queue, and atomic config writes

### LINKED ISSUE #540: [cli] Bracketed paste can hang and accumulate unbounded buffer
## Bug
Once `inBracketedPaste = true`, every subsequent chunk is appended to `pasteBuffer` until `\x1b[201~` arrives. If the terminal never sends the end (malformed paste, paste interrupted by a focus loss, EOF without terminator), `pasteBuffer` grows forever and the prompt is wedged in paste mode. There's no timeout or max-length check.

## Files
- `apps/cli/src/persistent-prompt.ts:46, 264-275`

## Fix
Cap `pasteBuffer` (e.g. 256 KB), drop with a notification if exceeded; or auto-finalize on a timer if no terminator within N ms.

## Severity
High — wedge, memory leak.

ahmadrosid:
## Summary

CLI paste, ESC abort, send queue, shutdown, and config writes no longer wedge, false-abort, race, or corrupt on crash.

**Before:** Unbounded paste buffer, leaked ESC pending, bare ESC aborted streams, parallel sends / leaked `isStreaming`, signal exit skipped awaitable cleanup, and non-atomic `cli.ini` / auth-token writes.
**After:** 256 KB paste cap, malformed ESC drain, debounced ESC abort, serialized rejection-safe sends, await cleanup then exit, `writeTextFile` via temp + fsync + rename.

Why safe:
1. Input and send paths stay local to the CLI; no API contract change
2. Atomic write keeps prior `chmod: false` mode behavior
3. Focused regression tests cover each bug class

Only re-add immediate bare-ESC abort if operators need zero-latency stream cancel without a quiet window.

Fixes #540 #541 #542 #614 #612 #620 #544

## Test plan
- [x] `bun test apps/cli/src/chat.test.ts apps/cli/src/terminal-screen.test.ts apps/cli/src/persistent-prompt.test.ts apps/cli/src/cli-layout.test.ts packages/core/src/fs.test.ts apps/cli/src/cli-config.test.ts` (57 pass)
- [x] `bun x ultracite check` on changed files (pass)
- [x] `bun run --filter @nakama/cli build` and `bun run --filter @nakama/core build` (pass)
- [ ] Paste a large unfinished bracketed paste in sticky CLI — prompt stays usable after the 256 KB drop notice

### DIFF
--- a/apps/cli/src/chat.ts
+++ b/apps/cli/src/chat.ts
@@ -283,12 +283,6 @@ async function runStickyChat(
       let current: PendingMessage | undefined = message;
 
       while (current && !exiting) {
-        if (isStreaming) {
-          queue.push(current);
-          syncPendingMessages();
-          return;
-        }
-
         try {
           await runOneSend(current);
         } catch (error) {
--- a/apps/cli/src/chat.ts
+++ b/apps/cli/src/chat.ts
@@ -45,7 +45,7 @@ import { ThinkingIndicator } from "./thinking-indicator";
 const HELP_TEXT = `${formatSlashCommands()}\n\n@/path/to/image.png [message]   attach an image from file\n/paste                            attach image from clipboard (recommended)\nCtrl+V / Cmd+V (empty paste)      attach image when terminal supports it\nPageUp/PageDown                   scroll conversation history\nHome/End                          jump to oldest/newest visible history`;
 
 /** Debounce bare ESC so alt-prefix / slow paste chunks do not abort. */
-export const ESC_ABORT_DEBOUNCE_MS = 50;
+const ESC_ABORT_DEBOUNCE_MS = 50;
 
 interface RunChatOptions {
   channel: AgentChannel;
--- a/apps/cli/src/chat.ts
+++ b/apps/cli/src/chat.ts
@@ -24,7 +24,7 @@ import {
   resolveSuggestions,
 } from "./commands";
 import { mergeSendInput, parseImageLine } from "./image-input";
-import type { PendingMessage } from "./message-queue";
+import { createSerializedQueue, type PendingMessage } from "./message-queue";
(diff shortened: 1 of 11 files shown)


apps/cli/src/message-queue.ts
```diff
@@ -59,3 +59,40 @@ export function formatPendingDisplayLines(
...
+      );
+      return run;
+    },
+  };
+}
+
+/**
+ * Clear streaming state, then drain. If drain throws, force streaming off.
+ */
+export async function runRejectionSafeDrain(options: {
```
ahmadrosid: `runRejectionSafeDrain` never gets called outside its own test — `createSerializedQueue` + the `startSend` while-loop already replaced that path. Safe to delete this helper and the `describe("runRejectionSafeDrain")` block in `cli-layout.test.ts`.

apps/cli/src/chat.ts
```diff
@@ -284,6 +274,35 @@ async function runStickyChat(
...
+  async function startSend(message: PendingMessage): Promise<void> {
+    await sendQueue.enqueue(async () => {
+      if (exiting) {
+        return;
+      }
+
+      let current: PendingMessage | undefined = message;
+
+      while (current && !exiting) {
+        if (isStreaming) {
```
ahmadrosid: With `sendQueue` serializing these callbacks, `runOneSend` always clears `isStreaming` in `finally` before the next loop iteration — so this re-queue branch should never fire. Call sites already park pending work on `queue` when a stream is live. You can drop the `if (isStreaming)` block and keep the while-drain.

ahmadrosid (commented):
Solid bug batch — paste cap, ESC debounce, serialized sends, and atomic writes all earn their keep. Two leftovers from the queue rewrite you can drop.

PR #771, opened 2026-08-30
fix(server): preserve usage totals during reload

### LINKED ISSUE #552: [server] LlmUsageTracker.record / reloadFromDatabase race can lose in-flight increments
## Bug
`record()` updates memory synchronously and fires `void this.persist(...)`. If `reloadFromDatabase()` runs concurrently, it can reset memory to the pre-persist DB snapshot, leaving DB ahead of memory until the next `record()` re-syncs them. Aggregate totals displayed in dashboards may temporarily underreport during the window.

## Files
- `apps/server/src/services/llm-usage-tracker.ts:55-63, 69-101`

## Fix
Serialize via an async lock around `record` + `reloadFromDatabase`, or read-modify-write under a single DB transaction.

## Severity
Medium — race condition (correctness, not security).

mamenesia:
## Summary

Within one active database generation, usage totals survive concurrent `reloadFromDatabase()` calls.

**Before:** reload could apply a pre-write snapshot and leave memory behind SQLite.
**After:** reload waits for tracked writes and retries any aggregate/per-model snapshot spanning a newer record.

Why safe:
1. `record()` stays synchronous and fire-and-forget; only reload waits
2. Deterministic tests force both race orders, and a stable revision gates snapshot publication

Only follow up if destructive restore/reopen must preserve pre-cutover writes, or aggregate/per-model usage must commit atomically.

## Test plan
- [x] `bun test apps/server/src/services/llm-usage-tracker.test.ts apps/server/src/providers/usage-tracking.test.ts apps/server/src/services/image-generation.test.ts packages/db/src/llm-usage-stats.test.ts` (24 pass on Windows; 24 pass in Podman/Linux; file-backed SQLite experiment aligned memory/disk at 1 request / 150 tokens)
- [x] `bun run --filter @nakama/server build` + targeted `bun x ultracite check` (pass)
- [ ] Invoke `reloadFromDatabase()` while a same-generation usage write is gated; memory and SQLite stay aligned

Fixes #552

### DIFF
--- a/apps/server/src/services/llm-usage-tracker.test.ts
+++ b/apps/server/src/services/llm-usage-tracker.test.ts
@@ -62,19 +62,13 @@ describe("LlmUsageTracker", () => {
   test("waits for an in-flight record before reloading", async () => {
     const db = createInMemoryDatabaseAdapter();
     const incrementStats = db.incrementLlmUsageStats.bind(db);
-    const incrementStatsByModel = db.incrementLlmUsageStatsByModel.bind(db);
     const persistGate = Promise.withResolvers<void>();
     const persistStarted = Promise.withResolvers<void>();
-    const persistFinished = Promise.withResolvers<void>();
 
     db.incrementLlmUsageStats = async (delta, trackedSince) => {
       persistStarted.resolve();
       await persistGate.promise;
       await incrementStats(delta, trackedSince);
-    };
-    db.incrementLlmUsageStatsByModel = async (modelId, delta, trackedSince) => {
-      await incrementStatsByModel(modelId, delta, trackedSince);
-      persistFinished.resolve();
     };
 
     const tracker = await LlmUsageTracker.create(db);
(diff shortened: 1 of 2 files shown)


apps/server/src/services/llm-usage-tracker.test.ts
```diff
@@ -58,4 +58,88 @@ describe("LlmUsageTracker", () => {
...
     ]);
   });
+
+  test("waits for an in-flight record before reloading", async () => {
+    const db = createInMemoryDatabaseAdapter();
+    const incrementStats = db.incrementLlmUsageStats.bind(db);
+    const incrementStatsByModel = db.incrementLlmUsageStatsByModel.bind(db);
+    const persistGate = Promise.withResolvers<void>();
+    const persistStarted = Promise.withResolvers<void>();
+    const persistFinished = Promise.withResolvers<void>();
```
ahmadrosid: Reload already waits for the whole `persist()` promise, so `persistFinished` and the `incrementLlmUsageStatsByModel` stub don't cover a new case. Gate the aggregate write, start reload, release, `await reload`. Same proof, and you can drop the `Promise.resolve()` yield — the gated write is already in `pendingWrites` before reload starts.

ahmadrosid (commented):
Wait-for-writes plus a revision gate is the right shape. One test is carrying extra harness.

apps/server/src/services/llm-usage-tracker.test.ts
mamenesia: Agreed — reload already awaits the full persistence promise. I simplified the test to gate only the aggregate write, start reload, release the gate, and await reload.

PR #775, opened 2026-08-30
fix(integrations): validate bot tokens before saving

### LINKED ISSUE #629: web(integrations): Telegram/Discord setup accepts fake bot tokens with a success toast (no validation)
## Summary

Saving an obviously fake bot token (`123456:QA-fake-token-xyz`) in the Telegram and Discord integration sections succeeds with a success toast and persists — no upstream validation of the token.

**Measured on:** `cf3d6ac6`

## Reproduction

1. `/integrations?section=telegram`
2. Paste `123456:QA-fake-token-xyz` as bot token → save
3. Success toast; state persists after reload. Same on Discord section.

## Expected vs actual

Expected:

```
token validated against provider API (e.g. Telegram getMe) or at least format-checked;
invalid → inline error, nothing enabled
```

Actual:

```
silent false-success; bridge shows enabled but can never work
```

Reference pattern: the Composio section already rejects invalid keys server-side with 400 and stays "Not configured" — replicate that behavior here.

Security: [x] touches credential handling (bot tokens)

mamenesia:
## Summary

Save Telegram and Discord bot credentials only after provider validation → invalid tokens stay unconfigured.

**Before:** Fake or revoked bot tokens returned success and persisted.
**After:** Submitted tokens are validated before save; Discord bypasses stale cached validation.

Why safe:
1. Rejection, malformed responses, and network failures return safe 400 responses before config writes
2. Rejected replacements preserve exact config; token-less edits retain existing credentials

Only follow up if provider outages need a different response from invalid credentials.

## Test plan
- [x] Windows focused + HTTP/RBAC suites (129 pass)
- [x] Podman/Linux focused suite (19 pass); server build, Knip, and Ultracite pass
- [ ] Paste a fake token in each dashboard integration — inline error appears and status remains Not configured

Closes #629

### DIFF
--- a/apps/server/src/services/agent-service.test.ts
+++ b/apps/server/src/services/agent-service.test.ts
@@ -881,37 +881,6 @@ describe("AgentService bot token validation", () => {
     expect(await loadTelegramConfigFile()).toBeNull();
   });
 
-  for (const failure of [
-    {
-      fetch: async () => new Response("not-json", { status: 200 }),
-      name: "a malformed response",
-    },
-    {
-      fetch: async () => {
-        throw new Error("network failure");
-      },
-      name: "a network failure",
-    },
-  ]) {
-    test(`does not persist a Telegram token after ${failure.name}`, async () => {
-      globalThis.fetch = failure.fetch as typeof fetch;
-      const botToken = "123456:unvalidated-token";
-      const service = new AgentService(
-        null,
-        null,
-        createInMemoryDatabaseAdapter()
-      );
-
-      const error = await captureError(() =>
-        service.setTelegramSettings({ botToken })
-      );
-
-      expect(error).toBeInstanceOf(Error);
-      expect(error?.message).not.toContain(botToken);
-      expect(await loadTelegramConfigFile()).toBeNull();
-    });
-  }
-
   test("persists a Telegram token accepted by Telegram", async () => {
     let requestUrl = "";
     globalThis.fetch = (async (input) => {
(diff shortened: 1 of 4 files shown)


apps/server/src/services/agent-service.test.ts
```diff
@@ -833,6 +839,319 @@ describe("AgentService skill_manage injection", () => {
...
+    expect({ configured, rejected: error !== null, requestCount }).toEqual({
+      configured: false,
+      rejected: true,
+      requestCount: 1,
+    });
+    expect(error?.message).not.toContain(botToken);
+    expect(await loadTelegramConfigFile()).toBeNull();
+  });
+
+  for (const failure of [
```
ahmadrosid: Telegram 401 already fails closed through the same outer catch as malformed/network. You could drop this loop and keep the same confidence.

apps/server/src/services/agent-service.test.ts
```diff
@@ -833,6 +839,319 @@ describe("AgentService skill_manage injection", () => {
...
+    expect({ configured, rejected: error !== null, requestCount }).toEqual({
+      configured: false,
+      rejected: true,
+      requestCount: 1,
+    });
+    expect(error?.message).not.toContain(botToken);
+    expect(await loadDiscordConfigFile()).toBeNull();
+  });
+
+  for (const failure of [
```
ahmadrosid: Same story on Discord — 401 already returns null from `resolveDiscordApplicationId`, so malformed/network don’t add a distinct path. Fine to drop this loop too.

ahmadrosid (commented):
Core change looks good — validate before write, and Discord `forceRefresh` is the right hook. One place the tests can shrink without losing coverage.

apps/server/src/services/agent-service.test.ts
mamenesia: Agreed — the 401 case already exercises the outer catch and proves the token is rejected without being persisted or exposed. I removed the malformed-response/network-failure loop in 780bc0fd.

apps/server/src/services/agent-service.test.ts
mamenesia: Agreed on removing these from the service suite — the service only needs to verify that a null helper result prevents persistence. I removed the loop in 780bc0fd and kept one compact helper-level test for invalid IDs and request failures, since those are distinct branches of resolveDiscordApplicationId whose null-return behavior was not otherwise covered.

PR #776, opened 2026-08-30
fix(agent): keep tool results when distillToolResult throws

### LINKED ISSUE #595: [agent/core] executeToolCall swallows the error class when distillToolResult throws
## Bug
Both `tool.run` and `distillToolResult` failures fall through the same `catch` and become `{ error: message }`. If `distillToolResult` throws (e.g. OMNI DB unreadable), the user sees the same generic shape as a tool failure, but the model never sees the original error.

## Files
- `packages/agent/src/tool-loop.ts:28-50`

## Fix
Distinguish tool-runtime errors from optimiser errors; on optimiser error, return the result untouched and log a warning.

## Severity
Medium — observability.

akmalsyrf:
## Summary

Successful tool runs keep their result even when OMNI distillation throws → the model no longer sees a fake tool `{ error }`.

**Before:** `tool.run` and `distillToolResult` shared one `catch`, so optimiser failures looked like tool failures.
**After:** nested catch returns the raw result and `console.warn`s (`executeToolCall` in `packages/agent/src/tool-loop.ts`).

Why safe:
1. Distillation is already fail-open inside OMNI; this only covers unexpected throws
2. Tool-runtime failures still return `{ error: message }` unchanged
3. Covered by a spy-based unit test that forces `distillToolResult` to throw

Only re-add collapsing into `{ error }` if operators need distillation outages to abort the turn.

## Test plan
- [x] `bun test packages/agent/src/tool-loop.test.ts` (4 pass)
- [x] Forced distill throw via spy — `executeToolCall` returns raw payload, warns, does not `{ error }`

Closes #595

### DIFF
--- a/packages/agent/src/tool-loop.test.ts
+++ b/packages/agent/src/tool-loop.test.ts
@@ -1,5 +1,6 @@
-import { describe, expect, test } from "bun:test";
+import { describe, expect, spyOn, test } from "bun:test";
 import type { ToolDefinition } from "@nakama/core";
+import * as core from "@nakama/core";
 import { canRunToolCallsInParallel, executeToolCall } from "./tool-loop";
 
 const sampleTool: ToolDefinition = {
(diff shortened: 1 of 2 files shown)


ahmadrosid (commented):
Looks lean — ship. Nested catch around distill is the right size for fail-open, and the spy test covers the throw path without extra scaffolding.

PR #777, opened 2026-08-30
fix(web): stop pushing webPublicUrl on Setup Wizard mount

### LINKED ISSUE #578: [frontend] setupWebPublicUrl sent on every Setup Wizard mount
## Bug
`useEffect(() => client.updateWebPublicUrl(window.location.origin))` fires on mount with empty deps. In dev/Strict Mode it fires twice; if the user navigates between steps without remounting the wizard it's fine, but if `SetupWizardPage` ever re-mounts while the user is on the Org Setup step, it re-pushes the URL — no error surfaced, but it silently mutates server state from a render-side effect.

## Files
- `apps/web/src/components/setup-wizard/SetupWizard.tsx:27-35`

## Fix
Move this into the Setup submission flow in `auth-context.tsx:setup` (which already handles `webPublicUrl`).

## Severity
Medium — side-effect footgun.

akmalsyrf:
## Summary

Setup Wizard remounts no longer PUT `webPublicUrl` from a mount `useEffect` → public URL is set only during auth setup.

**Before:** every `SetupWizard` mount called `client.updateWebPublicUrl(origin)` (Strict Mode = twice).
**After:** effect removed; `auth-context` `setup()` still sends `webPublicUrl` on `POST /v1/auth/setup`.

Why safe:
1. Fresh install path already persists URL in `setup()` from `window.location.origin`
2. Settings / system UI remains the place to change URL later
3. No other call site depended on the mount side effect

Only re-add a one-shot bootstrap if installs without going through `setup()` need an origin pin.

## Test plan
- [x] Diff review: mount effect + unused `client` import gone from `SetupWizard.tsx`
- [x] Code path check: `SetupStepOrganization` → `setup()` → `auth-context` injects `webPublicUrl` from `window.location.origin`

Closes #578

### DIFF
--- a/apps/web/src/components/setup-wizard/SetupWizard.tsx
+++ b/apps/web/src/components/setup-wizard/SetupWizard.tsx
@@ -11,7 +11,6 @@ import type {
   SetupWizardProps,
 } from "@/components/setup-wizard/setup-wizard.shared";
 import { useAppContext } from "@/context/use-app-context";
-import { client } from "@/lib/client";
 import { pathForPage } from "@/lib/navigation";
 
 export function SetupWizard({ onComplete }: SetupWizardProps) {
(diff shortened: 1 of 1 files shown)


ahmadrosid (commented):
Looks lean — ship. The mount PUT is gone; setup() still pins the origin. Nothing extra to drop.

PR #783, opened 2026-08-30
fix(server): force kill trapped bash processes

### LINKED ISSUE #526: [agent/core] kill via signal doesn't escalate to SIGKILL for trapped children
## Bug
`child.kill("SIGTERM")` is sent on abort, but the spawned bash is a login shell that may have already started an agent binary (claude codex, etc.) which traps SIGTERM and keeps writing. The python tool has a SIGKILL grace window but the bash tool does not — if `claude` traps SIGTERM, the turn never completes and `signal?.throwIfAborted()` is never observed.

## Files
- `apps/server/src/tools/bash.ts:151`
- `apps/server/src/services/python-tool-loader.ts:140-144`

## Fix
Mirror the python pattern: SIGTERM →5 s grace → SIGKILL.

## Severity
High — stuck turns, leaks resources.

mamenesia:
## Summary

Cancel a Bash tool run → its direct process exits even when it traps SIGTERM.

**Before:** Abort rejected the turn, but a SIGTERM-trapping shell or agent process could survive indefinitely.
**After:** `runShellCommand` sends SIGTERM, waits five seconds, then escalates to SIGKILL while preserving prompt `AbortError` rejection.

Why safe:
1. The retained child handle owns escalation; `close` clears the kill timer and abort listener
2. The real-shell cancellation regression covers the shared escalation path without a duplicate timeout case

Only follow up with process-group termination if independently forked descendants must be killed too.

Closes #526

## Test plan
- [x] `bun test apps/server/src/tools/bash.test.ts` in Podman (12 pass)
- [x] Server build, Ultracite, and Knip in Podman
- [ ] Cancel a real coding-agent CLI run that traps SIGTERM — the direct process exits after the grace period

### DIFF
--- a/apps/server/src/tools/bash.test.ts
+++ b/apps/server/src/tools/bash.test.ts
@@ -15,6 +15,30 @@ import path from "node:path";
 import { PathGuardError } from "@nakama/core";
 import { runBash } from "./bash";
 
+async function waitForPositivePid(pidPath: string): Promise<number> {
+  for (let attempt = 0; attempt < 100; attempt += 1) {
+    try {
+      const pid = Number((await readFile(pidPath, "utf8")).trim());
+      if (Number.isInteger(pid) && pid > 0) {
+        return pid;
+      }
+    } catch {
+      // The shell has not created the PID file yet.
+    }
+    await Bun.sleep(10);
+  }
+  return 0;
+}
+
+function isProcessAlive(pid: number): boolean {
+  try {
+    process.kill(pid, 0);
+    return true;
+  } catch {
+    return false;
+  }
+}
+
 describe("bash tool", () => {
   let workspaceRoot = "";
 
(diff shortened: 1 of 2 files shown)


apps/server/src/tools/bash.test.ts
```diff
@@ -44,10 +68,79 @@ describe("bash tool", () => {
...
+      await Bun.sleep(25);
+    }
+
+    if (survivedAbort) {
+      process.kill(childPid, "SIGKILL");
+    }
+    expect(survivedAbort).toBe(false);
+  }, 10_000);
+
+  test("force kills a shell that traps SIGTERM after timeout", async () => {
```
ahmadrosid: Cancel and timeout both just call the same `scheduleForcedKill` after SIGTERM — this second trap case doesn’t buy much past the cancellation one above. Keeping the cancel regression (matches the bug) is enough; you can drop this test.

ahmadrosid (commented):
Core SIGTERM→SIGKILL escalation looks tight. One overlapping regression test you could drop.

apps/server/src/tools/bash.test.ts
mamenesia: Removed the overlapping timeout trap case in f8a5337d. The cancellation regression remains as the single owner of SIGTERM-to-SIGKILL escalation coverage; the focused Podman suite passes 12/12.

Issue #439, opened 2026-08-22
fix(server): agent-browser install timeout can write to a closed SSE stream


isnaenihidayat:
## Summary

The agent-browser install endpoint has two independent 120-second timers. `streamInstallEvents` closes the SSE controller at 120 seconds, while `runTimedInstallCommand` also resolves its timeout at 120 seconds and the executor continues unwinding.

Because the outer timer is registered first, a slow install can close the stream before the installer emits its timeout/error or finishes child-process cleanup.

## Evidence

- `apps/server/src/http/coding-harness-install-stream.ts`
  - owns an outer `INSTALL_STREAM_TIMEOUT_MS = 120_000` timeout
  - the timeout sends an error and closes the controller
  - the executor is not aborted
  - `finally` closes the controller again
- `apps/server/src/services/cli-package-install.ts`
  - `runTimedInstallCommand` also uses a 120-second timeout
  - timeout resolution happens immediately after `SIGTERM`, before the child emits `close`
  - later `close` or `error` handlers can still flush buffered output through `onProgress`
- `apps/server/src/http/routes/models.ts`
  - progress and completion callbacks write directly to the SSE `send` callback

A possible timeline is:

1. The outer stream timer emits a timeout and closes the controller.
2. The inner command timer resolves with `timedOut` and sends `SIGTERM`.
3. The installer throws or later emits progress while unwinding.
4. A callback tries to enqueue into the already closed controller.
5. The stream `finally` attempts to close the controller again.

This can surface controller-state errors or an unhandled rejection while the underlying install is still settling.

## Impact

- The client receives a timeout while installation work continues briefly.
- Late progress or error events can target a closed stream.
- There is no test covering a stream timeout followed by late executor completion.

## Proposed direction

- Use a single timeout owner and propagate an `AbortSignal` from the stream to the installer/child process.
- Make terminal send/close behavior idempotent and ignore late events after the stream reaches a terminal state.
- Wait for child close/kill escalation before resolving, or detach progress callbacks after timeout.

## Acceptance criteria

- A slow executor cannot enqueue or close after the terminal event.
- Timeout cancels or fully settles the underlying install.
- The stream emits exactly one terminal error event and closes once.
- Late stdout, stderr, error, or close events do not invoke progress callbacks after timeout.
- Tests use an injectable short timeout and cover an executor that finishes after the deadline.

vianbas:
I'd like to work on this. No open PR references it.

[...]

PR #786, opened 2026-08-30
fix(server): settle a timed-out install once the child has actually exited

vianbas:
## Summary

A timed-out install reports back once the installer is actually gone, instead of once the signal was sent.

**Before:** `runTimedInstallCommand` resolved inside its own timeout callback, so an installer that ignores SIGTERM was reported as finished while it was still running. Measured at 302 ms against a 300 ms deadline, and still alive 500 ms later.
**After:** the deadline signals, then waits for the child's `exit` before resolving, bounded so no caller can wait forever (`apps/server/src/services/cli-package-install.ts`).

Why safe:
1. The path that finishes in time is untouched. It still resolves on `close`, and paired runs on both trees gave the same exit code, stdout, stderr, and latency in the same single-digit millisecond band.
2. `exit`, not `close`: `close` waits for the stdio pipes, so anything the installer leaves running holds it open. Measured on `2d4f2c07`, a child that exited at 220 ms had no `close` for 6 s, and in two other cases none arrived at all. A process already gone at the deadline settles straight away rather than sitting out the bound.
3. `bun run test` 2761 pass, 0 fail across all 11 workspaces (2758 before, the 3 new tests are the difference); `bun run check` and `bun run knip` clean; the `tsc --noEmit` error set is unchanged against `2d4f2c07`.

Only follow up if installers that no signal can reach turn out to be common. That case leaves through the settle bound with `timedOut` true, which is today's outcome, five seconds later.

Refs #439

## Test plan
- [x] `bun test apps/server/src/services/cli-package-install.test.ts` (6 pass; 10 consecutive runs, no flake)
- [x] `bun run test` (2761 pass, 0 fail)
- [x] The two tests for the new wait fail on `2d4f2c07`, settling at 55 ms on the deadline; the third guards a case `2d4f2c07` already got right
- [ ] Start an agent-browser install, kill the installer's process mid-run, and the SSE stream still ends with one error event and closes once

### DIFF
--- a/apps/server/src/services/cli-package-install.test.ts
+++ b/apps/server/src/services/cli-package-install.test.ts
@@ -17,6 +17,28 @@ const SELF_EXITING_PLAN = {
   displayCommand: "sh -c 'printf partial; exec sleep 1'",
 };
 
+/**
+ * Ignores SIGTERM, so it only dies once the escalation lands. Exits on its own
+ * after a second, which bounds every test that uses it and lets those tests put
+ * an upper bound on the wait: passing late enough to be the self-exit is a
+ * failure, not a pass.
+ */
+const SIGTERM_IGNORING_PLAN = {
+  args: ["-c", "trap '' TERM; printf partial; sleep 1"],
+  command: "sh",
+  displayCommand: "sh -c \"trap '' TERM; printf partial; sleep 1\"",
+};
+
+/**
+ * Exits at once but leaves something holding the stdout pipe, so `close` stays
+ * outstanding long after the process itself is gone.
+ */
+const PIPE_HOLDING_PLAN = {
+  args: ["-c", "sh -c 'sleep 3' & printf 'hi\\n'; exit 0"],
+  command: "sh",
+  displayCommand: "sh -c \"sh -c 'sleep 3' & printf 'hi\\n'; exit 0\"",
+};
+
 describe("runTimedInstallCommand", () => {
   test("gives up on an installer that outlives the timeout", async () => {
     const result = await runTimedInstallCommand(STALLING_PLAN, undefined, {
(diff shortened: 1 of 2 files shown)


vianbas:
One more thing, from working on the manual box in the test plan. It does not change
this PR and it is not a request to change it. It changes what you see if you
test the timeout through the UI.

**There are two 120 s deadlines on this path, and the outer one gets there first.**
`INSTALL_STREAM_TIMEOUT_MS` in `apps/server/src/http/coding-harness-install-stream.ts`
and `CLI_INSTALL_TIMEOUT_MS` in `apps/server/src/services/cli-package-install.ts`
are both `120_000`. The stream arms its timer in `start(controller)`, before the
executor spawns anything, so it reaches the deadline first. Measured on
`2d4f2c07` and on this branch, nothing injected, two runs each:

| Tree | Error event reaches the client | Executor settles | Installer alive when the error arrives |
|---|---|---|---|
| `2d4f2c07` | 120022 / 120006 ms | 120070 / 120107 ms | yes |
| this branch | 120007 / 120002 ms | 122089 / 122049 ms | yes |

**The stream deadline stops listening, it does not stop the install.** That is
why the settle column lands after the error column on both trees.

So over HTTP at 120 s the client is told the install timed out while the
installer is still running, before and after this PR. What this PR changes is
what `runTimedInstallCommand` reports to its caller. The part of that criterion
the user sees is decided by the wrapper above it.

Part of that box is measured now, and the box stays unchecked because the rest
is not. Measured, through the real stream wrapper and the real install function:
one error event, the stream ends once, zero unhandled rejections and zero
uncaught exceptions, across sixteen runs covering four deadline settings on both
trees. The detector behind those two counters was checked against a forced
rejection and a forced throw, and it does count them. Not measured: the setup
the box describes, since the installer here is `sh` standing in for a real
agent-browser install, and it dies from the timeout escalation rather than from
a kill by hand.

Repro, under five seconds. Save at the repo root and `bun run deadline-race.ts`:

```ts
import { streamInstallEvents } from "./apps/server/src/http/coding-harness-install-stream";
import { runTimedInstallCommand } from "./apps/server/src/services/cli-package-install";

const PID_FILE = "/tmp/nakama-install-probe.pid";
const SLEEP_PID_FILE = "/tmp/nakama-install-probe-sleep.pid";
const t0 = performance.now();
const ms = () => Math.round(performance.now() - t0);
const alive = (pid: number) => {
  try {
    process.kill(pid, 0);
    return true;
  } catch {
    return false;
  }
};

let executorSettledAtMs: number | null = null;
const response = streamInstallEvents<{ type: string }>(
  async (send) => {
    await runTimedInstallCommand(
      {
        args: [
          "-c",
          `echo $$ > ${PID_FILE}; trap '' TERM; sleep 600 & echo $! > ${SLEEP_PID_FILE}; wait`,
        ],
        command: "sh",
        displayCommand: "sh",
      },
      (message) => send({ message, type: "progress" } as never),
      { timeoutMs: 300 } // remove to use the real 120s
    );
    executorSettledAtMs = ms();
  },
  { timeoutMs: 300 } // remove to use the real 120s
);

const reader = response.body?.getReader();
if (!reader) {
  throw new Error("no body");
}
const decoder = new TextDecoder();
let errorEventAtMs: number | null = null;
let installerAliveAtErrorEvent: boolean | null = null;
let errorEvents = 0;

for (;;) {
  const { done, value } = await reader.read();
  if (done) {
    break;
  }
  const text = decoder.decode(value);
  errorEvents += (text.match(/"type":"error"/g) ?? []).length;
  if (errorEventAtMs === null && text.includes('"type":"error"')) {
    errorEventAtMs = ms();
    installerAliveAtErrorEvent = alive(Number(await Bun.file(PID_FILE).text()));
  }
}
const streamClosedAtMs = ms();

await new Promise((r) => setTimeout(r, 4000));

// The installer's own sleep outlives a SIGKILL aimed at the installer, so the
// repro cleans up after itself instead of leaving it behind for ten minutes.
try {
  process.kill(Number(await Bun.file(SLEEP_PID_FILE).text()), "SIGKILL");
} catch {
  // already gone
}

console.log({
  errorEventAtMs,
  errorEvents,
  executorSettledAtMs,
  installerAliveAtErrorEvent,
  streamClosedAtMs,
});
process.exit(0);
```

With both deadlines at 300 ms, two runs each: on `2d4f2c07` the error event
leaves at 303 / 304 ms and the executor settles at 335 / 327 ms, on this branch
the error event leaves at 304 / 303 ms and the executor settles at 2331 /
2332 ms. The installer was alive at the error event in all four. Deleting the
two `{ timeoutMs: 300 }` lines gives the table above. I ran that version once
per tree as well: the error event left at 120003 ms with the executor settling
at 120041 ms on `2d4f2c07`, and at 120004 ms with the executor settling at
122044 ms here.

Two ways to close that gap, both outside this PR and both your call:

1. Give the stream deadline more room than the install deadline plus the settle
   bound, so the install's own timeout is what the client hears.
2. Let the stream deadline cancel the install rather than only stop listening.

I have not built either, and I am not guessing which one you want.

Edited to narrow the claim about the manual test-plan box. The assertion in it
is measured, the setup it describes is not, so the box stays unchecked.

ahmadrosid (commented):
Looks lean — ship.

PR #787, opened 2026-08-30
fix(agent): preserve custom tool cancellation context

### LINKED ISSUE #594: [agent/core] withToolRetries swallows AbortError context
## Bug
`delay(ms, undefined, { signal })` honors the abort and rejects with `AbortError`. The retry loop catches errors and re-throws only if `signal.aborted`, but `AbortError` is then wrapped by `executeToolCall`'s try/catch into `{ error: "..." }`. The end user loses the abort context but gets a clean failure.

## Files
- `apps/server/src/services/custom-tool-handlers.ts:80-99`

## Fix
Verify the resulting error is actually the abort reason; otherwise surface the abort to the stream end signal.

## Severity
Medium — UX.

mamenesia:
## Summary

Cancel a failing custom-tool retry → the original abort reason reaches the stream instead of becoming a normal tool error.

**Before:** cancellation during a failed run or retry backoff was wrapped or serialized as `{ error }`.
**After:** `withToolRetries` and `executeToolCall` preserve and rethrow the exact `AbortSignal.reason`.

Why safe:
1. Native `throwIfAborted()` covers run failure, backoff rejection, and the agent serialization boundary.
2. Ordinary failures still retry twice and retain their existing final error behavior; focused tests own each boundary.

Only revert if an un-aborted custom-tool failure no longer completes the existing retry policy.

Closes #594

## Test plan
- [x] `bun test packages/agent/src/tool-loop.test.ts apps/server/src/services/custom-tool-handlers.test.ts` (14 pass)
- [x] `bun run --filter @nakama/agent test`; agent/server builds; Ultracite; Knip (85 pass, all checks green)
- [ ] Cancel a failing custom tool during retry backoff — cancellation reaches the stream and no second attempt starts


ahmadrosid (commented):
Looks lean — ship. Native `throwIfAborted` replaces the hand-rolled reason wrapper cleanly, and the three abort points each earn their check.

PR #788, opened 2026-08-31
feat(automation): configure workspace worker polling

### LINKED ISSUE #316: Make skill curator stale/archive clocks configurable in Organization settings
Stale (30d) and archive (90d) clocks are hardcoded in `packages/core/src/skills/freshness.ts`. Org admins should set them on **System → Organization → Skill curator**. Defaults stay 30 / 90.

- Persist `skills_curator_stale_after_days` and `skills_curator_archive_after_days` on the org
- Curator runs and the Skills stale badge use the org values
- Validate: positive integers, stale < archive, archive ≤ 3650

Depends on #274. Do not rebuild curator; thread org days into `classifySkillFreshness` and the Unused badge (`profile-skills-section.tsx` still imports `SKILL_STALE_AFTER_MS`). Unused still means last **match** or **patch** (`lastUsedAt`). Catalog views (`viewedAt`) do not reset the clock — do not change that here. Update the Organization card copy and `self-improving-skills.mdx` so 30/90 read as defaults, not fixed product numbers.

## Todos

- [ ] Clocks: stale/archive days on **System → Organization → Skill curator** (defaults 30 / 90)
- [ ] Poll interval: let an admin set the automation-worker poll from that same card (default 5 minutes). Today this is hardcoded as `AUTOMATION_POLL_INTERVAL_MS` in `apps/platform/automation/src/config.ts`. Do not add an env var. The poll is worker-global (it also reloads automation schedules), so one setting — not per-profile.

Out of scope: per-profile overrides, 7-day tick interval, restore UI (disk restore is documented in #274).

Follow-up to #274.

puspoaditya:
## Summary

Org admins can configure skill-curator freshness, while platform admins can configure the workspace-wide automation worker polling cadence without restarting the worker.

**Before:** worker polling was fixed locally and the curator/UI used hard-coded 30/90-day freshness clocks.
**After:** persisted workspace settings drive a five-minute-default worker cadence; organization settings drive stale/archive clocks.

Why safe:
1. Polling settings are persisted in SQLite with a migration and are restricted to platform admins because they apply to every organization.
2. The worker refreshes the setting on every tick, keeps the active cadence if the read fails, and reschedules only when it changes.
3. Archive remains a move, not deletion; invalid stale/archive ordering is rejected server-side.

Only re-add a process-local interval if an installation needs a deliberate offline-worker mode.

## Test plan
- [x] `bun run check`
- [x] `NODE_OPTIONS=--max-old-space-size=2048 bun run build`
- [x] `bun test packages/db/src/workspace-settings.test.ts packages/db/src/migrate.test.ts packages/core/src/skills/archive.test.ts apps/server/src/services/org-service.test.ts apps/server/src/http/automation-worker-settings.test.ts apps/server/src/http/routes/settings-rbac.test.ts`
- [x] `bun test apps/platform/automation/src/scheduler.test.ts` (3 pass)
- [ ] As a platform admin, set the poll interval to 10 minutes and confirm the worker applies it after its next tick.

Closes #316


packages/core/src/skills/freshness.ts
```diff
@@ -12,21 +12,35 @@ function toTimestamp(value: string | Date): number {
...
   lastUsedAt?: string | Date | null;
   now?: Date;
+  staleAfterDays?: number;
 }): SkillFreshness {
   const now = input.now?.getTime() ?? Date.now();
   const clock = input.lastUsedAt
     ? toTimestamp(input.lastUsedAt)
     : toTimestamp(input.createdAt);
   const unusedForMs = now - clock;
+  const staleAfterMs =
```
ahmadrosid, line 27: This divides the ms constant into days, then multiplies back — same as `input.staleAfterDays != null ? input.staleAfterDays * DAY_MS : SKILL_STALE_AFTER_MS` (and the same for archive). Easier to read and same result.

apps/web/src/components/settings/SkillsCuratorOrgCard.tsx
```diff
@@ -87,9 +113,15 @@ export function SkillsCuratorOrgCard() {
   return (
     <Card className="w-full overflow-hidden shadow-none">
       <div className="border-border border-b px-4 py-3">
-        <div className="flex items-center justify-between gap-4">
-          <p className="font-medium text-foreground text-sm">Skill curator</p>
-          <div className="flex shrink-0 items-center gap-2">
+        <div className="flex items-start justify-between gap-4">
+          <div className="min-w-0 space-y-0.5">
+            <p className="font-medium text-foreground text-sm">Skill curator</p>
+            <p className="max-w-prose text-muted-foreground text-xs leading-relaxed">
```
ahmadrosid: These helper lines (and the Freshness clocks blurb below) don’t buy much once the field labels are there — “Stale after” / “Archive after” already carry it. Fine to drop both and keep the heading + inputs.

ahmadrosid (commented):
Core change looks solid — a couple of spots that could stay shorter without losing behavior.

puspoaditya:
Addressed in 12a13d0e.

- Simplified the freshness override/default calculations without changing thresholds.
- Removed the redundant curator/archive helper copy while keeping the existing inputs and labels.

Focused tests and formatting checks pass locally.

Issue #768, opened 2026-08-30
Compaction does not fire on a reasoning-heavy chat-completions history, because the estimate skips message.thinking


vianbas:
## Summary

`estimateMessageTokens` still leaves `message.thinking` out of the count on the branch that #765 did not touch, so a turn built by `buildChatCompletionResult` is scored at a fraction of what is replayed.

`packages/agent/src/history-compaction.ts:99-105`

#765 fixed the `providerContent` branch. This is the `else` beside it.

## Measured on

`3b6fd6e8`

## Reproduction

Save as `apps/server/src/providers/compaction-thinking-estimate.test.ts` and run `bun test apps/server/src/providers/compaction-thinking-estimate.test.ts`. Both tests fail on `3b6fd6e8`.

The assistant message is not hand-written. It comes out of this repo's own streaming parser, and the payload it is compared against comes out of `toOpenAIMessages`.

```ts
import { describe, expect, mock, test } from "bun:test";
import type { ChatMessage, ProviderClient } from "@nakama/core";
// estimateHistoryTokens is not exported from @nakama/agent, hence the deep path.
import {
  compactHistory,
  estimateHistoryTokens,
} from "../../../../packages/agent/src/history-compaction";
import { createOpenAIProvider, toOpenAIMessages } from "./openai";

const j = (value: unknown) => JSON.stringify(value);

// Mirrors TOKEN_ESTIMATE_RATIO = 4 in history-compaction.ts.
const estimateTokens = (text: string) => Math.ceil(text.length / 4);

const REASONING = "The user wants last quarter invoices. ".repeat(20);
const USER: ChatMessage = { content: "invoices last quarter", role: "user" };

function eventStream(chunks: string[]): Response {
  const encoder = new TextEncoder();

  return new Response(
    new ReadableStream<Uint8Array>({
      start(controller) {
        for (const chunk of chunks) {
          controller.enqueue(encoder.encode(chunk));
        }

        controller.close();
      },
    }),
    { headers: { "Content-Type": "text/event-stream" }, status: 200 }
  );
}

const chunk = (delta: Record<string, unknown>) =>
  `data: ${j({ choices: [{ delta }] })}\n\n`;

// deepseek, xai, minimax, zhipu, and opencode_go all reach createOpenAIProvider
// with a providerName other than "openai", so every request stays on
// /chat/completions and the turn is built by buildChatCompletionResult.
async function chatCompletionsTurn(): Promise<ChatMessage> {
[...]

PR #789, opened 2026-08-31
fix(agent): stop counting a reasoning trace Gemini never replays

vianbas:
## Summary

A Gemini chat keeps the context the model was actually given, instead of one inflated by a reasoning trace Gemini never receives.

**Before:** #774 made `estimateMessageTokens` count `message.thinking` whenever `providerContent` is absent. That is right for the chat-completions mapper and wrong for Gemini: one turn holding 419 characters of reasoning was scored at 116 tokens against the 11 tokens Gemini is sent, and eight such turns read 945 against a usable 900, so the history is summarized on a trace the model never saw.
**After:** the count is gated on the provider, so the same turn scores 11 (`packages/agent/src/history-compaction.ts`). The chat-completions path is untouched and still scores 116.

Why safe:
1. `toGeminiAssistantParts` sends `content` and `toolCalls` only, and Gemini never sets `providerContent`, so its trace reaches nothing on the wire. Replaying a thought needs the `thoughtSignature` the parser drops.
2. Gemini is the only provider that needs the gate. The other 15 names in `ProviderName` reach `toOpenAIMessages`, which replays the trace as `reasoning_content`, or `toAnthropicMessages` and `toResponsesInput`, which carry it inside `providerContent` and are counted verbatim a branch earlier.
3. `bun run test` 2762 pass, 0 fail across all 11 workspaces, against 2758 on `2d4f2c07`; the 4 new tests are the difference. `bun run check` and `bun run knip` clean, and the `tsc --noEmit` error set is unchanged.

Only re-add the trace for Gemini if a future `thoughtSignature` round trip makes `toGeminiAssistantParts` send it back.

## Test plan
- [x] `bun test packages/agent/src/chat-context-usage.test.ts apps/server/src/providers/compaction-thinking-estimate.test.ts` (9 pass)
- [x] `bun run test` (2762 pass, 0 fail)
- [x] Seven mutations, two rounds with the same result: these tests catch all seven, the tests on `main` catch one
- [ ] Run a reasoning-heavy Gemini chat past the point compaction used to fire and watch the context number stay put

Refs #768

### DIFF
--- a/packages/agent/src/history-compaction.ts
+++ b/packages/agent/src/history-compaction.ts
@@ -82,12 +82,8 @@ function estimateTokens(text: string): number {
 // Gemini is the only provider whose mapper never replays the reasoning trace:
 // toGeminiAssistantParts sends content and toolCalls only, because replaying a
 // thought needs the thoughtSignature the parser drops. See #768.
-const PROVIDERS_THAT_DROP_THINKING: ReadonlySet<ProviderName> = new Set([
-  "gemini",
-]);
-
 export function providerReplaysThinking(provider: ProviderName): boolean {
-  return !PROVIDERS_THAT_DROP_THINKING.has(provider);
+  return provider !== "gemini";
 }
 
 function estimateMessageTokens(
--- a/packages/agent/src/history-compaction.ts
+++ b/packages/agent/src/history-compaction.ts
@@ -3,6 +3,7 @@ import type {
   CompactionResponse,
   LlmToolDefinition,
   ProviderClient,
+  ProviderName,
 } from "@nakama/core";
 import {
   estimateUserContentTokens,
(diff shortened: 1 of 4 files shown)


ahmadrosid (commented):
Core fix is right — Gemini shouldn’t inflate the estimate with a trace it never sends. One small cut below.

packages/agent/src/history-compaction.ts
```diff
@@ -78,7 +79,21 @@ function estimateTokens(text: string): number {
   return Math.ceil(text.length / TOKEN_ESTIMATE_RATIO);
 }
 
-function estimateMessageTokens(messages: readonly ChatMessage[]): number {
+// Gemini is the only provider whose mapper never replays the reasoning trace:
+// toGeminiAssistantParts sends content and toolCalls only, because replaying a
+// thought needs the thoughtSignature the parser drops. See #768.
+const PROVIDERS_THAT_DROP_THINKING: ReadonlySet<ProviderName> = new Set([
```
ahmadrosid: Only one name in the set — `return provider !== "gemini"` keeps the comment and drops the Set. You can grow a collection if a second provider ever lands.

vianbas:
Done, the set is now `provider !== "gemini"` with the comment kept above it. Rebased onto current `main` while I was there.

I checked that the tests still hold the behaviour rather than assuming an equality and a one-member lookup are interchangeable to them: removing the gate fails four tests, inverting it fails six. Suite is 2784 pass and 0 fail across eleven workspaces, `check` and `knip` clean.

PR #793, opened 2026-08-31
ponytail(platform): one typing loop for the three channel workers

### LINKED ISSUE #790: ponytail(platform): one typing loop for the three channel workers, drop the session-store re-exports
## Summary

Two files under `apps/platform/{telegram,discord,whatsapp}/src/` carry no channel-specific logic:

- `session-store.ts` is one line, byte-identical in all three: `export { ChannelSessionStore as SessionStore } from "@nakama/core/channel-session-store";`
- `typing-indicator.ts` is 60 / 60 / 63 lines. Telegram and Discord differ by 14 lines and every one of them is the import, the refresh constant, the parameter type, the send call, or comment wording. The loop itself (serialized sends, `active` re-checked before each send, swallowed error) is identical in all three.

Goal: the loop lives once in `@nakama/core` and each worker passes its own send closure and refresh interval. The three re-export files go away and call sites import `ChannelSessionStore` directly. About 120 lines and 3 files less, no behaviour change.

## Why it matters

A fix to that loop currently has to be made three times. #246 is what happens when it is not: the typing fix from #239 landed on Discord and never reached Telegram. The comment explaining why sends are serialized has already drifted in wording across the three copies. #562 is the next fix queued for this same area.

- `packages/core/src/` for the shared loop
- `apps/platform/{telegram,discord,whatsapp}/src/typing-indicator.ts` and their `.test.ts`
- `apps/platform/{telegram,discord,whatsapp}/src/session-store.ts`, 9 import sites found by `grep -rn 'from "./session-store"' apps/platform`

## Acceptance criteria

- [x] one `createTypingLoop(send, refreshMs)` in `@nakama/core`, three workers pass a closure
- [x] refresh interval stays per channel: 4000 ms Telegram, 8000 ms Discord, 4000 ms WhatsApp
- [x] WhatsApp still sends nothing while its socket is null
- [x] `stop()` still wins a ping flood: no send after stop, asserted per channel
- [x] the three `session-store.ts` files are deleted, call sites import `ChannelSessionStore`
- [x] `bun run test` green across all workspaces

fajarhide:
## Summary

Move the channel typing loop into `@nakama/core` and delete the two files each worker kept copies of.

**Before:** `createTypingLoop` existed three times, 183 lines. Telegram and Discord differed by 14 lines: the import, the refresh constant, the parameter type, the send call, and comment wording. `session-store.ts` was a one-line re-export renaming `ChannelSessionStore`, while `index.ts` already imported the real name.

**After:** one `createTypingLoop(send, refreshMs)` in core, called with a closure per channel. 180+/483-, 4 files gone.

Why safe:
1. The loop body is unchanged. Same single interval, same serialized chain, same `active` re-check before each send, so the send path costs one extra closure call per refresh and nothing else. No new timer, no new allocation per ping.
2. The refresh interval stays a parameter and stays per channel, asserted by a test that fails if it becomes a constant in core.
3. The rename is mechanical and the type checker covers it: no new `tsc` error mentions `ChannelSessionStore` or `createTypingLoop`.

## Test plan

- [x] `bun run test` across all workspaces: **2753 pass, 0 fail**, exit 0. The count drops by 5 against main because three copies of the same three cases became one file of four.
- [x] The three worker suites after the move: 289 pass, 0 fail
- [x] Interval kept per channel, checked in the tree:

```
$ grep -rn "TYPING_REFRESH_MS = " apps/platform/*/src/chat-handler.ts
telegram/src/chat-handler.ts:66:const TYPING_REFRESH_MS = 4000;
discord/src/chat-handler.ts:72:const TYPING_REFRESH_MS = 8000;
whatsapp/src/chat-handler.ts:45:const TYPING_REFRESH_MS = 4000;
```

- [x] WhatsApp still sends nothing without a socket (`typingSocket?.sendPresenceUpdate(...) ?? Promise.resolve()`). The deleted WhatsApp test never passed a null socket, so no coverage was lost here.

Closes #790


ahmadrosid:
Closing as superseded by #802, which already merged the shared typing loop + deleted the \`session-store\` re-exports (and went further with shared chat lock / turn-artifact delivery).

Issue #474, opened 2026-08-25
The seven channel subsets are not derived from AGENT_CHANNELS, so a ninth channel is silent at all of them


vianbas:
## Summary

#461 listed seven further places that hold a hand-written subset of the channels rather than a copy of all eight. #463 derived the six full lists from `AGENT_CHANNELS` and did not touch the subsets, which was the right scope for one PR. This is the half that is left: those seven still decide behaviour, and adding a ninth channel is still silent at every one of them.

Measured on `07b0a078`, adding a ninth member to `AGENT_CHANNELS` and changing nothing else:

```
typecheck errors  before 8015   after 8015   new 0
```

Two runs on each side, both byte-stable. The total moves with the checkout, so the number to read is the delta, not 8015.

## Why it matters

One of the seven is where #213 shipped, and it is still hand-written.

`skill-post-turn-review-service.ts` was created in `ae27f7b2` (4 Aug) with the gate `channel !== "web" && channel !== "cli"`, so Discord, Telegram and WhatsApp sessions skipped the review with `channel_not_interactive`. `e0026ea6` (10 Aug) fixed it by replacing that comparison with the five-member `new Set` that is still there, five days and 23 hours later. The list grew; nothing started asking for the decision. A ninth channel repeats it at the same line.

## Implementation touch points

Line numbers moved since #461 and are re-taken here. They are verified on `52670e8f`, which is main as I write this; the measurements above are labelled with the revision they were taken on.

| Where | Holds | Decides |
| --- | --- | --- |
| `skill-post-turn-review-service.ts:197` | 5, plain `Set` of strings | whether the post-turn skill review runs |
| `agent-service.ts:3221` | `web` or `cli` | whether skill-manage tools are in the session |
| `agent-service.ts:3229` | `web` or `cli` | whether a skill usage context is built |
| `chat-history.ts:335` | 4, `as const satisfies readonly AgentChannel[]` | which channels the web history lists |
| `chat-history.ts:342` | 3, equality chain | which channels the web treats as read only |
| `chat-prompt.ts:4` and `:12` | 3, a union plus a map over it | the reply style the model is told to use |
| `agent-service.ts:3226` | `discord` | whether the artifact tools are added |

The last row is out of scope on purpose. A check for one named channel is meant to exclude everything else, and pulling it into a table would drag `createSendDiscordArtifactTools` somewhere it does not belong. Saying so here so the omission reads as a decision.

[...]

PR #797, opened 2026-08-31
fix(skills): refresh catalog after live mutations

puspoaditya:
## Summary

Creating, editing, or deleting a profile skill refreshes that session’s skills catalog before its next turn.

Closes #136

**Before:** `skill_manage` mutations left the in-memory session catalog stale until a new session was created.
**After:** live mutations call `onSkillCatalogChange`, and `AgentService` drops that session’s cached chat instance so the next turn rebuilds the catalog.

Why safe:
1. Only successful live mutations invalidate; approval-gated proposals remain untouched because they are not live.
2. Session history stays persisted; only the cache is rebuilt on the next resolve.
3. Regression coverage asserts create and delete both signal invalidation.

Only re-add the cache without invalidation if catalog composition must deliberately remain immutable throughout a session.

## Test plan
- [x] `bun test apps/server/src/tools/skill-manage-tool.test.ts apps/server/src/services/agent-service.test.ts` (40 pass)
- [x] `bun x ultracite check packages/core/src/contract.ts apps/server/src/tools/skill-manage-tool.ts apps/server/src/tools/skill-manage-tool.test.ts apps/server/src/services/agent-service.ts`
- [x] `NODE_OPTIONS="--max-old-space-size=2048" bun run build`
- [ ] Create a skill in an existing web or CLI chat, then use it in the next turn without opening a new chat

PR #798, opened 2026-08-31
refactor(channels): derive the channel subsets from AGENT_CHANNELS

vianbas:
## Summary

Adding a channel to `AGENT_CHANNELS` now fails the typecheck at each site that has to decide about it, instead of compiling clean and going silent at seven of them.

**Before:** seven hand-written subsets decide behaviour, and a ninth channel adds zero typecheck errors, so nothing asks whether it should be interactive, listed, read only, or given a reply style.
**After:** five total maps over `AgentChannel` back those seven sites. A ninth channel adds 14 errors across five files, none of them in `contract.ts`.

Why safe:
1. The map values reproduce the old predicates exactly. `tsc` on the branch has the same error set as main, 8298 either way, 0 new and 0 gone, each side run twice byte-identical.
2. Twelve behaviour changes were written against both main and this branch and scored by each tree's own suite, each proved applied before the suite ran. Main catches 4 of 12, this branch 12 of 12.
3. The `skill_manage` offer and its runtime gate now read one map, so the tool cannot be attached in a channel that then refuses to run it.

Only re-add the hand-written lists if you want the two `agent-service` sites to diverge; today they are the same list, and splitting them later means splitting one map.

Two things worth knowing, both raised in #474 and neither settled by this PR. `StoredSessionRecord.channel` is still `string`; this reuses the `parseAgentChannel` that already existed in `agent-service` rather than narrowing the column, which is a separate change. And `apps/server/src/tools/skill-manage-tool.ts` was missing from the issue's table, so it is included here and the issue comment says so.

## Test plan
- [x] `bun run test` (2768 pass, 0 fail across 11 workspaces; main is 2758)
- [x] `bun run check` and `bun run knip` (both exit 0, no findings)
- [x] `bun x tsc@5.9.3 --noEmit -p tsconfig.json` before and after adding a ninth channel, twice per side (0 new on main, 14 new here)
- [ ] Open a Telegram or Discord session in the web history panel and confirm it still renders read only, and a web session still writable

Refs #474

### DIFF
--- a/apps/server/src/services/skill-post-turn-review-service.test.ts
+++ b/apps/server/src/services/skill-post-turn-review-service.test.ts
@@ -201,55 +201,6 @@ describe("SkillPostTurnReviewService", () => {
     expect(ran).toBe(0);
   });
 
-  test("skips headless automation channel as channel_not_interactive", async () => {
-    const db = createInMemoryDatabaseAdapter();
-    const now = new Date().toISOString();
-    await db.upsertOrganization({
-      createdAt: now,
-      id: "org_1",
-      name: "Org",
-      skillsPostTurnReview: true,
-      slug: "org",
-      updatedAt: now,
-    });
-    await db.upsertProfile({
-      createdAt: now,
-      id: "profile_1",
-      isSuper: false,
-      model: null,
-      name: "Bot",
-      orgId: "org_1",
-      systemPrompt: "",
-      updatedAt: now,
-    });
-    await db.upsertSession({
-      agentQuestionnaire: null,
-      agentTodos: [],
-      channel: "automation",
-      createdAt: now,
-      id: "session_1",
-      model: null,
-      orgId: "org_1",
-      profileId: "profile_1",
-      title: null,
-      userId: "user_1",
-    });
-
-    let ran = 0;
-    const service = new SkillPostTurnReviewService(
-      db,
-      () => null,
-      async () => {
-        ran += 1;
-      }
-    );
-
-    expect(await service.runPostTurnSkillReview("session_1")).toBe(
-      "channel_not_interactive"
-    );
-    expect(ran).toBe(0);
-  });
-
   test("skips when flag disabled", async () => {
     const db = createInMemoryDatabaseAdapter();
     const now = new Date().toISOString();
--- a/apps/server/src/services/skill-post-turn-review-service.test.ts
+++ b/apps/server/src/services/skill-post-turn-review-service.test.ts
@@ -140,7 +140,7 @@ describe("evaluatePostTurnReviewTurnEligibility", () => {
 
 describe("SkillPostTurnReviewService", () => {
   test("runs runner once for eligible interactive channels when flag on and manage-skills assigned", async () => {
-    for (const channel of ["web", "discord", "telegram", "whatsapp"]) {
+    for (const channel of ["web", "cli", "discord", "telegram", "whatsapp"]) {
       const db = createInMemoryDatabaseAdapter();
       await seedEligibleTurn(db, channel);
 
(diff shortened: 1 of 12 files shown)


vianbas:
Three things a review here would otherwise have to work out from scratch.

**#797 edits four of the same files, and nothing collides.** Hunk starts on the pre-merge file, their head `c384dbf8` against this one `822c44f6`:

| file | #797 | this PR |
| --- | --- | --- |
| `skill-manage-tool.ts` | 112, 127, and six more from 422 to 563 | 1, 32, 46 |
| `skill-manage-tool.test.ts` | 122, 416, 429 | 393 |
| `agent-service.ts` | 3510 | 104, 148, 206, 3335, 3343, 3828 |
| `contract.ts` | 2246 | 89 |

I compared the full ranges rather than these starts, and none of them overlaps. The tests added here land between their 122 and their 416, so either merge order applies clean. Four shared files reads worse than it is.

**`formatSessionChannelLabel` is left alone deliberately.** It names four channels and ends in `default: return channel`, so a ninth channel renders as its own id instead of disappearing. That fallback is visible rather than silent, which is the opposite of the failure this PR is about, and making the map total would turn a working default into a decision someone has to make. No changed line in this PR touches it; the one mention in the diff is a context line. Worth saying, because otherwise the omission reads as something I missed.

**This PR does not close #474, and that may not be what you want.** The body says `Refs`. Criterion 5 in the issue asks that whatever is decided for `session.channel` be applied at one boundary rather than repeated at each call site. What this does is reuse the `parseAgentChannel` that already existed privately in `agent-service`, move it to `packages/core`, and share it across all five readers. `StoredSessionRecord.channel` is still `string`. If narrowing the column is what that criterion meant, this is not finished and the issue should stay open. If one shared narrow satisfies it, say so and I will change the body to `Closes`.

apps/server/src/services/skill-post-turn-review-service.test.ts
```diff
@@ -158,6 +158,49 @@ describe("SkillPostTurnReviewService", () => {
     }
   });
 
+  // The other half of the same table. Together these two cover all eight
+  // channels, so flipping any value in POST_TURN_REVIEW_CHANNELS fails a test
+  // rather than only shifting behaviour.
+  test("skips the non-interactive channels even when the turn is eligible", async () => {
```
ahmadrosid, line 164: This already runs automation (plus task/subagent) through the same outcome. The older `skips headless automation channel` case below is the same lock with more setup — fine to drop it.

apps/server/src/tools/skill-manage-tool.test.ts
```diff
@@ -391,6 +391,44 @@ Profile body.
...
+        memberContext({ channel })
+      );
+
+      expect(result).toMatchObject({ created: true, name: "research-paper" });
+      // Each pass needs its own config dir; afterEach only removes the last.
+      await rm(configDir, { force: true, recursive: true });
+    }
+  });
+
+  test("refuses every channel outside the skill-management table", async () => {
```
ahmadrosid, line 401: This already includes telegram. The single-channel `refuses non-interactive channel context` just above doesn't add a second property — drop it and keep the full-table pair.

apps/web/src/lib/chat-history.test.ts
```diff
@@ -435,3 +438,33 @@ describe("formatSessionTimestamp", () => {
...
+
+  test("marks the three messaging channels read only and no others", () => {
+    const readOnly = AGENT_CHANNELS.filter((channel) =>
+      isReadOnlySessionChannel(channel)
+    );
+
+    expect(readOnly).toEqual(["telegram", "whatsapp", "discord"]);
+  });
+
+  test("every listed channel except web is read only", () => {
```
ahmadrosid: With the exact history list and the global read-only filter above, "only web is writable among listed" is already implied. Two tests are enough to pin both tables.

ahmadrosid (commented):
Core change is solid — total maps over `AgentChannel` are the right shape. A few tests now double-cover the same lock once the full-table cases landed.

vianbas:
All three applied, and rebased onto current `main`.

Telegram and the writable set were redundant: with them gone the mutation score is unchanged, nothing dropped to zero.

The automation case was not. It seeded a session with no manage-skills skill, where the table test I replaced it with uses `seedEligibleTurn`, which assigns one. The order is session, channel, profile, org, flag, manage-skills, so it pinned where the gate sits in that sequence and mine pins only its result. Moving the gate below the manage-skills check is caught before the removal, by that test alone, and survives after it. Concretely: a non-interactive session that also lacks manage-skills would report `manage_skills_unassigned` instead of `channel_not_interactive`.

So I did not restore it. The last commit pins the ordering as its own property, `reports the channel before a later check can claim the skip`, and the duplicate you flagged stays deleted.

Suite is 2788 pass, 0 fail. If you would rather leave the ordering unpinned, say so and I will drop that test too.

PR #800, opened 2026-08-31
fix(server): stop returning raw internal error messages to clients

### LINKED ISSUE #360: Stop returning raw internal error messages to clients
# Summary

Unexpected internal errors are returned to clients verbatim, disclosing implementation details.

# Evidence (validated against source)

- `formatServerError` (`packages/core/src/api-error.ts:139-157`) returns `error.message.trim()` verbatim for any `Error` with a non-empty message; its only special-casing is `SyntaxError` and empty messages. Stack traces are excluded (good).
- Consumed by the global handler at `apps/server/src/http/app.ts:58` → the raw message becomes the JSON `error` field on any uncaught error.
- Route-level catches echo `error.message` too: `sessions.ts:485-487,554-557`; `tools.ts:351-354,367-371`; `models.ts:1108-1110` (502 upstream catalog fetch) and `1131-1133`; `setup-import.ts:200-201`; `notification-webhooks.ts:27-30` — the last one on a **public** route (`/v1/notify/:id`).
- No middleware sanitizes error bodies (`app.ts:61-101` adds headers only).

# Impact

SQLite failures, fs errors containing absolute paths, and upstream provider HTTP errors (URLs, model IDs, request IDs) are disclosed to any authenticated user — and in one case unauthenticated. Aids reconnaissance; not a direct breach.

# Fix

- In `formatServerError`: return generic text ("Internal server error") for unexpected errors while logging full detail server-side; keep intentional `NakamaApiError` messages.
- Sweep the listed route catches to stop echoing `error.message` for non-API errors.

## Verification notes

- Confirmed `formatServerError` at `packages/core/src/api-error.ts:139-157`: returns `error.message.trim()` verbatim for any `Error` with a non-empty message; only special-cases `SyntaxError` and empty input; no stack traces are emitted.
- Confirmed the global handler at `apps/server/src/http/app.ts:58` feeds `formatServerError(err)` into the 500 JSON `error` field; `NakamaApiError` messages pass through intentionally at `app.ts:46-52`.
- Confirmed every cited route-level catch echoes raw `error.message`: `apps/server/src/http/routes/sessions.ts:485-487` and `:554-557`; `tools.ts:351-354` and `:367-371`; `models.ts:1108-1110` (502 upstream catalog fetch) and `:1131-1133`; `setup-import.ts:200-201` (`formatImportError`); `notification-webhooks.ts:27-30`.
- Confirmed `/v1/notify/:id` bypasses auth middleware via `public-routes.ts:27`, so the raw-message echo there is reachable without authentication. Nuance: the handler still validates a per-destination `x-api-key` inside the service, but any non-`NakamaApiError` failure (including internal ones) echoes its message with status 400.
- Confirmed `app.ts:61-101` middleware only rewrites response headers; nothing sanitizes error bodies.
- All cited line numbers matched the current source exactly; no corrections needed.

## Acceptance criteria

- [ ] An uncaught unexpected error (e.g., simulated SQLite/fs failure) returns a generic 500 body ("Internal server error"-class text), not the underlying `error.message`
- [ ] Full error detail for such failures is still logged server-side
- [ ] Intentional `NakamaApiError` responses still reach clients unchanged (status + message preserved)
- [ ] Malformed JSON requests still return the friendly "Invalid JSON in request body." 400
- [ ] The listed route catches (`sessions.ts`, `tools.ts`, `models.ts`, `setup-import.ts`, `notification-webhooks.ts`) no longer echo raw `error.message` for non-API errors; existing client-facing behavior keeps working (e.g., "not found" → 404 mapping in the tools routes)
- [ ] The public `/v1/notify/:id` route never returns internal error text to unauthenticated callers
- [ ] Regression test: an endpoint that throws a plain `Error` carrying an implementation-detail message returns the generic text, while a thrown `NakamaApiError` still returns its intended message and status

## Related issues

- #357 — same request/response hygiene theme: #357 hardens what comes in (runtime body validation + size limits), this issue hardens what goes out (error bodies). Independent fixes.
- #361 — shares the unauthenticated-surface concern: #361(c) avatar endpoint auth and this issue's `/v1/notify/:id` raw-error echo both affect pre-auth routes; #361(b) setup-import caps touch the same file whose catch echoes raw messages here. Overlapping surface only; fixes don't conflict.

---
Found by internal security audit 2026-08-21. Verdict VALID / MEDIUM after second-pass validation.

xietsunzao:
## Summary

Unexpected internal errors reached clients verbatim. `formatServerError` returned any `Error`'s raw `.message`, so SQLite failures, fs errors with absolute paths, and upstream provider details leaked through the global error handler and several route catches, including one unauthenticated route.

**Before:** `formatServerError` returned `error.message.trim()` for any `Error`, only special-casing `SyntaxError`.
**After:** it only passes through a `NakamaApiError`'s own message (the codebase's existing signal for "this is safe to show"), logs everything else server-side, and returns a generic message for the rest.

Why safe:
1. The cited catches (`sessions.ts`, `tools.ts`, two catches in `models.ts`, `setup-import.ts`, `notification-webhooks.ts`) now build their response through `formatServerError` instead of hand-echoing `error.message`. `tools.ts`'s "not found" → 404 status mapping still reads the real message server-side, only the response body changed.
2. Two services threw plain `Error` for legitimate validation messages that would have gone generic too: `services/data-portability.ts` (archive import validation, "Invalid ZIP archive." and friends) and `packages/core/src/notification-destinations.ts` (webhook payload validation). Both now throw `NakamaApiError` so those messages keep reaching clients, no UX regression.
3. `shared.ts`'s chat-turn timeout messages and `coding-harness-install-stream.ts`'s installer failure messages also ran through `formatServerError` before this change and would have gone generic too, breaking their existing tests. Neither is cited by the issue and both are diagnostic text for the operation's own initiator, not an arbitrary API response, so they were reverted to their original raw-message behavior instead of converting their throw sites.

Closes #360

Six more files have the same raw-echo pattern (`composio.ts`, `notification-destinations.ts` route, `profile-portability.ts`, `auth.ts`, `system.ts`, plus ~18 more catches in `models.ts` itself) but aren't cited by the issue. Left out of this PR to keep it reviewable; candidate for a follow-up.

## Test plan
- [x] `bun test` in `packages/core`: 728 pass, 3 fail (Windows symlink `EPERM`, pre-existing, unrelated)
- [x] `bun test` full server workspace: 1057 pass, 29 fail. Same 29 fail on unmodified `upstream/main` (missing python3/bash/agent-browser CLIs on this dev machine, confirmed via a stash-based diff)
- [x] `bunx tsc --noEmit` whole repo: baseline 8302 errors, after 8307. The +5 is pre-existing `typeof fetch` "preconnect" noise (already present repo-wide) showing up in the 2 new test files; zero new errors from the actual source changes
- [x] `bun run knip` clean
- [x] New regression tests: `formatServerError` unit tests (NakamaApiError passthrough, generic for plain Error, SyntaxError), plus route-level tests for `tools.ts`, `models.ts`, `sessions.ts`, `setup-import.ts`, `notification-webhooks.ts` asserting a thrown implementation-detail message never reaches the response body

### DIFF
--- a/apps/server/src/http/routes/models-error-format.test.ts
+++ b/apps/server/src/http/routes/models-error-format.test.ts
@@ -30,33 +30,4 @@ describe("model routes error formatting", () => {
       error: "An unexpected server error occurred.",
     });
   });
-
-  test("discover models does not leak an unexpected error's message", async () => {
-    const { app, databaseAdapter } = createMinimalHonoApp({
-      agent: {
-        discoverModels: async () => {
-          throw new Error(
-            "getaddrinfo ENOTFOUND internal-provider.nakama.local"
-          );
-        },
-      },
-    });
-    const session = await setupFreshInstallSession(app, databaseAdapter);
-
-    const response = await app.fetch(
-      new Request("http://localhost:4310/v1/models/discover", {
-        body: JSON.stringify({ baseUrl: "https://example.com" }),
-        headers: session.headers({
-          "Content-Type": "application/json",
-          "X-CSRF-Token": session.csrfToken,
-        }),
-        method: "POST",
-      })
-    );
-
-    expect(response.status).toBe(400);
-    await expect(response.json()).resolves.toEqual({
-      error: "An unexpected server error occurred.",
-    });
-  });
 });
--- a/apps/server/src/http/routes/models-error-format.test.ts
+++ b/apps/server/src/http/routes/models-error-format.test.ts
@@ -1,0 +1,62 @@
+import { afterEach, describe, expect, test } from "bun:test";
+import { createMinimalHonoApp } from "../test-app-helpers";
+import { setupFreshInstallSession } from "../test-session-helpers";
+
+describe("model routes error formatting", () => {
+  const originalFetch = globalThis.fetch;
+
+  afterEach(() => {
+    globalThis.fetch = originalFetch;
+  });
+
+  test("catalog fetch failure does not leak the upstream error's message", async () => {
+    globalThis.fetch = async () => {
+      throw new Error(
+        "connect ETIMEDOUT 198.51.100.4:443 at TCPConnectWrap.afterConnect"
+      );
+    };
+
+    const { app, databaseAdapter } = createMinimalHonoApp();
+    const session = await setupFreshInstallSession(app, databaseAdapter);
+
+    const response = await app.fetch(
+      new Request("http://localhost:4310/v1/model-catalogs/openrouter", {
+        headers: session.headers(),
+      })
+    );
+
+    expect(response.status).toBe(502);
+    await expect(response.json()).resolves.toEqual({
+      error: "An unexpected server error occurred.",
+    });
+  });
+
+  test("discover models does not leak an unexpected error's message", async () => {
+    const { app, databaseAdapter } = createMinimalHonoApp({
+      agent: {
+        discoverModels: async () => {
+          throw new Error(
+            "getaddrinfo ENOTFOUND internal-provider.nakama.local"
+          );
+        },
+      },
+    });
+    const session = await setupFreshInstallSession(app, databaseAdapter);
+
+    const response = await app.fetch(
+      new Request("http://localhost:4310/v1/models/discover", {
+        body: JSON.stringify({ baseUrl: "https://example.com" }),
+        headers: session.headers({
+          "Content-Type": "application/json",
+          "X-CSRF-Token": session.csrfToken,
+        }),
+        method: "POST",
+      })
+    );
+
+    expect(response.status).toBe(400);
+    await expect(response.json()).resolves.toEqual({
+      error: "An unexpected server error occurred.",
+    });
+  });
+});
--- a/apps/server/src/http/routes/models.ts
+++ b/apps/server/src/http/routes/models.ts
@@ -10,6 +10,7 @@ import {
   type DiscordSettingsResponse,
   type DiscoverModelsRequest,
   type EmailSettingsResponse,
+  formatServerError,
   type GenerateImageRequest,
   type GenerateImageResponse,
   type ImageGenerationSettingsResponse,
@@ -1184,8 +1185,7 @@ export function registerModelRoutes(
     try {
       return json(await getExternalModelCatalog(catalogId));
     } catch (error) {
-      const message = error instanceof Error ? error.message : String(error);
-      return errorResponse(message, 502);
+      return errorResponse(formatServerError(error), 502);
     }
   });
 
@@ -1207,8 +1207,7 @@ export function registerModelRoutes(
       const result = await agent.discoverModels(body);
       return json<ModelsResponse>(result);
     } catch (error) {
-      const message = error instanceof Error ? error.message : String(error);
-      return errorResponse(message, 400);
+      return errorResponse(formatServerError(error), 400);
     }
   });
 
--- a/apps/server/src/http/routes/models.ts
+++ b/apps/server/src/http/routes/models.ts
@@ -1202,13 +1202,8 @@ export function registerModelRoutes(
   app.post("/v1/models/discover", async (c) => {
     requireOrgAdminOrPlatformAdminFromContext(c);
     const body = await readJson<DiscoverModelsRequest>(c.req.raw);
-
-    try {
-      const result = await agent.discoverModels(body);
-      return json<ModelsResponse>(result);
-    } catch (error) {
-      return errorResponse(formatServerError(error), 400);
-    }
+    const result = await agent.discoverModels(body);
+    return json<ModelsResponse>(result);
   });
 
   app.get("/v1/providers", async (c) => {
--- a/apps/server/src/http/routes/notification-webhooks.ts
+++ b/apps/server/src/http/routes/notification-webhooks.ts
@@ -1,5 +1,5 @@
 import type { NotificationWebhookRequest } from "@nakama/core";
-import { NakamaApiError } from "@nakama/core";
+import { formatServerError, NakamaApiError } from "@nakama/core";
 import { NotificationWebhookService } from "../../services/notification-webhook-service";
 import type { ServerOptions } from "../context";
 import { errorResponse, readJson } from "../shared";
@@ -24,10 +24,7 @@ export function registerNotificationWebhookRoutes(
       if (error instanceof NakamaApiError) {
         return errorResponse(error.message, error.status);
       }
-      return errorResponse(
-        error instanceof Error ? error.message : String(error),
-        400
-      );
+      return errorResponse(formatServerError(error), 400);
     }
   });
 }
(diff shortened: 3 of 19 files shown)


fajarhide (commented):
Hardening `formatServerError` around `NakamaApiError` is the right seam, and converting the two validation services to throw it instead of plain `Error` is the right call. The route-level half mostly reimplements `app.onError`, which already does this.

I checked the branch out, applied the cuts below and ran the server suite: 1086 pass, 0 fail. Only injected failures, nothing against a live provider.

@ahmadrosid one of these is yours to call: whether a route catch should exist at all once the global handler covers it. @xietsunzao the `tools.ts` one changes what a user sees, so it is worth fixing either way.

apps/server/src/http/routes/notification-webhooks.ts
```diff
@@ -24,10 +24,7 @@ export function registerNotificationWebhookRoutes(
       if (error instanceof NakamaApiError) {
         return errorResponse(error.message, error.status);
       }
-      return errorResponse(
-        error instanceof Error ? error.message : String(error),
-        400
-      );
+      return errorResponse(formatServerError(error), 400);
```
fajarhide: `app.onError` at `app.ts:55` already does this, `NakamaApiError` keeping its own status and everything else going through `formatServerError`. Drop the catch and let it throw: the file loses 8 lines and three imports, and the disk-I/O case becomes a 500 instead of a 400. Ran it, 4 pass.

apps/server/src/http/routes/setup-import.ts
```diff
@@ -190,13 +191,6 @@ async function assertSetupImportAllowed(
...
-    return errorResponse(error.message, error.status);
-  }
-
-  return errorResponse(formatImportError(error), 500);
-}
-
-function formatImportError(error: unknown): string {
-  return error instanceof Error ? error.message : String(error);
+  const status = error instanceof NakamaApiError ? error.status : 500;
+  return errorResponse(formatServerError(error), status);
```
fajarhide: Same as the webhook catch. `setupImportErrorResponse` is now `app.onError`'s body copied into the file. The only difference is 500 vs the handler's 500, so there is nothing left to keep.

apps/server/src/http/routes/tools.ts
```diff
@@ -349,9 +350,9 @@ export function registerToolRoutes(app: HonoApp, options: ServerOptions): void {
         })
       );
     } catch (error) {
-      const message = error instanceof Error ? error.message : String(error);
-      const status = message.includes("not found") ? 404 : 400;
-      return json({ error: message }, status);
+      const rawMessage = error instanceof Error ? error.message : String(error);
+      const status = rawMessage.includes("not found") ? 404 : 400;
```
fajarhide: The sniff reads a message the response no longer carries, so a missing tool now answers `404 {"error":"An unexpected server error occurred."}`. `agent-service.ts:2619` and `:2673` throw that string already; make them `new NakamaApiError("Tool not found.", 404)` and the sniff goes with them.

apps/server/src/http/shared.ts
```diff
@@ -666,7 +665,9 @@ export function streamMessage(
           error:
             turnSignal.aborted && !timedOut
               ? "Turn cancelled."
-              : formatServerError(error),
+              : error instanceof Error
+                ? error.message
```
fajarhide: This is the raw echo the PR removes elsewhere, just out of reach of a `formatServerError` grep. `failAfter` on line 625 rejects with a repo-authored string, so `new NakamaApiError(message, 504)` there keeps this catch on `formatServerError` and stops a provider-side SQLite failure streaming to the browser. Same for `coding-harness-install-stream.ts:78`.

apps/server/src/http/routes/models-error-format.test.ts
```diff
@@ -0,0 +1,62 @@
...
+import { setupFreshInstallSession } from "../test-session-helpers";
+
+describe("model routes error formatting", () => {
+  const originalFetch = globalThis.fetch;
+
+  afterEach(() => {
+    globalThis.fetch = originalFetch;
+  });
+
+  test("catalog fetch failure does not leak the upstream error's message", async () => {
```
fajarhide, line 12: `api-error.test.ts` already pins "a plain Error never reaches the caller", and `tools.test.ts` proves one route is wired to it. These two files spend 128 lines proving the same property twice more. If the catches above are dropped, they are asserting `app.onError`, which the unit test already owns.

apps/server/src/http/routes/sessions.ts
```diff
@@ -545,8 +545,7 @@ export function registerSessionRoutes(
 
       return json<BranchSessionResponse>(result, 201);
     } catch (error) {
-      const message = error instanceof Error ? error.message : String(error);
-      return errorResponse(message, 400);
+      return errorResponse(formatServerError(error), 400);
```
ahmadrosid: This catch only forces 400 — a corrupt DB branch shouldn't look like a bad request. Drop it and let `app.onError` return 500 with the generic message. Keep the send-message catch below; that one still needs `sessionTurnRegistry.endTurn`.

apps/server/src/http/routes/models.ts
```diff
@@ -1207,8 +1207,7 @@ export function registerModelRoutes(
       const result = await agent.discoverModels(body);
       return json<ModelsResponse>(result);
     } catch (error) {
-      const message = error instanceof Error ? error.message : String(error);
-      return errorResponse(message, 400);
+      return errorResponse(formatServerError(error), 400);
```
ahmadrosid: Same shape as the branch catch: status remap, no other work. Drop it. Keep the catalog catch above if you want the 502 for upstream failures — `onError` can't give you that.

ahmadrosid (commented):
Core change is right — `NakamaApiError` as the safe-to-show signal, and the validation throw sites switched over. Two catches below still exist only to pin a 400; unexpected failures are happier as 500 via `app.onError`.

xietsunzao:
Thanks both, this is exactly right. Pushed be22a2a addressing all of it.

Dropped the route catches that were only reimplementing app.onError:
- notification-webhooks.ts: catch removed entirely.
- setup-import.ts: setupImportErrorResponse and its two call sites removed.
- sessions.ts: the branch catch removed.
- models.ts: the discover-models catch removed.

Kept the two that do work onError can't:
- sessions.ts send-message, which also feeds sessionTurnRegistry.endTurn.
- models.ts catalog fetch, which needs a 502 for upstream failures.

tools.ts: the not-found sniff read a message the body no longer carries. Fixed at the source instead, agent-service.ts now throws NakamaApiError("Tool not found.", 404) for both playground routes, so the sniff is gone and the status comes from the error itself.

shared.ts and coding-harness-install-stream.ts: these got reverted to raw-message echoing in the first round because their tests needed specific timeout/installer messages preserved, and formatServerError would have hidden those. Same fix as the rest of the PR really, the throw sites were the problem. shared.ts's failAfter and agent-browser-service.ts's install failures now throw NakamaApiError, so the catches go back to formatServerError and keep the messages.

Same treatment for a few validation paths I found while making these drops safe: sessions.ts's branchSession (messageIndex bounds, invalid channel) and several agent-service.ts discoverModels checks (missing API key, missing base URL, provider not found) now throw NakamaApiError instead of plain Error, since dropping their catches would have made those messages generic too.

Trimmed the two new test files down to the cases that still exercise a real code path once the redundant catches were gone.

Full server suite: 1057 pass, 28 fail, same pre-existing environment gaps as before (missing python3/bash/agent-browser CLIs on my machine). knip and tsc clean against baseline.

PR #801, opened 2026-08-31
fix(cli): guard against raw-mode leak in readPassword setup

### LINKED ISSUE #615: [cli] setup.ts setRawMode without removing the listener on success path
## Bug
`stdin.setRawMode(true)` is set; on Enter the listener removes itself and sets raw mode back. But if `readPassword` is called twice in `ensureUserConfiguredViaCli`, and a SIGWINCH or other signal arrives between the two, raw mode could leak.

## Files
- `apps/cli/src/setup.ts:22-63`

## Fix
Wrap in try/finally to ensure raw mode is restored.

## Severity
Medium — stdin hygiene.

idsoke:
## Summary
- Fixes #615 — `readPassword()` in `apps/cli/src/setup.ts` could leave the terminal stuck in raw mode if enabling raw mode or attaching the `data` listener threw synchronously (e.g. the tty handle erroring on `resume()`). The `Promise` executor would reject with no cleanup path since the listener that calls `restoreStdin()` was never attached.
- Wrapped the raw-mode setup (`setRawMode(true)` → `resume()` → `setEncoding()` → `on("data", ...)`) in try/catch, tracking whether raw mode was actually enabled, and restore it on failure before rejecting.
- Exported `readPassword` and added `apps/cli/src/setup.test.ts` covering the happy path and the synchronous-setup-failure path (repros the leak against `main` before the fix).

## Test plan
- [x] `bun test` in `apps/cli` — 108 pass
- [x] `ultracite check` on changed files — clean

### DIFF
--- a/apps/cli/src/setup.test.ts
+++ b/apps/cli/src/setup.test.ts
@@ -1,0 +1,88 @@
+import { describe, expect, test } from "bun:test";
+import { EventEmitter } from "node:events";
+import { readPassword } from "./setup";
+
+class MockStdin extends EventEmitter {
+  isTTY = true;
+  rawModeEnabled = false;
+  paused = true;
+  failOnResume = false;
+
+  setRawMode(enabled: boolean) {
+    this.rawModeEnabled = enabled;
+    return this;
+  }
+
+  isPaused() {
+    return this.paused;
+  }
+
+  resume() {
+    this.paused = false;
+    if (this.failOnResume) {
+      throw new Error("simulated stdin failure");
+    }
+    return this;
+  }
+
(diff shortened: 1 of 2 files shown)


idsoke:
Reopening this PR — I noticed I hadn't been assigned to issue #615 before submitting. @ahmadrosid could you assign me to the issue? Happy to keep this PR open in the meantime.

ahmadrosid (commented):
Looks lean — ship. The try/catch + rawModeEnabled flag is the right size for the leak, and the two tests cover the happy restore and the sync-setup failure without extra scaffolding.

Issue #367, opened 2026-08-21
Enterprise compliance: session governance APIs (list/revoke, offboarding revocation)


vincentri:
**Enterprise compliance gap** — identified in the nakama security & compliance audit (2026-08-21). Required for SOC 2 CC6.2/CC6.3 (timely deprovisioning).

# Summary

Users cannot see or manage their active login sessions, admins cannot revoke a user's sessions, and removing a member from an org leaves their session cookie fully alive.

# Why this blocks enterprise compliance

- Offboarding SLAs: when an employee leaves, access must die now — not at cookie expiry (7 days here).
- Auditors test "revoke a stolen session" — today the only lever is logout-of-that-device.

# Current state (verified)

- DB adapter surface for `browser_sessions` (`packages/db/src/types.ts:553-870`): create, get-by-hash, revoke-by-hash (own logout, `auth.ts:534`), update-active-org, update-last-used. **No list, no revoke-by-user.**
- `removeMember` (`org-service.ts:477-484`) deletes only the membership row — removed users keep working sessions.
- No session-management UI; web "session" vocabulary is chat history, not login sessions.
- Related already-filed issue: #355 (password change doesn't revoke sessions).

# Proposed approach

1. Adapter methods: `listBrowserSessionsByUser`, `revokeBrowserSessionById` (scoped to owner or platform-admin).
2. Routes: `GET /v1/auth/sessions` (own devices), `DELETE /v1/auth/sessions/:sessionId` (self), platform-admin revoke-all-for-user.
3. Call revoke-all inside `removeMember` and `changePassword`.
4. Minimal settings UI listing devices with last-used timestamps.

# Acceptance criteria

- [x] Removed member's sessions are revoked at removal time (test).
- [x] Users can enumerate and revoke their own sessions.
- [x] Platform admin can force-revoke all sessions for any user.

vincentri:
Extension worth scoping with this work: an org-wide emergency "revoke all sessions" action (platform-admin), beyond per-user revocation — needed for breach containment.

fajarhide:
Took the first acceptance criterion in #803: `removeMember` now revokes the removed user's browser sessions, reusing the `revokeBrowserSessionsForUser` call that `changePassword` already makes.

[...]

PR #803, opened 2026-09-01
fix(orgs): revoke a removed member's sessions at removal time

fajarhide:
## Summary

Removing someone from an organization now ends their login sessions at removal instead of at cookie expiry.

**Before:** `removeMember` deleted the membership row and nothing else, so the removed user's session cookie stayed valid on `/v1/auth/*` for up to 7 days.
**After:** it also calls `revokeBrowserSessionsForUser`, the same revocation `changePassword` already uses 250 lines down the same file.

Why safe:
1. Reuses the existing adapter method and the existing call shape, so there is no second revocation path to keep in sync.
2. Org-scoped routes already 404 a non-member in `org-middleware.ts`, so this only closes the `/v1/auth/*` remainder rather than changing how org access works.
3. Full suite green across all 11 workspaces.

Only follow up if revoking every session proves too blunt for multi-org users. The narrower version revokes only sessions whose active org is the one they left, and that needs a new adapter method for one caller.

## Test plan
- [x] `bun test apps/server/src/http/org-member-removal-sessions.test.ts` (1 pass). On `main` the same test fails at the assertion that matters: `Expected: 401, Received: 200`.
- [x] `bun run test` (all 11 workspaces exited 0; 1086 server, 733 core, 313 web, 92 db)
- [ ] Remove a member from the dashboard while they have a tab open, then confirm their next request bounces to the login page.

The test asserts through `/v1/auth/me` on purpose. An org-scoped route would have passed before this change too, since the middleware already blocks a non-member there.

`Refs #367`, not `Closes`. This is the first of its three acceptance criteria; `listBrowserSessionsByUser` and a platform-admin force-revoke have no adapter method or route yet, and are left out rather than bundled.

### DIFF
--- a/apps/server/src/http/org-member-removal-sessions.test.ts
+++ b/apps/server/src/http/org-member-removal-sessions.test.ts
@@ -1,0 +1,104 @@
+import { describe, expect, test } from "bun:test";
+import { setupTestConfigDir } from "../test-config-dir";
+import { createMinimalHonoApp } from "./test-app-helpers";
+import {
+  loginPlatformAdminSession,
+  loginUserSession,
+} from "./test-session-helpers";
+
+setupTestConfigDir("nakama-member-removal-sessions-test-");
+
+describe("removing an org member", () => {
+  test("revokes the removed member's browser sessions", async () => {
+    const { app, authService, databaseAdapter } = createMinimalHonoApp();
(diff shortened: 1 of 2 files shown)


ahmadrosid (commented):
Looks lean — ship. One call to the existing revoke path plus a single `/v1/auth/me` check is the right size for this.

PR #804, opened 2026-09-01
fix(db): unassign profiles inside deleteTool transaction

### LINKED ISSUE #605: [db] seed.ts removal loops are not transactional
## Bug
`removeLegacyBuiltinTools`, `removeDeprecatedBuiltinTools`, `removeDeprecatedServerTools`, `removeUnsupportedTools` each iterate `tools`, unassign from every profile, then delete. None of the per-tool cycles are atomic.

## Files
- `packages/db/src/seed.ts:58-138`

## Fix
Wrap each tool's unassign+delete cycle in `db.transaction`.

## Severity
Medium — partial state during seed.

akmalsyrf:
## Summary

Seed/admin tool delete now unassigns every profile and deletes the tool in one SQLite transaction — no partial profile_tools leftovers.

**Before:** seed loops unassigned per profile then `deleteTool`, with FK cascade off, so a mid-loop failure could leave assignments without a tool (or the reverse race).
**After:** `deleteTool` runs `DELETE FROM profile_tools WHERE tool_id=?` + `DELETE FROM tools` inside `db.transaction`; seed just calls `deleteTool`.

Why safe:
1. Same end state as the old unassign-then-delete loops
2. Existing seed cleanup tests still pass; added multi-profile delete assertion
3. `profile-service.deleteTool` benefits from the same atomic path

Only re-add / follow up if another adapter without this transactional delete is introduced.

## Test plan
- [x] `bun test packages/db/src/seed.test.ts` (15 pass)
- [x] `deleteTool unassigns every profile before deleting` — multi-profile assignment cleared atomically
- [x] Existing seed cleanup paths (unsupported/deprecated/legacy) still pass without per-profile unassign loops

Closes #605

### DIFF
--- a/packages/db/src/adapters/sqlite.ts
+++ b/packages/db/src/adapters/sqlite.ts
@@ -670,6 +670,14 @@ function createSqliteDatabaseAdapter(db: Database): DatabaseAdapter {
     DELETE FROM profile_tools
     WHERE profile_id = ? AND tool_id = ?
   `);
+  const unassignToolFromAllProfilesStmt = db.prepare(`
+    DELETE FROM profile_tools
+    WHERE tool_id = ?
+  `);
+  const deleteToolEverywhereTransaction = db.transaction((toolId: string) => {
+    unassignToolFromAllProfilesStmt.run(toolId);
+    return deleteToolStmt.run(toolId);
+  });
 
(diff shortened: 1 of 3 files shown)


ahmadrosid (commented):
Looks lean — ship. Unassign+delete in one transaction, seed loops gone, one assert that covers the multi-profile case. Nothing to cut.

PR #805, opened 2026-09-01
fix(client): throw NakamaAuthExpiredError on unchanged 401 token

### LINKED ISSUE #604: [client] 401 retry indistinguishable from generic 401 when token file missing
## Bug
On 401 the client refreshes from `loadLocalAuthToken()` and retries. The guard is "freshToken !== this.authToken" — if the file is gone or empty, no retry happens (good), but if the file exists with the same expired token, no retry happens either and the caller sees a 401 with no indication it was even attempted.

## Files
- `packages/client/src/client.ts:2327-2339`

## Fix
Throw a distinct `NakamaApiError` subclass (e.g. `auth-expired`) when 401 is returned without a token change.

## Severity
Medium — UX / error reporting.

akmalsyrf:
## Summary

Local-token 401 reload failures now throw `NakamaAuthExpiredError` instead of a generic `NakamaApiError`, so CLI/clients can tell “session expired” from other auth failures.

**Before:** unchanged/missing-token 401s fell through to `createApiError` (same shape as any other 401).
**After:** no usable token change → `NakamaAuthExpiredError`; reload that still 401s after retry → same subclass.

Why safe:
1. Extends `NakamaApiError` — existing `instanceof NakamaApiError` / `formatClientError` keep working
2. Successful reload+retry path unchanged (existing test still green)
3. Exported from `@nakama/client` for callers that want the narrow type

Only re-add / follow up if browser cookie sessions need a parallel distinct code (this path is local-token reload).

## Test plan
- [x] `bun test packages/client/src/client.test.ts` (15 pass)
- [x] Unchanged local token on 401 → `NakamaAuthExpiredError` (name + status 401)
- [x] Missing token file on 401 → `NakamaAuthExpiredError`
- [x] `fetchRaw` skips reload on `/v1/auth/local-token/rotate` (parity with `request()`)

Closes #604

### DIFF
--- a/packages/client/src/client.test.ts
+++ b/packages/client/src/client.test.ts
@@ -367,6 +367,13 @@ test("non-browser clients throw NakamaAuthExpiredError when 401 token is unchang
       "tc_local_same\n",
       "utf8"
     );
+    await saveUserConfig({
+      defaultProviderId: null,
+      localAuthTokenHash: createHash("sha256")
+        .update("tc_local_same")
+        .digest("hex"),
+      providers: [],
+    });
 
     const client = new NakamaClient({
       authToken: "tc_local_same",
@@ -394,7 +401,7 @@ test("non-browser clients throw NakamaAuthExpiredError when 401 token is unchang
   }
 });
 
-test("non-browser clients throw NakamaAuthExpiredError when token file is missing", async () => {
+test("non-browser clients throw NakamaAuthExpiredError when token reload still 401s", async () => {
   const configDir = await mkdtemp(
     join(tmpdir(), "nakama-client-auth-missing-")
   );
--- a/packages/client/src/client.test.ts
+++ b/packages/client/src/client.test.ts
@@ -4,7 +4,7 @@ import { mkdtemp, rm, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { getUserConfigDir, saveUserConfig } from "@nakama/core";
-import { NakamaClient } from "./index";
+import { NakamaAuthExpiredError, NakamaClient } from "./index";
 
 test("chat stream request includes cookie CSRF protection", async () => {
   const originalDocument = (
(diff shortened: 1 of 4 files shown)


packages/client/src/client.ts
```diff
@@ -2441,6 +2445,18 @@ export class NakamaClient {
...
           return this.request(path, init, true);
         }
+
+        throw new NakamaAuthExpiredError(
+          await readApiErrorMessage(response),
+          path
+        );
+      }
+
+      if (response.status === 401 && retried) {
```
ahmadrosid, line 2455: These two throws are the same outcome. You can fold them and keep rotate → generic API error:

```ts
if (response.status === 401 && this.authToken && (retried || path !== "/v1/auth/local-token/rotate")) {
  if (!retried) {
    const freshToken = await loadLocalAuthToken();
    if (freshToken && freshToken !== this.authToken) {
      this.authToken = freshToken;
      return this.request(path, init, true);
    }
  }
  throw new NakamaAuthExpiredError(
    await readApiErrorMessage(response),
    path
  );
}
```

Same shape in `fetchRaw`.

packages/client/src/client.test.ts
```diff
@@ -355,6 +355,78 @@ test("non-browser clients reload the local auth token once after a 401", async (
...
+      );
+      expect((error as NakamaAuthExpiredError).status).toBe(401);
+    }
+  } finally {
+    delete process.env.NAKAMA_CONFIG_DIR;
+    await rm(configDir, { force: true, recursive: true });
+  }
+});
+
+test("non-browser clients throw NakamaAuthExpiredError when token reload still 401s", async () => {
```
ahmadrosid, line 404: This doesn’t actually exercise “reload then still 401” (`retried`) — missing file and unchanged token both fail the same `freshToken && freshToken !== this.authToken` check and throw before retry. One of these two tests is enough; keep the unchanged-token one (or switch it to `rejects.toBeInstanceOf` and drop the manual try/catch + `.name` assert).

ahmadrosid (commented):
Core change looks right — subclassing `NakamaApiError` is the clean hook. Two small tightenings below.

PR #806, opened 2026-09-01
perf(db): cache prepared statements for listSkillSuggestions

### LINKED ISSUE #602: [db] listSkillSuggestions uses db.query (no caching) on hot path
## Bug
`WHERE ${conditions.join(" AND ")}` is interpolated into the SQL string. Values are still bound via `...all(...params)` so this is NOT SQL injection — but `db.query` re-prepares the statement on every call (4 permutations × every list query). The rest of the adapter uses `db.prepare` and caches.

## Files
- `packages/db/src/adapters/sqlite.ts:2467-2495`

## Fix
Build a tiny `Map<key, Statement>` cache keyed by which of `{sessionId, status, profileId}` were provided.

## Severity
Medium — perf.

akmalsyrf:
## Summary

`listSkillSuggestions` reuses prepared statements keyed by which filters are present, instead of `db.query()` re-preparing on every call.

**Before:** each list built SQL with string-joined conditions and prepared via `db.query` (hot path, up to 8 shape permutations).
**After:** `Map` cache of `db.prepare` statements keyed by `{sessionId,status,profileId}` presence; values stay bound.

Why safe:
1. SQL shapes and bind order match the previous dynamic builder
2. Filter permutation test covers org / status / profile / session / combined
3. No schema or API change

Only re-add / follow up if a new filter column is added without extending the cache key.

## Test plan
- [x] `bun test apps/server/src/services/skill-suggestion-service.test.ts` (8 pass)
- [x] `listSuggestions filter permutations reuse prepared statements` — all five filter shapes return the same row

Closes #602

### DIFF
--- a/packages/db/src/adapters/sqlite.ts
+++ b/packages/db/src/adapters/sqlite.ts
@@ -1612,6 +1612,44 @@ function createSqliteDatabaseAdapter(db: Database): DatabaseAdapter {
     SET status = 'applied', applied_at = ?
     WHERE org_id = ? AND id = ?
   `);
+  const listSkillSuggestionsStmtCache = new Map<
+    string,
+    ReturnType<Database["prepare"]>
+  >();
+  const getListSkillSuggestionsStmt = (filters: {
+    profileId: boolean;
+    sessionId: boolean;
+    status: boolean;
+  }) => {
+    const key = `${filters.sessionId ? "s" : "_"}${filters.status ? "t" : "_"}${filters.profileId ? "p" : "_"}`;
+    const cached = listSkillSuggestionsStmtCache.get(key);
+    if (cached) {
+      return cached;
+    }
+
+    const conditions = ["org_id = ?"];
+    if (filters.sessionId) {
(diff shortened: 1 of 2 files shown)


packages/db/src/adapters/sqlite.ts
```diff
@@ -1612,6 +1612,44 @@ function createSqliteDatabaseAdapter(db: Database): DatabaseAdapter {
     SET status = 'applied', applied_at = ?
     WHERE org_id = ? AND id = ?
   `);
+  const listSkillSuggestionsStmtCache = new Map<
```
ahmadrosid, line 1615: You could skip the Map + key helper and match the rest of this adapter with one `db.prepare` that treats missing filters as null, e.g. `AND (? IS NULL OR session_id = ?)` (same for status / profile_id). Bind `sessionId ?? null` twice (and friends). Same behavior, no permutation cache.

ahmadrosid (commented):
Core fix is solid — stops re-preparing on every list. One shape idea below that would drop the cache machinery.

PR #813, opened 2026-09-02
fix(web): allowlist Composio OAuth redirect hosts

### LINKED ISSUE #580: [frontend] use-composio.ts redirects via window.location.href without allow-list
## Bug
If `response.redirectUrl` is server-controlled (Composio OAuth), it's a known destination. But the value is set without an allow-list; any compromise of the OAuth response could redirect to a phishing page.

## Files
- `apps/web/src/hooks/use-composio.ts:81`

## Fix
Validate the URL host against a Composio allow-list before assigning.

## Severity
Medium — open-redirect-adjacent.

akmalsyrf:
## Summary

Connect Composio toolkit → browser only navigates to `https://*.composio.dev` OAuth URLs. Fixes #580.

**Before:** `window.location.href = response.redirectUrl` with no host check.
**After:** `isAllowedComposioRedirectUrl` gates https + `composio.dev` / `*.composio.dev`; else toast and stay.

Why safe:
1. Matches Composio connect hosts (`backend.composio.dev`, `dashboard.composio.dev`)
2. Rejects http, lookalike hosts (`composio.dev.evil.com`, `notcomposio.dev`), empty/garbage URLs
3. Failure mode is stay-on-page + toast, not a broken allow path for real links

Only reopen if Composio starts issuing connect links on a non-`composio.dev` host.

## Test plan
- [x] `bun test apps/web/src/hooks/use-composio.test.ts` (2 pass)
- [ ] Integrations → Connect on a toolkit — still lands on Composio OAuth

### DIFF
--- a/apps/web/src/hooks/use-composio.test.ts
+++ b/apps/web/src/hooks/use-composio.test.ts
@@ -24,6 +24,16 @@ describe("isAllowedComposioRedirectUrl", () => {
     expect(
       isAllowedComposioRedirectUrl("https://composio.dev.evil.com/phish")
     ).toBe(false);
+    expect(
+      isAllowedComposioRedirectUrl("https://notcomposio.dev/connect")
+    ).toBe(false);
+    expect(isAllowedComposioRedirectUrl("")).toBe(false);
     expect(isAllowedComposioRedirectUrl("not a url")).toBe(false);
+  });
+
+  test("allows documented dashboard host", () => {
+    expect(
+      isAllowedComposioRedirectUrl("https://dashboard.composio.dev/oauth")
+    ).toBe(true);
   });
 });
--- a/apps/web/src/hooks/use-composio.test.ts
+++ b/apps/web/src/hooks/use-composio.test.ts
@@ -1,0 +1,29 @@
+import { describe, expect, test } from "bun:test";
+import { isAllowedComposioRedirectUrl } from "./use-composio";
+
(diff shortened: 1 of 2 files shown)


apps/web/src/hooks/use-composio.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    expect(
+      isAllowedComposioRedirectUrl("https://composio.dev.evil.com/phish")
+    ).toBe(false);
+    expect(
+      isAllowedComposioRedirectUrl("https://notcomposio.dev/connect")
+    ).toBe(false);
+    expect(isAllowedComposioRedirectUrl("")).toBe(false);
+    expect(isAllowedComposioRedirectUrl("not a url")).toBe(false);
+  });
+
```
ahmadrosid: `dashboard.composio.dev` is already covered by the first test’s subdomain case (`backend.composio.dev` + `endsWith(".composio.dev")`). You can drop this third test — or tuck that URL into the first one if you want the docs host named — without losing confidence.

ahmadrosid (commented):
Core fix is right-sized — allowlist before `location.href` is the whole job. One small test trim below.

apps/web/src/hooks/use-composio.test.ts
akmalsyrf: Done — folded `dashboard.composio.dev` into the first test and dropped the third.

PR #814, opened 2026-09-02
fix(web): keep chat message actions visible on touch

### LINKED ISSUE #581: [frontend] chat-message-list.tsx action buttons visible only on hover/focus
## Bug
`opacity-0 transition-opacity group-focus-within:opacity-100 group-hover:opacity-100` — keyboard-only users can focus the action buttons because they're inside the bubble and `focus-within` reveals them. Touch users on mobile may miss them entirely.

## Files
- `apps/web/src/components/chat/chat-message-list.tsx:534-596`

## Fix
Reveal on `group-active` as well, or keep them at `opacity-60` baseline.

## Severity
Medium — accessibility / discoverability.

akmalsyrf:
## Summary

Chat message copy / retry / branch actions stay at least faintly visible on touch (no hover). Fixes #581.

**Before:** `opacity-0` until `group-hover` / `group-focus-within`.
**After:** `opacity-60` baseline + `group-active` / hover / focus-within → full opacity.

Why safe:
1. Same buttons and handlers; CSS-only
2. Desktop hover/focus still brighten to 100%
3. No layout shift (controls were already in the DOM)

Only tweak if designers want actions fully hidden on desktop again.

## Test plan
- [x] Diff review: single className change on `AssistantMessageActions` row
- [x] Parent turn wrapper already has `group` (hover/active/focus-within apply)
- [ ] Phone / touch: open a chat reply — copy/retry visible without long-press hover

### DIFF
--- a/apps/web/src/components/chat/chat-message-list.tsx
+++ b/apps/web/src/components/chat/chat-message-list.tsx
@@ -543,7 +543,7 @@ function AssistantMessageActions({
     : null;
 
   return (
-    <div className="flex items-center gap-1 pt-1 opacity-0 transition-opacity group-focus-within:opacity-100 group-hover:opacity-100">
+    <div className="flex items-center gap-1 pt-1 opacity-60 transition-opacity group-focus-within:opacity-100 group-hover:opacity-100 group-active:opacity-100">
       <button
         aria-label={copied ? "Copied" : "Copy response"}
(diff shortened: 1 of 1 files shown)


ahmadrosid (commented):
Looks lean — ship. One class swap, same handlers, no extra machinery.

PR #815, opened 2026-09-02
fix(web): await prior attachment panel onClose before swap

### LINKED ISSUE #574: [frontend] chat-attach-panel-context opens new panel without closing previous owner's onClose
## Bug
`show()` does call `current.onClose?.()` when swapping, but if the previous owner is a panel that registered `onClose` to abort a fetch, and the new panel renders before the abort completes, both panels can stream into the same DOM node for a tick.

## Files
- `apps/web/src/context/chat-attachment-panel-context.tsx:65-74`

## Fix
Await `current.onClose?.()` (return a Promise) before calling `setConfig(nextConfig)`.

## Severity
Medium — flicker, content leak.

akmalsyrf:
## Summary

Swapping the chat attachment panel waits for the previous owner’s `onClose` before mounting the next panel, and rapid swaps reuse one in-flight close. Fixes #574.

**Before:** `onClose?.()` then immediate `setConfig(next)` — abort/fetch cleanup could race; rapid A→B→C could fire A’s `onClose` twice.
**After:** `beginAttachmentPanelClose` + generation guard at `show()`; close button also awaits and cancels pending show applies.

Why safe:
1. Sync `onClose` still applies via microtask/`finally`
2. Same panel id reuses one close promise (no double fire)
3. Closing the panel bumps generation so a late swap cannot remount

Only revisit if a panel’s `onClose` hangs without resolving (would delay the next open).

## Test plan
- [x] `bun test apps/web/src/context/chat-attachment-panel-swap.test.ts` (2 pass)
- [ ] Chat: open artifact A, then quickly artifact B — no flash of A’s body in B

### DIFF
--- a/apps/web/src/context/chat-attachment-panel-context-shared.ts
+++ b/apps/web/src/context/chat-attachment-panel-context-shared.ts
@@ -15,6 +15,54 @@ export interface ChatAttachmentPanelConfig {
   typeLabel?: string | null;
 }
 
+/** In-flight prior-panel close so rapid show() swaps do not re-fire onClose. */
+export type AttachmentPanelCloseInFlight = {
+  id: string;
+  promise: Promise<void>;
+} | null;
+
+/**
+ * Starts (or reuses) `current.onClose`. Returns null when there is nothing to
+ * await.
+ */
+export function beginAttachmentPanelClose(
+  current: Pick<ChatAttachmentPanelConfig, "id" | "onClose"> | null,
+  inFlight: { current: AttachmentPanelCloseInFlight }
+): Promise<void> | null {
+  if (!current) {
+    return null;
+  }
+
+  if (inFlight.current?.id === current.id) {
+    return inFlight.current.promise;
+  }
+
+  if (!current.onClose) {
+    return null;
+  }
+
+  const id = current.id;
+  const promise = Promise.resolve(current.onClose()).finally(() => {
+    if (inFlight.current?.id === id) {
+      inFlight.current = null;
+    }
+  });
+  inFlight.current = { id, promise };
+  return promise;
+}
+
+/** Await prior panel close before mounting a different `nextId`, or null. */
+export function beginPriorAttachmentPanelClose(
+  current: Pick<ChatAttachmentPanelConfig, "id" | "onClose"> | null,
+  nextId: string,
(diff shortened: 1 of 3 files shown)


apps/web/src/context/chat-attachment-panel-context-shared.ts
```diff
@@ -8,13 +8,61 @@ export interface ChatAttachmentPanelConfig {
...
+    if (inFlight.current?.id === id) {
+      inFlight.current = null;
+    }
+  });
+  inFlight.current = { id, promise };
+  return promise;
+}
+
+/** Await prior panel close before mounting a different `nextId`, or null. */
+export function beginPriorAttachmentPanelClose(
```
ahmadrosid: `beginPriorAttachmentPanelClose` is only used from `show` — fine to drop it and do `current && current.id !== nextConfig.id ? beginAttachmentPanelClose(current, closeInFlightRef) : null` at the call site. Keeps one helper for the actual reuse logic.

apps/web/src/context/chat-attachment-panel-swap.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    expect(first).toBe(second);
+    expect(closeCount).toBe(1);
+
+    releaseClose();
+    await first;
+    expect(inFlight.current).toBeNull();
+  });
+});
+
+describe("beginAttachmentPanelClose", () => {
```
ahmadrosid: This reuse case is the same property as the rapid-swaps test above (one promise, `closeCount === 1`). If you keep only `beginAttachmentPanelClose`, one hang/reuse test plus the null cases is enough — you can drop this second describe.

ahmadrosid (commented):
Core await + generation guard looks right. Two small flatten opportunities below.

apps/web/src/context/chat-attachment-panel-context-shared.ts
akmalsyrf: Done — inlined the id check at `show()` and dropped `beginPriorAttachmentPanelClose`.

apps/web/src/context/chat-attachment-panel-swap.test.ts
akmalsyrf: Done — kept one reuse test on `beginAttachmentPanelClose`.

PR #816, opened 2026-09-02
fix(db): stop reading legacy users.user_context

### LINKED ISSUE #550: [server] users.user_context is read but never written (dead column drift)
## Bug
`getUserContextStmt` does `COALESCE(om.user_context, u.user_context)`. `setUserContextStmt` only updates `org_members.user_context`. The `users.user_context` column becomes a tombstone: any per-user fallback that path returned at install time is now permanently shadowed by the per-org row and is never refreshed.

## Files
- `packages/db/src/adapters/sqlite.ts:1196-1206`
- `packages/db/sql/schema.sql:249`

## Fix
Either (a) drop `users.user_context` and stop reading it, or (b) mirror writes into both columns. Document the chosen behavior next to the schema.

## Severity
Medium — type drift / dead column.

akmalsyrf:
## Summary

`getUserContext` reads only `org_members.user_context`; a one-shot migrate copies any leftover `users.user_context` into empty memberships first. Fixes #550.

**Before:** `COALESCE(om.user_context, u.user_context)` while writes only update `org_members` — dead fallback, and dropping it alone would lose pre-org USER.md.
**After:** `migrateLegacyUserContextToOrgMembers` then select from `org_members` alone; schema comment marks the users column as legacy.

Why safe:
1. Writes already target `org_members` only since the org-scoped move
2. Migrate fills NULL memberships from users (idempotent; does not overwrite org-scoped values)
3. Column kept in place — no DROP / migration pain for existing installs

Only drop the users column later if we want a hard schema cleanup.

## Test plan
- [x] `bun test packages/db/src/user-context.test.ts` (2 pass)
- [x] `bun test packages/db/src/migrate.test.ts` — includes legacy copy case (22 pass in org/migration suite run)
- [ ] System → user context edit/save still persists per org

### DIFF
--- a/packages/db/src/migrate.ts
+++ b/packages/db/src/migrate.ts
@@ -26,6 +26,7 @@ export function migrateDatabase(db: Database): void {
   atomic(migrateSkillsTables);
   atomic(migrateUsersTable);
   atomic(migrateOrgTables);
+  atomic(migrateLegacyUserContextToOrgMembers);
   atomic(migrateOrgMemoryProposalsTable);
   atomic(migrateSkillProposalsTable);
   atomic(migrateSkillSuggestionsTable);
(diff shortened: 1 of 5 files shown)


packages/db/src/user-context.test.ts
```diff
@@ -87,4 +91,106 @@ describe("user context storage", () => {
...
+      expect(await reopened.adapter.getUserContext("org_1", "user_1")).toBe(
+        "# Legacy user-level context"
+      );
+      reopened.close();
+    } finally {
+      rmSync(dir, { force: true, recursive: true });
+    }
+  });
+
+  test("does not prefer users.user_context over org_members", async () => {
```
ahmadrosid: This case already passed under the old `COALESCE(om.user_context, u.user_context)` whenever org_members was non-null, so it doesn’t prove the read path stopped using users.

The reopen dance above it is also covered by the new migrate.test case (copy into empty memberships). I’d drop both file-based tests here and keep that one — same confidence, a lot less setup.

packages/db/src/migrate.ts
```diff
@@ -422,6 +423,45 @@ function migrateOrgTables(db: Database): void {
...
+    UPDATE org_members
+    SET user_context = (
+      SELECT users.user_context
+      FROM users
+      WHERE users.id = org_members.user_id
+        AND users.user_context IS NOT NULL
+        AND TRIM(users.user_context) != ''
+    )
+    WHERE user_context IS NULL
+      AND EXISTS (
```
ahmadrosid: The `EXISTS` repeats the SET subquery. If the subquery finds no non-empty legacy value, SQLite assigns NULL and the row stays NULL anyway — `WHERE user_context IS NULL` is enough.

ahmadrosid (commented):
Core change looks right — migrate once, then read org_members only. Two places that could shrink without losing coverage.

packages/db/src/user-context.test.ts
akmalsyrf: Done — dropped both file-based tests; migrate.test covers the copy.

packages/db/src/migrate.ts
akmalsyrf: Done — `WHERE user_context IS NULL` only.

Issue #361, opened 2026-08-21
Harden bootstrap & edge endpoints (setup race, import size caps, avatar auth, origin validation)


vincentri:
# Summary

Four hardening fixes for the bootstrap window and edge endpoints.

## a) Setup is not transactional (TOCTOU orphan org)

`auth.ts:370-419` checks `countHumanUsers() > 0` then creates; `org-service.ts:403-450` inserts the org **before** the user (fixed id `user_admin`, `:433`). No transaction wraps the sequence (the only `BEGIN` in packages/db is `migrate.ts:871`). Two concurrent setups on a fresh instance: winner becomes admin, loser dies on PK conflict *after* its org insert committed → orphan memberless org. Takeover on initialized instances is properly blocked.

**Fix:** wrap count-check + creation in a transaction, or roll back the org insert on unique-conflict failure.

## b) Pre-first-admin import has no size limits (memory DoS)

`setup-import.ts:126,149` → `decodeArchiveRequestData` (`data-portability.ts:170-177`) base64-decodes uncapped; `readZip` (`:350-372`) `unzipSync`s everything into memory with no entry/total uncompressed limits; `Bun.serve` sets no `maxRequestBodySize`. The route is unauthenticated during the pre-first-admin window (`public-routes.ts:7-8`, gated by `assertSetupImportAllowed` at `:180-190` — gate ordering is correct).

**Fix:** cap request body size and zip entry/total uncompressed sizes.

## c) Avatar endpoint is unauthenticated with global lookup

`public-routes.ts:28` exempts `GET /v1/profiles/{id}/avatar`; handler `profiles.ts:872-878` → `profile-service.ts:583-593` does a global `getProfile(profileId)` with no org filter. Profile ids are name-derived slugs (`profile-service.ts:123-131,686-706`) — enumerable cross-org. The dashboard consumes avatars same-origin (`ProfileAvatar.tsx:81-88`), so the exemption appears unnecessary.

**Fix:** remove the public exemption (cookies flow same-origin anyway) or scope the lookup.

## d) Client-controlled origin trusted verbatim

[...]

PR #819, opened 2026-09-03
fix(setup): claim the first admin in one transaction

fajarhide:
## Summary

Setup claims the first admin inside one transaction, so two concurrent requests cannot leave an orphan organization.

**Before:** two `POST /v1/auth/setup` requests both pass `countHumanUsers() === 0`, both insert an organization, one creates `user_admin` and the other gets a 500 on the fixed id. Its organization stays committed with no members.
**After:** `bootstrapInitialSetup` re-checks for a human user and writes the org, the admin and the membership in one sqlite transaction. The loser writes nothing and gets a 409.

Why safe:
1. The claim check moves into the transaction, so the check and the writes cannot interleave.
2. A failure partway through rolls the org back. The new db test proves it: without the transaction the org survives the failed admin insert.
3. Existing behavior kept, setup still 409s once a human user exists and invalid setup still creates nothing.

## Test plan
- [x] `bun run test`, 11 workspaces, 0 fail (server 1107, db 98)
- [x] New concurrency test goes red on main (`[201, 500]`) and green here (`[201, 409]`)

Part of #361 (a).

### DIFF
--- a/apps/server/src/http/app.test.ts
+++ b/apps/server/src/http/app.test.ts
@@ -3,6 +3,7 @@ import { mkdtemp, rm } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join, resolve } from "node:path";
 import { loadLocalAuthToken, verifyLocalAuthToken } from "@nakama/core";
+import { LOCAL_CLIENT_USER_ID } from "@nakama/core/local-auth";
 import { createInMemoryDatabaseAdapter } from "@nakama/db";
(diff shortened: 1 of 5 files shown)


ahmadrosid (commented):
Looks lean — ship. Atomic claim in one sqlite transaction is the right shape; the shared stmt runners and the org-record split both earn their keep.

ahmadrosid (commented):
Looks lean — ship.

Issue #357, opened 2026-08-21
Add runtime request-body validation + body size limits


vincentri:
# Summary

HTTP request bodies are never schema-validated at runtime, and there is no request-body size limit anywhere.

# Evidence (validated against source)

- `readJson` (`apps/server/src/http/shared.ts:355-364`) is `request.json() as T` — a type assertion with only a SyntaxError→400 mapping. Zero schema checking.
- Zod schemas in route files are OpenAPI documentation stubs (e.g. `profiles.ts:86-147` uses `z.object({}).passthrough()`), registered via `registerPath` while handlers use plain `app.post(...)` + `readJson` — so @hono/zod-openapi's runtime validation never runs. `setup-import.ts` defines real schemas (`:27-37`) but its handlers bypass them (`:126`, `:149`).
- `app.openapi()` — the only registration path that validates at runtime — is used solely for five body-less GETs (`auth.ts:475,516`; `system.ts:171,188,192`).
- No `bodyLimit` middleware anywhere; `Bun.serve` sets no `maxRequestBodySize` (`index.ts:334-344`).
- Related: query params cast without validation — `skill-suggestions.ts:97`, `skill-proposals.ts:99-103`, `org-memory.ts:696-700`.

# Impact

Wrong-typed fields flow into services and SQLite bindings, surfacing as 500s or silent misbehavior instead of 400s. Multi-MB/Gb JSON bodies are parsed before any check. Safety currently lives in per-service discipline rather than at the boundary — individual sinks were audited clean, so this is defense-in-depth rather than an open injection path.

# Fix

Pick one approach and apply boundary-wide:

- migrate mutating routes to `app.openapi()` + real zod schemas, or
- extend `readJson` to accept a zod schema param and validate inline,

plus add global `bodyLimit` middleware and validate enum query params instead of casting.

## Verification notes

Spot-checked against source on 2026-08-21:

- Confirmed `readJson` at `apps/server/src/http/shared.ts:355-364` — `(await request.json()) as T` with only a SyntaxError→400 mapping. Exact match.
- Confirmed `profiles.ts:86-147` — every request/response schema is a `z.object({}).passthrough()` stub, registered via `registerPath` while handlers use plain `app.post(...)` + `readJson`.
- Confirmed `setup-import.ts` — real schemas at lines 27-37, bypassed by `readJson` calls at lines 126 and 149.
[...]

PR #820, opened 2026-09-03
fix(import): cap archive size on the setup import routes

fajarhide:
## Summary

The pre-first-admin import routes cap what they decode and unzip.

**Before:** `/v1/auth/setup/import/preview|restore` are unauthenticated until the first admin exists, and they base64 decoded and `unzipSync`ed whatever arrived. fflate sizes each output buffer from the entry's declared uncompressed size, so a few KB of archive could ask for gigabytes.
**After:** `decodeArchiveRequestData` rejects a payload over 100 MB with a 413 before it copies or decodes. `readZip` refuses an entry over 100 MB, or an archive over 500 MB uncompressed, from the unzip filter, which fflate runs before it allocates anything.

Why safe:
1. Both caps sit in the shared functions, so the authenticated platform import routes get them too.
2. Existing behavior kept, a valid archive still previews and restores and both routes still 409 after the first admin.
3. The request body limit itself belongs to #357, so it is not duplicated here.

## Test plan
- [x] `bun run test`, 11 workspaces, 0 fail (server 1109)
- [x] The three new cases go red with the caps raised to `Number.MAX_SAFE_INTEGER` and green as written

Part of #361 (b).

### DIFF
--- a/apps/server/src/http/routes/data-portability.ts
+++ b/apps/server/src/http/routes/data-portability.ts
@@ -95,6 +95,10 @@ export function registerDataPortabilityRoutes(
           content: { "application/json": { schema: errorSchema } },
           description: "Error",
         },
+        413: {
+          content: { "application/json": { schema: errorSchema } },
+          description: "Error",
+        },
         500: {
           content: { "application/json": { schema: errorSchema } },
           description: "Error",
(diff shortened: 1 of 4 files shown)


ahmadrosid (commented):
Looks lean — ship. Caps sit in the shared decode/unzip path, routes only hoist decode so 413 isn’t swallowed, and the zip test helper grows just enough for the declared-size cases.

PR #821, opened 2026-09-03
fix(origin): allowlist the client-supplied origin

fajarhide:
## Summary

A client-supplied origin has to be one we recognise before it goes into a generated URL.

**Before:** `resolveRequestClientOrigin` took `clientOrigin` from the body, or the `Origin` or `Referer` header, stripped a trailing slash and returned it. That value becomes the Composio OAuth `redirect_uri` and the artifact share link. A configured `webPublicUrl` won for the callback, but a mismatching origin was ignored silently rather than refused, and an install with none accepted any host.
**After:** the origin must be http or https, and must either match the configured `webPublicUrl` or, on an install with none, be loopback or the host the request arrived on. Anything else is a 400 on Composio connect, artifact share publish and status, session send and setup.

Why safe:
1. Hosts are compared without the scheme, so a TLS terminator that drops `X-Forwarded-Proto` does not reject its own users.
2. Loopback stays allowed when nothing is configured, which is what the split-port dev setup and first-time setup need.
3. The agent-tool hop carries an origin its own route already checked and has no request to compare against, so it passes through.

This takes the stricter of the two options #361(d) names. An install that serves the UI from a host other than its configured `webPublicUrl` now gets a 400 instead of a silently rewritten link.

## Test plan
- [x] `bun run test`, 11 workspaces, 0 fail (server 1108)
- [x] The four allowlist cases go red with the check stubbed out and green as written

Part of #361 (d).

### DIFF
--- a/apps/server/src/http/routes/artifact-shares.test.ts
+++ b/apps/server/src/http/routes/artifact-shares.test.ts
@@ -273,7 +273,7 @@ describe("artifact share routes", () => {
     );
   });
 
-  test("publish prefers clientOrigin over loopback request URL", async () => {
+  test("publish refuses a clientOrigin off the request host", async () => {
     const { app, databaseAdapter } = createApp();
     const session = await setupFreshInstallSession(app, databaseAdapter);
     const orgId = session.orgId!;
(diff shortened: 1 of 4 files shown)


ahmadrosid (commented):
Looks lean — ship.

PR #822, opened 2026-09-03
fix(core): pin web_fetch to the addresses it verified

### LINKED ISSUE #523: [security] web_fetch SSRF check doesn't pin DNS — DNS rebinding defeats it
## Bug
DNS resolution is checked once at the start of the request, but the actual fetch uses `fetch()` which may re-resolve and return a different address (DNS rebinding). Combined with redirect handling, a hostname that returns a public IP first and a private IP second can reach internal services. Each hop still re-resolves without IP pinning.

## Files
- `packages/core/src/tools/web-fetch.ts:163-202`

## Fix
Resolve once, pass the resulting IP to `fetch` as `lookup:` override, and re-check `assertPublicHostname` on every redirect.

## Severity
High — SSRF via DNS rebinding.

vianbas:
## Summary

`web_fetch` connects to the addresses it verified, so a name that resolves public during the SSRF check cannot resolve private for the connection. Fixes #523.

**Before:** the hostname was checked, then handed to `fetch`, which resolved it a second time.
**After:** `resolvePublicAddresses` returns every public record and each hop tries those in order, carrying the logical host as `Host` and, over https, as `tls.serverName`.

Why safe:
1. The deny set is untouched: same `isPrivateIp` checks, same messages, and all 33 existing tests pass unmodified
2. Keeping the whole list preserves the fallback `fetch` does today, which matters because most public names answer with an IPv6 record first, so pinning only the first record would break every fetch from an IPv4-only host
3. Certificate identity is still checked against the hostname rather than the address, through `tls.serverName`, so SNI-dependent hosts keep working

Only re-add / follow up if a name's records are all public but the working one is not reachable within the 30s budget, since the hops are tried in resolver order rather than in parallel.

Note for whoever reads the linked issue: the fix it proposes cannot work. Bun's `fetch` has no `lookup` option and ignores one silently, measured in the issue thread.

## Test plan
- [x] `bun test packages/core/src/tools/web-fetch.test.ts` (40 pass, 7 new) and `bun test packages/core/src` (748 pass)
- [x] Reverted the implementation with the new tests in place: 5 of the 7 fail, so they discriminate
- [x] Rebinding reproduced against `main`: with DNS answering public to the guard while the OS still maps the name to loopback, `main` returns 200 from a local service; this branch never reaches it, including when the attacker offers loopback as a second record
- [x] Real end-to-end versus `main`: `https://example.com/`, `http://github.com/` and `https://bun.sh/` return identical status, `finalUrl` and byte counts
- [ ] Fetch a public https URL from the tool playground and confirm the body still returns

### DIFF
--- a/packages/core/src/tools/web-fetch.test.ts
+++ b/packages/core/src/tools/web-fetch.test.ts
@@ -333,19 +333,6 @@ describe("web_fetch pins the address it verified", () => {
     ]);
   });
 
-  test("never connects to a private record in a mixed answer", async () => {
-    const inputs: string[] = [];
-    stubFetch(async (input) => {
-      inputs.push(String(input));
-      return htmlResponse("<p>ok</p>");
-    });
-
-    // github-pages.test answers with a public IPv4 and a unique-local IPv6.
-    await webFetchTool.run({ url: "https://github-pages.test/" }, CTX);
-
-    expect(inputs).toEqual(["https://185.199.111.153/"]);
-  });
-
   test("re-pins each redirect hop and still reports the logical url", async () => {
     const inputs: string[] = [];
     const hosts: string[] = [];
--- a/packages/core/src/tools/web-fetch.test.ts
+++ b/packages/core/src/tools/web-fetch.test.ts
@@ -13,6 +13,13 @@ mock.module("node:dns/promises", () => ({
         { address: "fd00:aa:bb:2250::b9c7:6f99" },
       ]);
     }
+    if (hostname === "dual.test") {
+      // Two public records, IPv6 first — the shape most real names have.
+      return Promise.resolve([
+        { address: "2606:4700:20::681a:58a" },
+        { address: "185.199.111.153" },
+      ]);
+    }
     if (hostname === "private-alias.test") {
       return Promise.resolve([{ address: "fd00:aa:bb:2250::0a00:0001" }]);
     }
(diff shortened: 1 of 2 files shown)


packages/core/src/tools/web-fetch.test.ts
```diff
@@ -255,6 +262,121 @@ describe("web_fetch SSRF guard", () => {
...
+    const out = await webFetchTool.run({ url: "https://dual.test/" }, CTX);
+
+    expect(out.status).toBe(200);
+    expect(inputs).toEqual([
+      "https://[2606:4700:20::681a:58a]/",
+      "https://185.199.111.153/",
+    ]);
+  });
+
+  test("never connects to a private record in a mixed answer", async () => {
```
ahmadrosid: `github-pages.test` is already public + ULA, and the first pin test already asserts a single connect to `185.199.111.153`. This doesn’t cover a different failure mode — fine to drop it.

ahmadrosid (commented):
Pinning to the verified addresses is the right shape. One small overlap in the new tests.

packages/core/src/tools/web-fetch.test.ts
vianbas: Dropped it, thanks. You are right that it covers no separate failure mode: both tests fetch `github-pages.test`, whose stubbed answer is the public IPv4 plus a ULA IPv6, and both assert a single connect to `185.199.111.153`, so the first one already proves the private record is never dialled.

I checked what the removal costs before making it, by mutating the implementation and counting which tests still fail. Reverting the pinning entirely now fails 4 instead of 5, and the one lost is exactly the test you flagged. Removing the private-record filter still fails 2, but neither pin test was catching that one; it is the existing `rejects localhost hostname` and `rejects hostnames with only private addresses` that do. Disabling the per-address fallback still fails 1. So the coverage is unchanged.

`bun test packages/core/src` is 747 pass, and knip and ultracite are clean.

PR #823, opened 2026-09-03
fix(web): confirm before deleting a task

### LINKED ISSUE #488: Task delete fires without confirmation (and pattern repeated across app)
## Bug

The "Delete" button in the task detail dialog calls `void onDelete()` immediately on click — no confirmation dialog, no two-step confirm, no `window.confirm`. A single mis-click (or a Cmd+W while focused on the button) deletes the task and its run history.

## Files

- `apps/web/src/components/tasks/TaskDetailDialog.tsx:275-283`

## Impact

- Data loss from accidental click
- Pattern is repeated across the app — see related issue for unguarded deletes in profiles/MCP/skills/tools/automations/allowed-users

## Fix

Wrap the delete in a confirmation step (the rest of the app uses `*DeleteDialog` components for this; reuse `profiles-dialogs.tsx`'s pattern).

## Severity

Critical — UX, data loss.

vianbas:
## Summary

Deleting a task takes two clicks, so one mis-click no longer removes it and its run history. Fixes #488.

**Before:** the footer's Delete called `onDelete` straight from the click.
**After:** Delete arms `confirmingDelete` in the existing form reducer and the footer swaps to Cancel plus Delete task.

Why safe:
1. Follows the inline step `SkillAssignPicker` already uses, rather than nesting a second `Dialog` inside one that is open, and adds no shared confirm component
2. Run agent and Save changes are hidden while the confirmation is up, so nothing else can fire mid-decision, and the existing `busy` already covers the delete so no new pending state was needed
3. A failed delete leaves the confirmation in place to retry or dismiss, and nothing settles state on a dialog that a successful delete has already unmounted

Only re-add / follow up if the confirmation should also name the task in the footer; the dialog header already shows the title, so I left the copy out.

A note on the test: `apps/web` has no component test harness, no `.test.tsx` and no DOM dependency, so the footer decision lives in `taskDetailFooterActions` and is tested directly, the way `chat-composer-actions.ts` already does for the composer's Stop and Submit buttons. What that leaves untested is the JSX click binding itself.

## Test plan
- [x] `bun test apps/web/src` (332 pass, 8 new) and `bunx tsc --noEmit` in `apps/web` clean
- [x] Every branch mutated: making the delete one-click again fails 3 tests, leaving Run and Save reachable while confirming fails 2, never showing the confirmation fails 2, and keeping the arming button visible fails 1
- [x] `bun x ultracite check` clean on both files; `bun run knip` clean
- [ ] Open a task, click Delete, confirm the task survives Cancel and disappears on Delete task

### DIFF
--- a/apps/web/src/components/tasks/TaskDetailDialog.test.ts
+++ b/apps/web/src/components/tasks/TaskDetailDialog.test.ts
@@ -2,7 +2,6 @@ import { describe, expect, test } from "bun:test";
 import type { StoredTask } from "@nakama/core/contract";
 import {
   createFormStateFromTask,
-  taskDetailFooterActions,
   taskDetailFormReducer,
 } from "./TaskDetailDialog";
 
@@ -15,10 +14,6 @@ const TASK = {
 } as StoredTask;
 
 describe("taskDetailFormReducer delete confirmation", () => {
-  test("starts without a pending delete", () => {
-    expect(createFormStateFromTask(TASK).confirmingDelete).toBe(false);
-  });
-
   test("askDelete arms the confirmation instead of deleting", () => {
     const next = taskDetailFormReducer(createFormStateFromTask(TASK), {
       type: "askDelete",
@@ -61,35 +56,5 @@ describe("taskDetailFormReducer delete confirmation", () => {
 
     expect(synced.confirmingDelete).toBe(false);
     expect(synced.title).toBe("Other");
-  });
-});
-
-describe("taskDetailFooterActions", () => {
-  test("offers Delete, Run and Save before anything is armed", () => {
-    expect(taskDetailFooterActions({ confirmingDelete: false })).toEqual({
-      showConfirmDelete: false,
-      showDelete: true,
-      showRunAndSave: true,
-    });
-  });
-
-  test("offers only the confirmation once Delete is armed", () => {
-    // The regression this guards: the delete used to run on the first click.
-    // Nothing else may be reachable while the decision is on screen, or a
-    // mis-click could run or save the task mid-confirmation.
-    expect(taskDetailFooterActions({ confirmingDelete: true })).toEqual({
-      showConfirmDelete: true,
-      showDelete: false,
-      showRunAndSave: false,
-    });
-  });
-
-  test("arming through the reducer is what flips the footer", () => {
-    const armed = taskDetailFormReducer(createFormStateFromTask(TASK), {
-      type: "askDelete",
-    });
-
-    expect(taskDetailFooterActions(armed).showRunAndSave).toBe(false);
-    expect(taskDetailFooterActions(armed).showConfirmDelete).toBe(true);
   });
 });
--- a/apps/web/src/components/tasks/TaskDetailDialog.test.ts
+++ b/apps/web/src/components/tasks/TaskDetailDialog.test.ts
@@ -1,0 +1,95 @@
+import { describe, expect, test } from "bun:test";
+import type { StoredTask } from "@nakama/core/contract";
+import {
+  createFormStateFromTask,
+  taskDetailFooterActions,
+  taskDetailFormReducer,
+} from "./TaskDetailDialog";
+
+const TASK = {
+  description: "nightly",
+  id: "task-1",
+  profileId: "profile-1",
+  prompt: "do the thing",
+  title: "Nightly report",
+} as StoredTask;
+
+describe("taskDetailFormReducer delete confirmation", () => {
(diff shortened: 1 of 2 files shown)


apps/web/src/components/tasks/TaskDetailDialog.tsx
```diff
@@ -62,7 +67,24 @@ function createFormStateFromTask(task: StoredTask): TaskDetailFormState {
...
+ * that decision is still on screen.
+ */
+export function taskDetailFooterActions(state: { confirmingDelete: boolean }): {
+  showConfirmDelete: boolean;
+  showDelete: boolean;
+  showRunAndSave: boolean;
+} {
+  return {
+    showConfirmDelete: state.confirmingDelete,
+    showDelete: !state.confirmingDelete,
```
ahmadrosid: `showDelete` and `showRunAndSave` are always the same bit. One `!confirmingDelete` in the footer (or a single flag) is enough — then the glue test that wires the reducer into the helper can go too.

apps/web/src/components/tasks/TaskDetailDialog.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    expect(taskDetailFooterActions({ confirmingDelete: true })).toEqual({
+      showConfirmDelete: true,
+      showDelete: false,
+      showRunAndSave: false,
+    });
+  });
+
+  test("arming through the reducer is what flips the footer", () => {
+    const armed = taskDetailFormReducer(createFormStateFromTask(TASK), {
+      type: "askDelete",
```
ahmadrosid: `askDelete` plus the armed footer case already prove this flip. Fine to drop.

apps/web/src/components/tasks/TaskDetailDialog.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  description: "nightly",
+  id: "task-1",
+  profileId: "profile-1",
+  prompt: "do the thing",
+  title: "Nightly report",
+} as StoredTask;
+
+describe("taskDetailFormReducer delete confirmation", () => {
+  test("starts without a pending delete", () => {
+    expect(createFormStateFromTask(TASK).confirmingDelete).toBe(false);
```
ahmadrosid: Initial `false` and the patch-keeps-armed case don’t add much — `createFormStateFromTask` is a literal, and patch is a plain spread. Keep ask / cancel / sync; those are the real branches.

ahmadrosid (commented):
Core confirm flow looks right. The test file can shed a few cases without losing the regression cover.

vianbas:
All three done, with one exception I want to put numbers behind rather than just assert.

**1. `showDelete` and `showRunAndSave` are the same bit.** They were, literally: both returned `!state.confirmingDelete`. The footer now branches on `form.confirmingDelete` directly and `taskDetailFooterActions` is gone, so the confirmation and the ordinary actions are two sides of one condition instead of three flags that can drift apart. Net -50 lines.

**2. The glue test can go.** Dropped. Mutation testing agrees it guarded nothing on its own: every mutation it caught (`askDelete` not arming, Run/Save still shown while armed, the confirmation inverted) was also caught by a test that stays.

**3a. The initial-`false` case.** Dropped. `switching to another task clears a pending delete` already fails if that literal flips, because `sync` returns `createFormStateFromTask`.

**3b. The patch-keeps-armed case: I kept this one.** I measured before deciding, and it is the only test that fails for two mutations:

| Mutation in `patch` | tests failing with it | after dropping the test |
|---|---|---|
| also sets `confirmingDelete: false` | 1 | **0** |
| ignores `action.values` entirely | 1 | **0** |

The second is why the test asserts `title` too: that assertion is its positive control, not padding. You are right that `patch` is a plain spread today, and this test is the thing that keeps it one. The regression it blocks is mild, since it disarms rather than deletes, so if you would still rather drop it, say so and I will.

**Where the file lands:** 4 tests, catching 6 of 7 mutations. The survivor is inverting the footer condition itself. Nothing here can catch that: `apps/web` has no component harness, and it shows the wrong buttons the moment the dialog opens. Before this change the equivalent mutation was caught inside the helper, but no test ever checked that the JSX consumed the helper correctly, so that coverage was of a proxy rather than of the footer.

Gate: `bun test apps/web/src` 328 pass 0 fail (332 before, minus the 4 removed tests), `tsc --noEmit` in `apps/web` exit 0, `ultracite check` clean on both files, `knip` exit 0. The branch is behind `main` at `32d67089`, so I merged that in locally to check: no conflict, and the same four gates pass on the merge result.

PR #825, opened 2026-09-03
feat(web-search): allow a custom search back-end for the web_search tool

### LINKED ISSUE #592: [agent/core] web_search flag suppression is asymmetric across providers
## Bug
`webSearch` is disabled for openrouter and for gemini when local tools are present. Intentional but undocumented — if the provider changes its decision, the model will silently lose web-search capability.

## Files
- `packages/agent/src/chat.ts:399-407`

## Fix
Document the gating in `chat-prompt.ts` so the model does not assume web search when only local tools are present.

## Severity
Medium — behavioral consistency.

### LINKED ISSUE #842: feat(web-search): allow a custom search back-end for the web_search tool
### Summary

Let an operator point `web_search` at a search API of their own (Exa, Firecrawl, or any JSON search endpoint) from **Settings → Web search**. Once a back-end is configured the tool stops being a provider-hosted stub and runs as an ordinary local tool, so it works on every provider.

### Why

On `main` (32d67089) `web_search` exists only as a stub that the provider is expected to answer:

- `packages/core/src/tools/web-search.ts:22-25`: `run()` throws *"web_search runs on the configured OpenAI or Anthropic provider and cannot be executed locally."*, and the tool description states it requires OpenAI or Anthropic.
- `packages/agent/src/chat.ts:403-411`: hosted search reaches the provider only when the provider is not OpenRouter, not Gemini with any local tool in the same turn, and the turn carries no attachment.

Two consequences today:

1. Anyone running Nakama on OpenRouter, Gemini, Ollama, or an OpenAI-compatible endpoint cannot give an agent web search at all. There is no way to bring your own search key.
2. Even on OpenAI or Anthropic, a profile with `web_search` assigned loses search silently on any attachment turn, and on Gemini whenever another tool shares the turn. Nothing tells the model that search was unavailable, so it answers from memory as if it had searched.

Point 2 is tracked separately as #592. This issue is about point 1: the missing capability.

### Proposed scope and non-goals

**Scope**

- Settings → Web search: choose Exa, Firecrawl, or a custom JSON endpoint, and store the endpoint and API key.
- When a back-end is configured, `web_search` resolves to a real local tool, so it composes with other local tools and with attachment turns on every provider.
- Normalise vendor responses to the shape the chat UI already renders for Exa MCP results, `{ provider, query, results: [{ title, url, snippet }] }`, so no front-end parser work is needed.
- Store the key the way the mailbox password is stored (`~/.nakama/config.ini`), return it masked, and keep the write route org-admin only.

**Non-goals**

- Replacing hosted search where it already works. With nothing configured, behaviour stays exactly as it is today.
- Crawling or scraping beyond what the search API itself returns.
- Per-profile back-ends. One per install is enough for a first pass.

- `packages/core/src/tools/web-search.ts`: the stub, and `partitionTools`, which currently splits on the tool name.
- `packages/core/src/contract.ts`: a `hosted` flag on `ToolDefinition` so partitioning splits on a property rather than a name, with the built-in stub as `hosted: true`.
- `packages/agent/src/chat.ts:396-411`: the gating that decides whether hosted search is offered.
- The profile tool-resolution path, so web, CLI, Telegram, WhatsApp, Discord and automations all resolve the same tool with no per-channel drift.
- Settings routes and RBAC, plus the web settings page.

### Acceptance criteria

- [ ] With no back-end configured, behaviour is unchanged: the hosted stub, the same provider gating
- [ ] With a back-end configured, `web_search` runs on OpenRouter, on Gemini beside local tools, and on attachment turns
- [ ] The same tool resolves for web, CLI, Telegram, WhatsApp, Discord and automations
- [ ] Exa and Firecrawl work against their documented request and response shapes, and a custom endpoint works with no key set
- [ ] Results render in the chat UI with no front-end parser change
- [ ] The API key is only ever returned masked, and the write route is org-admin only
- [ ] `bun test` and `bun run knip` stay clean

### Note on ordering

Filed after the fact. PR #825 already implements this and has been reviewed. I built it before filing and claiming an issue, which is not the order CONTRIBUTING.md asks for. Filing it now so the tracker reflects the work, and so the feature half is not left implicit behind #592.

yusupsupriyadi:
## Summary

Point `web_search` at Exa, Firecrawl, or any JSON search endpoint from **Settings → Web search**. On the turns where no search can run at all, the model is finally told so.

**Before:** `web_search` only ran as a provider-hosted tool. Never on OpenRouter, dropped on Gemini whenever local tools shared the turn, dropped on any attachment turn, and when it was dropped the model kept it in its instructions and answered from memory as though it had searched.
**After:** a configured back-end makes `web_search` an ordinary local tool (`hosted: false`) that works on every provider; where hosted search is still dropped, the named `hostedWebSearch` condition feeds a per-turn notice built in `chat-prompt.ts`.

Why safe:
1. `partitionTools` splits on the new `ToolDefinition.hosted` flag and the built-in stub is `hosted: true`, so the unconfigured path behaves exactly as it does today.
2. The tool swap happens once inside `resolveProfileStoredTools`, so web, CLI, Telegram, WhatsApp, Discord and automations all resolve the same tool; no per-channel drift.
3. The API key lives in `~/.nakama/config.ini` like the mailbox password, is only ever returned masked, and `PUT` is org-admin only (added to `settings-rbac.test.ts`).

Only revert if a vendor changes its response shape out from under the normaliser. A wrong key or endpoint surfaces on the first `web_search` call, not at save time.

## Configuration

| Provider | Endpoint (pre-filled) | Auth | Request body | Results read from |
|---|---|---|---|---|
| Exa | `https://api.exa.ai/search` | `x-api-key` | `{ query, numResults, contents.text }` | `results[]` |
| Firecrawl | `https://api.firecrawl.dev/v2/search` | `Authorization: Bearer` | `{ query, limit }` | `data.web[]` |
| Custom | operator-supplied | `Bearer` only when a key is set | `{ query, limit }` | `results` / `data.web` / `organic` / `items` |

Request and response shapes for the two hosted vendors come from their current official docs. The tool returns `{ provider, query, results: [{ title, url, snippet }] }`, the shape the chat UI already renders for Exa MCP results, so no front-end parser changes were needed. A key is required for Exa and Firecrawl and optional for a custom endpoint, since self-hosted engines often need none. Switching vendors clears the previous endpoint and key rather than inheriting them.

Fixes #842

## Second commit: Fixes #592

`fix(agent): tell the model when hosted web search is dropped` names the previously inline gating as `hostedWebSearch`, records at the call site why each provider is excluded (OpenRouter has no hosted-search path, Gemini rejects `googleSearch` grounding beside function declarations, no provider accepts hosted search beside attachments), and appends the notice the issue asked for. It composes with the first commit rather than overlapping it: once a search back-end is configured, `hasWebSearch` is `false`, the suppression no longer applies, and the notice correctly stays silent. There is a regression test for exactly that case.

Fixes #592

## Test plan
- [x] `bun test` packages/core: 758 pass (17 new: config round-trip, request/response normalisation, partitioning)
- [x] `bun test` packages/agent: 93 pass (4 new: OpenRouter and Gemini notice, `web_fetch` hint, silence when search is local)
- [x] `bun test` apps/server: 1112 pass; the 4 `composio-callback-url` / `composio-tool-bridge` failures also fail on a clean `main`
- [x] `bun test` apps/web: 324 pass, plus `bun run build` (tsc + vite)
- [x] `bunx biome check` on changed files and `bunx knip`: both clean, no new `tsc` errors
- [ ] Save an Exa key under Settings → Web search, ask a profile that has `web_search` assigned a current-events question, and confirm sources render while the model provider runs no search of its own

### DIFF
--- a/apps/web/src/components/settings/WebSearchSettingsCard.tsx
+++ b/apps/web/src/components/settings/WebSearchSettingsCard.tsx
@@ -146,11 +146,6 @@ export function WebSearchSettingsCard() {
       <div className="flex flex-wrap items-center justify-between gap-3">
         <div className="min-w-0 space-y-0.5">
           <p className="font-medium text-foreground text-sm">Web search</p>
-          <p className="text-muted-foreground text-xs [text-wrap:pretty]">
-            {provider
-              ? "Nakama calls this endpoint for web_search instead of the model provider."
-              : "Leave unset to use the model provider's own hosted web search."}
-          </p>
           {savedHint ? (
             <p
               className="text-emerald-700 text-xs dark:text-emerald-300"
@@ -258,13 +253,6 @@ export function WebSearchSettingsCard() {
               {saveMutation.isPending ? <Spinner className="size-4" /> : "Save"}
             </Button>
           </div>
-
-          {provider === "custom" ? (
-            <p className="text-muted-foreground text-xs [text-wrap:pretty]">
-              The endpoint receives POST {"{ query, limit }"} and may return
-              results under results, data.web, organic or items.
-            </p>
-          ) : null}
         </div>
       ) : null}
 
--- a/apps/web/src/components/settings/WebSearchSettingsCard.tsx
+++ b/apps/web/src/components/settings/WebSearchSettingsCard.tsx
@@ -1,0 +1,278 @@
+import type { WebSearchProvider } from "@nakama/core/contract";
+import { ViewIcon, ViewOffIcon } from "hugeicons-react";
+import { useEffect, useState } from "react";
+import { Button } from "@/components/ui/button";
+import { Input } from "@/components/ui/input";
+import {
+  InputGroup,
+  InputGroupAddon,
+  InputGroupButton,
+  InputGroupInput,
+} from "@/components/ui/input-group";
+import {
+  Select,
+  SelectContent,
+  SelectItem,
+  SelectTrigger,
+  SelectValue,
+} from "@/components/ui/select";
+import { Spinner } from "@/components/ui/spinner";
+import {
+  useSaveWebSearchSettings,
+  useWebSearchSettings,
+} from "@/hooks/use-web-search-settings";
+import { formatError } from "@/lib/client";
+
+const BUILT_IN_VALUE = "__web_search_builtin__";
+
+/**
+ * Mirrors WEB_SEARCH_PROVIDER_ENDPOINTS in packages/core/src/web-search-config.ts.
+ * Duplicated because that module reads the config file and cannot be bundled
+ * for the browser; the server re-derives the endpoint on save anyway.
+ */
+const PROVIDER_PRESETS: Array<{
+  endpoint: string;
+  label: string;
+  value: WebSearchProvider;
+}> = [
+  { endpoint: "https://api.exa.ai/search", label: "Exa", value: "exa" },
+  {
+    endpoint: "https://api.firecrawl.dev/v2/search",
+    label: "Firecrawl",
+    value: "firecrawl",
+  },
+  { endpoint: "", label: "Custom endpoint", value: "custom" },
+];
+
+function presetEndpoint(provider: WebSearchProvider): string {
+  return (
+    PROVIDER_PRESETS.find((preset) => preset.value === provider)?.endpoint ?? ""
+  );
+}
+
+export function WebSearchSettingsCard() {
+  const { data: settings } = useWebSearchSettings();
+  const saveMutation = useSaveWebSearchSettings();
+  const [provider, setProvider] = useState<WebSearchProvider | null>(null);
+  const [endpoint, setEndpoint] = useState("");
+  const [apiKey, setApiKey] = useState("");
+  const [showApiKey, setShowApiKey] = useState(false);
+  const [formError, setFormError] = useState<string | null>(null);
+  const [savedHint, setSavedHint] = useState<string | null>(null);
+
+  useEffect(() => {
+    if (!settings) {
+      return;
+    }
+
+    setProvider(settings.provider);
+    setEndpoint(settings.endpoint ?? "");
+    setApiKey("");
+  }, [settings]);
+
+  useEffect(() => {
+    if (!savedHint) {
+      return;
+    }
+
+    const timeout = window.setTimeout(() => setSavedHint(null), 2500);
+    return () => window.clearTimeout(timeout);
+  }, [savedHint]);
+
+  const savedProvider = settings?.provider ?? null;
+  const keyAlreadySaved =
+    savedProvider === provider && Boolean(settings?.apiKeyMasked);
+
+  function handleProviderChange(value: string | null) {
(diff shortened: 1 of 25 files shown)


packages/core/src/web-search-config.ts
```diff
@@ -0,0 +1,287 @@
...
+
+export function parseWebSearchProvider(
+  value: string | null | undefined
+): WebSearchProvider | null {
+  const trimmed = value?.trim().toLowerCase();
+
+  return isWebSearchProvider(trimmed) ? trimmed : null;
+}
+
+export function validateWebSearchMaxResults(value: number): number {
```
ahmadrosid: Nothing in the product sets `maxResults` — the settings card never touches it, and every real save path leaves the default 5. You could hardcode 5 for now and drop the min/max constants, this validator, and the update-request field until there's a control that needs them.

apps/web/src/components/settings/WebSearchSettingsCard.tsx
```diff
@@ -0,0 +1,278 @@
...
+      }
+    );
+  }
+
+  return (
+    <div className="space-y-3 px-4 py-3" id="web-search-settings">
+      <div className="flex flex-wrap items-center justify-between gap-3">
+        <div className="min-w-0 space-y-0.5">
+          <p className="font-medium text-foreground text-sm">Web search</p>
+          <p className="text-muted-foreground text-xs [text-wrap:pretty]">
```
ahmadrosid: Built-in / Exa / Firecrawl / Custom already say what the select does — these subtitle lines (and the custom POST-shape blurb below) are the helper copy the other settings tiles skip. Safe to drop; the docs already cover the endpoint contract.

ahmadrosid (commented):
Hosted→local swap and the unavailable-turn guidance look right. Two spots that can still get shorter:

packages/core/src/web-search-config.ts
yusupsupriyadi: Fixed in eb63146a. `maxResults` is gone from `WebSearchConfigFile`, the public settings shape, `UpdateWebSearchSettingsRequest`, the OpenAPI schema and the `[web_search]` ini section; the validator, `parseMaxResults` and the min/max constants went with it. The tool now caps at a module-level `MAX_RESULTS = 5` — Exa gets `numResults: 5`, Firecrawl and custom endpoints get `limit: 5`, and the normaliser stops at five. Tests updated to match.

apps/web/src/components/settings/WebSearchSettingsCard.tsx
yusupsupriyadi: Fixed in eb63146a. Both subtitle variants and the custom-endpoint POST blurb are removed; the tile is now title + select like its siblings, with the endpoint and API-key fields appearing below only while a provider is chosen. The endpoint contract stays in `builtin-tools.mdx`.

ahmadrosid:
LGTM THANKS 👍

PR #839, opened 2026-09-04
fix(web): keep the last picked chat model on a new chat

### LINKED ISSUE #522: [frontend] use-chat-page.ts:728 hard-codes channel 'web' in createSession
## Bug
`executeSend` reads `sessionChannel` from line 130 but `client.createSession("web", { … })` at line 728 hard-codes `"web"`. For non-web sessions, the closure creates a web session instead of a channel-correct one. `isReadonlySessionChannel` masks it for `sendMessage` but `handleTryAgainMessage` can still hit this path.

## Files
- `apps/web/src/pages/chat/use-chat-page.ts:677-849`

## Fix
Pass `sessionChannel` into `executeSend` and use it in `createSession`.

## Severity
High — channel-correctness bug in chat core.

### LINKED ISSUE #840: [frontend] chat model picker resets to the profile default on every new chat
### Summary

The chat model picker resets to the profile default on every new chat, discarding the model the user just picked.

`sessionModel` is state that both draft-entry paths clear to `null`:

- `apps/web/src/pages/chat/use-chat-page.ts:136` sets `const [sessionModel, setSessionModel] = useState<string | null>(null)`
- `apps/web/src/pages/chat/use-chat-page.ts:359` calls `setSessionModel(null)` in `enterDraftChat`
- `apps/web/src/pages/chat/use-chat-page.ts:643` calls `setSessionModel(null)` in the `?new=1` layout effect

With `null`, `currentModelSelection` falls through to the profile model (`use-chat-page.ts:221-227`), and `createSession` is called with no model (`use-chat-page.ts:758`). Nothing persists the choice, so it cannot survive a new chat, a profile switch, or a reload.

### Measured on

32d67089

### Reproduction

1. Open the web dashboard on a profile whose default model is A.
2. In the composer's model picker, select model B (any non-default model the profile can reach).
3. Send a message. The turn runs on B, as expected.
4. Start a new chat (sidebar "New chat", or `/chat?new=1`).
5. Look at the model picker.

### Expected vs actual

Picker label after step 5:

```
expected: B      (the model picked in step 2)
actual:   A      (the profile default)
```

`createSession` for the new chat is likewise sent without a `model`, so the turn runs on A too. The choice is not just visually lost.

Nothing stores the user's pick. `handleModelChange` (`use-chat-page.ts:276-320`) writes `sessionModel` and, for an existing session, PATCHes the session; for a draft chat it only sets state that the next draft entry clears.

Proposed fix: persist the explicit pick per profile in `localStorage` (alongside the existing `nakama:session:<profileId>` key) and seed `sessionModel` from it in `enterDraftChat` and the `?new=1` effect. Seeding `sessionModel` itself keeps `createSession`, the picker label and the thinking/vision capability checks on their current paths.

Two details worth deciding in review:

- Store only an explicit `handleModelChange` pick, not the model a resumed session reports back, so "last model" means what the user chose.
- A stored pick can outlive its provider; it should fall back to the profile default rather than resolve to `__unknown__`.

Per profile rather than global, since each profile has its own default and its own reachable models.

Prior art: ChatGPT, Gemini and Claude all carry the last picked model into the next chat rather than snapping back to a default. That is the habit users bring with them, and it is why the current behaviour reads as a bug rather than as a deliberate reset. The profile default is a starting point, not a preference that should win over an explicit choice the user just made.

I have this implemented and tested locally and would like to pick it up. Happy to adjust the approach if you would rather it lived on the profile or on the server instead of `localStorage`.

yusupsupriyadi:
## Summary

Pick a model, start a new chat → the picker stays on that model, per profile. Fixes #840.

**Before:** every new chat cleared `sessionModel`, so the selection fell back to `activeProfile.model` and the user's pick was gone.
**After:** an explicit pick is stored per profile (`nakama:last-model:<profileId>`) and seeded back into `sessionModel` at both draft-entry points.

Why safe:
1. Seeding `sessionModel` reuses the existing path, so `createSession`, the picker label and the thinking/vision checks follow unchanged
2. Only a `handleModelChange` pick is stored, and it rolls back with the visible state when the session update fails
3. `knownModelSelection` drops a pick whose provider or model is gone, so a stale value falls back to the profile default instead of `__unknown__`

Only revisit if the model catalog can be empty at draft entry for a reason other than "not loaded yet". The stored pick is deliberately kept in that window.

## Test plan
- [x] `bun test` in `apps/web` (330 pass)
- [x] `bun x tsc --noEmit` in `apps/web` (clean)
- [ ] Pick a non-default model → New chat → picker still shows it; then delete that provider → picker falls back to the profile default

Related, not fixed: #522, the hard-coded `"web"` channel in the same `createSession` call.

### DIFF
--- a/apps/web/src/lib/chat-history-route.test.ts
+++ b/apps/web/src/lib/chat-history-route.test.ts
@@ -4,12 +4,15 @@ import {
   buildNewChatPath,
   CHAT_DRAFT_STORAGE_PREFIX,
   chatProfileIdFromPath,
+  clearLastChatModel,
   consumeStoredChatDraft,
   isChatSessionPath,
   isProfilesPath,
+  lastChatModelStorageKey,
   parseChatRouteParams,
   pickKnownProfileId,
   readInitialDraftChatProfileId,
+  readLastChatModel,
   readRequestedDraftFromNewChatSearch,
   readRequestedDraftKeyFromNewChatSearch,
   readRequestedProfileFromNewChatSearch,
@@ -18,6 +21,7 @@ import {
   resolveHistoryProfileId,
   resolveProfilesPageProfileId,
   storeChatDraft,
+  writeLastChatModel,
   writeStoredActiveChatProfileId,
 } from "./chat-history";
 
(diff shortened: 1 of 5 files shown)


yusupsupriyadi:
Closing this. I opened it before filing and claiming an issue, which skips the workflow step in CONTRIBUTING.md. Filing the issue first; will reopen a PR once it is assigned.

yusupsupriyadi:
Reopened now that the issue exists: #840 is filed and this PR links it (`Fixes #840`), which is the workflow step my earlier comment was closing for. Nothing in the diff changed since then.

CI is green (test, knip, react-doctor, sync), and the branch still merges cleanly into `main`.

PR #847, opened 2026-09-05
fix(cli): bound pending chat messages

### LINKED ISSUE #538: [cli] Pending-message queue grows unbounded
## Bug
`const queue: PendingMessage[] = [];` has no max size. A user can hold Enter (or paste huge images) repeatedly while a long stream runs. Each pending message holds `images: ImageAttachment[]` with base64 payloads up to 5 MB. 10 pasted screenshots = ~50 MB queued forever.

## Files
- `apps/cli/src/chat.ts:151, 311-322`

## Fix
Cap `queue.length` to e.g. 20 messages or 50 MB total image bytes; reject overflow with an error line.

## Severity
High — memory leak.

mamenesia:
## Summary

Send follow-ups during a CLI response → at most 20 prepared messages wait, with a visible rejection when full.

**Before:** queued text and image payloads grew without a limit; simultaneous preparation could also hide messages in the serialized send queue.
**After:** one drain owns busy state synchronously, caps pending messages at 20, and releases ownership without stranding arrivals.

Why safe:
1. Admission rejects overflow before echoing it or updating the last accepted message; delayed `/paste` rechecks busy after clipboard IO.
2. Regression cases preserve FIFO, ESC cancellation, error recovery, and exactly-once dispatch at both finalization boundaries.

Follow up separately if a total byte budget or concurrent preprocessing limit is needed; this bounds prepared-message count, not total heap usage.

Closes #538

## Test plan
- [x] Linux Bun 1.3.14: `bun test apps/cli/src` (149 pass); `bun test apps/cli/src/chat-queue.test.ts --rerun-each 10` (60 pass)
- [x] Linux: `bun run --cwd apps/cli build`, targeted Ultracite check, and `bun run knip`; a deliberately delayed busy reset makes the boundary regression fail, and restoring the fix passes
- [ ] During a slow CLI response, fill the pending queue, submit one more follow-up, then press ESC; overflow is rejected and accepted messages continue in order

### DIFF
--- a/apps/cli/src/chat-queue.test.ts
+++ b/apps/cli/src/chat-queue.test.ts
@@ -1,0 +1,260 @@
+import { expect, spyOn, test } from "bun:test";
+import { NakamaClient } from "@nakama/client";
+import type { ProfileSummary } from "@nakama/core";
+import { runChat } from "./chat";
+import * as clipboard from "./clipboard-image";
+import * as imageInput from "./image-input";
+import type { PendingMessage } from "./message-queue";
+import * as profile from "./profile";
+import { TerminalInput } from "./terminal-input";
+import { TerminalRenderer } from "./terminal-renderer";
+
+test.each([
+  {
+    boundary: "none",
+    delayedParsing: false,
+    delayedPaste: false,
+    withImages: false,
+  },
+  {
+    boundary: "none",
+    delayedParsing: false,
+    delayedPaste: false,
+    withImages: true,
+  },
+  {
+    boundary: "none",
+    delayedParsing: true,
+    delayedPaste: false,
+    withImages: true,
+  },
+  {
+    boundary: "none",
+    delayedParsing: false,
+    delayedPaste: true,
+    withImages: false,
+  },
+  {
+    boundary: "endStream",
+    delayedParsing: false,
+    delayedPaste: false,
+    withImages: false,
+  },
+  {
+    boundary: "emptyQueue",
+    delayedParsing: false,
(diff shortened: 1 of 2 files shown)


apps/cli/src/chat-queue.test.ts
```diff
@@ -0,0 +1,260 @@
...
+import { TerminalInput } from "./terminal-input";
+import { TerminalRenderer } from "./terminal-renderer";
+
+test.each([
+  {
+    boundary: "none",
+    delayedParsing: false,
+    delayedPaste: false,
+    withImages: false,
+  },
```
ahmadrosid: delayedParsing, delayedPaste, endStream, and emptyQueue already run the full 20-cap → ESC → FIFO → after-drain path (and the image / paste races). The two plain `boundary: "none"` rows mostly restate that — you could drop them and keep the same confidence with a shorter file.

ahmadrosid (commented):
Core drain ownership + pending cap looks tight. One trim on the regression matrix.

apps/cli/src/chat-queue.test.ts
mamenesia: Trimmed in [7d8f3ef7](https://github.com/ahmadrosid/nakama/pull/847/commits/7d8f3ef721a669941571c4d8476184b49236cf52): removed the two plain `boundary: "none"` rows, keeping delayed parsing, delayed paste, end-stream, and empty-queue coverage. All 4 remaining queue tests and the targeted lint/format check pass. No production changes.

PR #853, opened 2026-09-05
fix(cli): cap retained transcript messages at 1000

### LINKED ISSUE #539: [cli] VirtualMessageList keeps every message forever
## Bug
`this.messages: VirtualMessage[] = []` is appended to forever; `wrappedCache` holds the wrapped lines for each. After thousands of turns, `offsets[]` and `wrappedCache` keep growing without bound. The only place that clears is `clear()` (called only in `terminal-layout.ts:151` on reset). Long-running CLI sessions leak memory.

## Files
- `apps/cli/src/virtual-message-list.ts:30-91`

## Fix
Cap retained messages (e.g. last 1000) or evict from `wrappedCache` when their message scrolls out of any plausible viewport.

## Severity
High — memory leak.

mamenesia:
## Summary

Long-running CLI sessions retain only the latest 1,000 completed display messages. Fixes #539.

**Before:** transcript messages, wrapped caches, and offsets grow until reset.
**After:** oldest completed messages are evicted, retained cache indices are rebased, and offsets are rebuilt.

Why safe:
1. Active messages are preserved; persisted conversations are unchanged.
2. The new first message is rewrapped to remove its obsolete leading gap.
3. A 10,000-message probe passed 315 retained-history comparisons across three widths, ending with 1,000 messages, 1,000 cache entries, and 1,001 offsets.

Follow up if unusually large messages require a byte-based limit; this change caps message count only.

## Test plan
- [x] `bun test apps/cli/src/virtual-message-list.test.ts apps/cli/src/cli-layout.test.ts` (28 pass); `bun run --cwd apps/cli build` and targeted Ultracite checks pass.
- [ ] Manually scroll older history during output, resize the terminal, then return to live output.

Full Windows CLI suite: `bun test apps/cli/src` reports 144 pass and 3 pre-existing path-formatting failures in `chat.test.ts` and `display-path.test.ts`.

Draft status: assignment requested on #539 and pending. Direct standards/spec review found no actionable issues. The requested external Oracle verdict has not been obtained; this PR remains draft pending that gate.

### DIFF
--- a/apps/cli/src/virtual-message-list.test.ts
+++ b/apps/cli/src/virtual-message-list.test.ts
@@ -1,0 +1,72 @@
+import { expect, test } from "bun:test";
+import { VirtualMessageList } from "./virtual-message-list";
+
+test("retains only the latest 1000 implicit messages after cached renders", () => {
+  const list = new VirtualMessageList();
+  for (let i = 0; i < 1100; i++) {
+    list.appendLine(`message ${i}`);
+    list.totalLines(80);
+  }
+  expect(list.messageCount).toBe(1000);
+  expect(list.totalLines(80)).toBe(1000);
+  expect(list.messageLines(0, 80)).toEqual([" message 100 "]);
+  expect(list.messageLines(999, 80)).toEqual([" message 1099 "]);
+  expect(list.messageAtLine(999, 80)).toEqual({ index: 999, lineOffset: 0 });
+});
+
+test("eviction preserves styled wrapping, gaps, navigation and the open message", () => {
+  const list = new VirtualMessageList();
+  const expected = new VirtualMessageList();
+  for (let i = 0; i < 1005; i++) {
+    const kind = i % 2 === 0 ? "user" : "assistant";
+    list.beginMessage(kind);
+    list.appendLine(`message ${i}\nsecond line`);
+    list.sealMessage();
+    list.totalLines(40);
+    if (i >= 5) {
+      expected.beginMessage(kind);
+      expected.appendLine(`message ${i}\nsecond line`);
+      expected.sealMessage();
+    }
+  }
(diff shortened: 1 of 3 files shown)


ahmadrosid (commented):
Core change looks lean — the splice + cache rebase in `invalidateOffsets` is the right size for the bug.

One trim in the golden test below.

apps/cli/src/virtual-message-list.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    messages.appendLine("still streaming");
+  }
+  expect(list.messageCount).toBe(1000);
+  for (const width of [40, 10, 80]) {
+    const total = expected.totalLines(width);
+    expect(list.totalLines(width)).toBe(total);
+    expect(list.getLines(0, total, width)).toEqual(
+      expected.getLines(0, total, width)
+    );
+    for (const line of [0, 1, 5, total - 1]) {
```
ahmadrosid: Full-range `getLines` equality already locks wrapping, gaps, and the offset table. The inner `messageAtLine` / `snapToMessage` matrix mostly re-reads those same offsets — you could keep the `getLines` check (one wrapping width is enough) plus the open-message seal asserts and drop this loop.

mamenesia: Trimmed in [f2578629](https://github.com/ahmadrosid/nakama/pull/853/commits/f2578629d9e3ddaadd565472ee1981fbaa457d58): kept the full-range `getLines` comparison at width 10 and the open-message seal assertions; removed the navigation matrix and extra comparison widths. No production changes.

All 28 targeted transcript/layout tests and the targeted lint check pass. The full CLI suite reports 144 pass and 3 unrelated Windows path-formatting failures.

PR #857, opened 2026-09-06
fix(web): recover from already-revoked artifact links

### LINKED ISSUE #573: [frontend] use-artifact-share-controls reads stale statusQuery.data?.id after refetch
## Bug
`handleRotateLink` / `handleRevoke` fall back to `storedShareIdRef.current` when `statusQuery.data?.id` is stale. A rapid revoke → rotate sequence can pass an already-revoked share id back to the server, returning 404 to the user with no retry path.

## Files
- `apps/web/src/components/chat/use-artifact-share-controls.ts:197, 243`

## Fix
Refetch share status before issuing rotate, or surface the 404 as a clear "already revoked" toast.

## Severity
Medium — UX.

mamenesia:
## Summary

Rotate an already-revoked artifact link without getting stuck or disturbing a replacement link.

**Before:** a cached share ID returned 404, leaving rotation stuck and local sharing state stale.
**After:** 404 refreshes share status, clears the obsolete local link, and closes rotation with an already-revoked message; reopening sharing uses the current state.

Why safe:
1. Only revoke 404 is recovered; authentication, permission, and server errors still fail.
2. Stale rotation never publishes or revokes a replacement share; normal successful rotation still creates a link.

Concurrent revoke/publish remains non-atomic; if publishing returns no new URL, the UI asks users to reopen sharing rather than claiming a new link was created.

Fixes #573.

## Test plan
- [x] `bun test apps/web` (326 pass); targeted Ultracite check and `bun run knip` pass.
- [x] Isolated API + Vite with `agent-browser`: stale rotation closes with an already-revoked message and sends zero publish requests; old token returns 404 and replacement remains 200. Reopen and rotate normally: a new URL is displayed and returns 200.
- [ ] Full web typecheck: blocked by 21 pre-existing errors, previously reproduced on unchanged `origin/main` with the same dependencies; no diagnostics in the corrected files.

### DIFF
--- a/apps/web/src/hooks/use-resource-mutations.test.ts
+++ b/apps/web/src/hooks/use-resource-mutations.test.ts
@@ -84,18 +84,15 @@ test.each([200, 404])(
   }
 );
 
-test.each([401, 403, 500])(
-  "revoke preserves HTTP %s failures rather than allowing rotation to continue",
-  async (status) => {
-    queryClient.setQueryData(queryKey, { active: true, id: variables.shareId });
-    const error = new NakamaApiError("Request failed", status);
-    revoke.mockRejectedValue(error);
+test("revoke preserves HTTP 403 failures rather than allowing rotation to continue", async () => {
+  queryClient.setQueryData(queryKey, { active: true, id: variables.shareId });
+  const error = new NakamaApiError("Request failed", 403);
+  revoke.mockRejectedValue(error);
 
-    const mutate = renderMutation();
-    await expect(mutate()).rejects.toBe(error);
-    expect(queryClient.getQueryState(queryKey)?.isInvalidated).toBe(false);
-  }
-);
+  const mutate = renderMutation();
+  await expect(mutate()).rejects.toBe(error);
+  expect(queryClient.getQueryState(queryKey)?.isInvalidated).toBe(false);
+});
 
 describe("artifact share controls with a stale share ID", () => {
   const unusedAuthAction = mock(async () => {});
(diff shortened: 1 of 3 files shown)


apps/web/src/hooks/use-resource-mutations.test.ts
```diff
@@ -0,0 +1,245 @@
...
+    await expect(mutate()).resolves.toEqual({
+      id: variables.shareId,
+      revoked: status === 200,
+    });
+    expect(revoke).toHaveBeenCalledWith(variables.profileId, variables.shareId);
+    expect(queryClient.getQueryState(queryKey)?.isInvalidated).toBe(true);
+  }
+);
+
+test.each([401, 403, 500])(
```
ahmadrosid: 401 / 403 / 500 all prove the same thing: non-404 still throws and skips invalidation. One status (403 is enough) keeps the same confidence without the matrix.

ahmadrosid (commented):
Core fix is tight — 404 soft-success in the mutation plus early return before publish is the right shape. One small test trim below.

apps/web/src/hooks/use-resource-mutations.test.ts
mamenesia: Trimmed in [f1c29f19](https://github.com/ahmadrosid/nakama/pull/857/commits/f1c29f19befd7a64d3b2a30497b267ff44b5e23c): replaced the 401/403/500 matrix with one 403 case, preserving the rejection and skipped-invalidation assertions. All 7 tests in the file and the targeted lint/format check pass. No production changes.

PR #858, opened 2026-09-06
fix(core): support data URLs with parameters and fallback mediatypes

### LINKED ISSUE #854: fix(core): data URLs with parameters (e.g. charset) fail regex parsing in message-content
### Summary

`parseDataUrl` and `parseDocumentDataUrl` in `@nakama/core` use strict regex `^data:([^;]+);base64,(.+)$` which fails when standard data URLs include additional parameters such as `;charset=utf-8` or omit explicit media types (e.g. `data:text/markdown;charset=utf-8;base64,...`).

### Classification & Labels
- `bug`
- `stage: now`
- `priority: high`

### Measured on

Commit `main` (`packages/core/src/message-content.ts`)

### Reproduction

Self-contained reproduction script:

```bash
bun -e '
import { parseDataUrl, parseDocumentDataUrl } from "./packages/core/src/message-content";
console.log(parseDataUrl("data:image/png;charset=utf-8;base64,AAAA"));
console.log(parseDocumentDataUrl("data:text/markdown;charset=utf-8;base64,IyBRQQ==", "test.md"));
'
```

Both calls return `null` instead of parsing the attachment.

### Expected vs actual

- **Expected:** Successfully extracts media type (e.g. `image/png`, `text/markdown`) and base64 data.
- **Actual:** Returns `null`, causing uploaded text/markdown documents and charset-annotated images to be dropped with an ingestion error.

Update regex in `packages/core/src/message-content.ts` to match optional parameters before `;base64,`:

```ts
const match = /^data:(?:([^;,]+))?(?:;[^,]*)*;base64,(.+)$/s.exec(dataUrl.trim());
```

I have a fix with 23 passing tests ready and would like to take this issue.

itw-code:
## Summary

Allow `parseDataUrl` and `parseDocumentDataUrl` to parse base64 data URLs containing MIME parameters (e.g. `;charset=utf-8`) or implicit mediatypes without returning null.

**Before:** Data URLs with parameters like `data:text/markdown;charset=utf-8;base64,...` failed regex and dropped attachments.
**After:** Regex accepts optional semicolon-delimited parameters before `;base64,` and falls back to normalized file extensions.

Why safe:
1. Pure regex relaxation matching standard RFC 2397 data URL syntax.
2. Unparameterized standard data URLs continue matching identically.
3. Fully covered by 23 passing tests in `message-content.test.ts`.

Only re-add / follow up if other exotic URI schemes need support.

## Test plan
- [x] `bun test packages/core/src/message-content.test.ts` (23 pass, 0 fail)
- [x] Manual: verified parameterized data URLs return extracted mediaType and base64 payload.

Closes #854

### DIFF
--- a/packages/core/src/message-content.test.ts
+++ b/packages/core/src/message-content.test.ts
@@ -6,6 +6,7 @@ import {
   getUserMessageText,
   normalizeUserContent,
   parseDataUrl,
+  parseDocumentDataUrl,
   stripImagesForCompaction,
   validateCombinedAttachmentCount,
   validateDocumentAttachments,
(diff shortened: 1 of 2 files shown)


ahmadrosid (commented):
Looks lean — ship. Regex + the two defaults cover the parameterized / missing-mediatype cases without extra machinery.

PR #863, opened 2026-09-06
fix(openrouter): serialize user multimodal content

### LINKED ISSUE #844: fix(openrouter): image and document parts are rejected by SDK input validation, so every image chat fails
### Summary

Any chat turn carrying an image attachment fails on every OpenRouter model, including vision-capable ones such as `google/gemini-2.5-flash`. The request never leaves the process: the OpenRouter SDK rejects it during outbound input validation.

`@openrouter/sdk` is a Speakeasy-generated SDK. Its **input** model is camelCase (`ChatContentImage.imageUrl`) and the SDK serializes that to the OpenAI wire shape (`image_url`) itself. `openAIMessageToSdkMessage` converts the `assistant` and `tool` branches to camelCase (`toolCalls`, `toolCallId`) but passes user messages through untouched:

apps/server/src/providers/openrouter/index.ts:138

```ts
  return message as ChatMessages;
```

The parts it passes through are built in snake_case by `toOpenAIChatUserContent`:

packages/core/src/message-content.ts:465-468

```ts
    (part) => ({
      image_url: { url: toDataUrl(part.mediaType, part.data) },
      type: "image_url",
    })
```

So `{ image_url: {...} }` reaches a schema that wants `{ imageUrl: {...} }`, and the whole turn is rejected.

Documents hit the same wall. `toOpenAIResponsesDocumentBlock` emits `{ file_data, filename, type: "input_file" }` (packages/core/src/document-content.ts:171-175), while the SDK expects `{ file: { fileData, filename }, type: "file" }`. OpenRouter declares native support for PDF/DOCX/CSV/plain text (packages/core/src/document-content.ts:85-90), so that path is reachable too.

The vision fallback does not mask this. `preprocessUserContent` only swaps images for text descriptions when the primary model reports **no** vision support (apps/server/src/services/agent-service.ts:3429, `if (primarySupportsVision !== false) return content;`). A vision-capable OpenRouter model therefore sends the image parts straight into the failing conversion, and configuring an "Image parsing model" in Settings has no effect on it.

### Measured on

ace19027 (`@openrouter/sdk` ^1.2.61)

### Reproduction

Self-contained, no API key and no network:

```bash
cd apps/server
bun -e '
import { ChatMessages$outboundSchema } from "@openrouter/sdk/models";
const built = { role: "user", content: [
  { text: "what is this?", type: "text" },
  { image_url: { url: "data:image/png;base64,AAAA" }, type: "image_url" },
] };
const r = ChatMessages$outboundSchema.safeParse(built);
console.log(r.success ? "accepted" : "rejected: " + r.error.issues[0].message);
'
```

`{ image_url: ... }` is rejected; the same object with `imageUrl` is accepted and serializes back out as `image_url`.

As a failing test in `apps/server/src/providers/openrouter/index.test.ts`, which fails on ace19027 and passes once the user branch is converted:

```ts
  test("keeps image parts in the wire format OpenRouter expects", async () => {
    const bodies: Array<{ messages?: Array<{ content?: unknown }> }> = [];

    const fetchMock = mock(
      async (input: RequestInfo | URL, init?: RequestInit) => {
        const request =
          input instanceof Request ? input : new Request(input, init);
        bodies.push((await request.json()) as (typeof bodies)[number]);

        return new Response(chatCompletionResponse("A cat."), {
          headers: { "Content-Type": "application/json" },
          status: 200,
        });
      }
    );

    const provider = createOpenRouterProvider({
      apiKey: "sk-or-v1-test",
      fetcher: fetchMock as typeof fetch,
      model: "google/gemini-2.5-flash",
    });

    const result = await provider.generateChat({
      messages: [
        {
          content: [
            { text: "What is this?", type: "text" },
            { data: "AAAA", mediaType: "image/png", type: "image" },
          ],
          role: "user",
        },
      ],
      system: "You are helpful.",
    });

    expect(result.content).toBe("A cat.");
    expect(bodies.at(-1)?.messages?.at(-1)?.content).toEqual([
      { text: "What is this?", type: "text" },
      { image_url: { url: "data:image/png;base64,AAAA" }, type: "image_url" },
    ]);
  });
```

End to end: attach any image in the web chat with an OpenRouter vision model selected and send it.

### Expected vs actual

Expected: the model answers about the image. A real call to `google/gemini-2.5-flash` with a 32x32 red PNG and the prompt "Reply with only the dominant color name of this image." returns:

```
Red
```

Actual on ace19027, from the server log (truncated; the union is reported against every content branch):

```
SDKValidationError: Input validation failed: [
  {
    "code": "invalid_union",
    "errors": [
      ...
          [
            {
              "expected": "object",
              "code": "invalid_type",
              "path": [
                "imageUrl"
              ],
              "message": "Invalid input: expected object, received undefined"
            }
          ],
      ...
    ],
    "path": [
      "chatRequest",
      "messages",
      1
    ],
    "message": "Invalid input"
  }
]
    at async withOpenRouterError (apps/server/src/providers/openrouter/index.ts:91:18)
```

The turn is marked failed in the UI. Retry is not a way out either: it refuses any prompt carrying images or documents with "Retry is available for text-only prompts." (apps/web/src/pages/chat/use-chat-page.ts:981, and :1014 for Try again), so an image turn that fails cannot be retried at all.

Convert user content parts to the SDK's input shape instead of passing them through, in `openAIMessageToSdkMessage`:

- `{ image_url, type: "image_url" }` to `{ imageUrl, type: "image_url" }`
- `{ file_data, filename, type: "input_file" }` to `{ file: { fileData, filename }, type: "file" }`
- text parts already match and need no change

Text-only turns are unaffected, since `message.content` is a string there and the existing pass-through stays correct.

There is no test covering image or document content on this provider today, which is why the regression went unnoticed; the test above closes that gap.

Related, not fixed: #435 (image-parsing error wording in Settings), same vision area.

I have a fix and a passing test ready and would like to take this issue.

puspoaditya:
## Summary
- convert user image and document parts to the shapes expected by `@openrouter/sdk`
- preserve text-only user messages
- add regression coverage for image and PDF document requests

Closes #844

## Validation
- `bun test apps/server/src/providers/openrouter/index.test.ts`
- `bun test apps/server/src` (1111 pass)
- `bun x ultracite check apps/server/src/providers/openrouter/index.ts apps/server/src/providers/openrouter/index.test.ts`
- `bun run --filter @nakama/server build`

### DIFF
--- a/apps/server/src/providers/openrouter/index.ts
+++ b/apps/server/src/providers/openrouter/index.ts
@@ -108,6 +108,14 @@ function toSdkTools(
   }));
 }
 
+function isImageUrl(value: unknown): value is { url: string } {
+  return (
+    typeof value === "object" &&
+    value !== null &&
+    typeof (value as { url?: unknown }).url === "string"
+  );
+}
+
 function toSdkUserContent(
   content: Extract<OpenAIMessage, { role: "user" }>["content"]
 ): string | ChatContentItems[] {
@@ -116,12 +124,7 @@ function toSdkUserContent(
   }
 
   return content.map((part): ChatContentItems => {
-    if (
-      part.type === "image_url" &&
-      typeof part.image_url === "object" &&
-      part.image_url !== null &&
-      typeof part.image_url.url === "string"
-    ) {
+    if (part.type === "image_url" && isImageUrl(part.image_url)) {
       return {
         imageUrl: { url: part.image_url.url },
         type: "image_url",
--- a/apps/server/src/providers/openrouter/index.ts
+++ b/apps/server/src/providers/openrouter/index.ts
@@ -14,6 +14,7 @@ import type {
 import type { Fetcher } from "@openrouter/sdk";
 import { HTTPClient, OpenRouter } from "@openrouter/sdk";
 import type {
+  ChatContentItems,
   ChatFunctionTool,
   ChatMessages,
   ChatRequest,
@@ -107,7 +108,50 @@ function toSdkTools(
   }));
 }
 
+function toSdkUserContent(
+  content: Extract<OpenAIMessage, { role: "user" }>["content"]
+): string | ChatContentItems[] {
+  if (typeof content === "string") {
+    return content;
+  }
+
+  return content.map((part): ChatContentItems => {
+    if (
+      part.type === "image_url" &&
+      typeof part.image_url === "object" &&
+      part.image_url !== null &&
+      typeof part.image_url.url === "string"
+    ) {
+      return {
+        imageUrl: { url: part.image_url.url },
+        type: "image_url",
+      };
+    }
+
+    if (part.type === "input_file" && typeof part.file_data === "string") {
+      return {
+        file: {
+          fileData: part.file_data,
+          ...(typeof part.filename === "string"
+            ? { filename: part.filename }
+            : {}),
+        },
+        type: "file",
+      };
+    }
+
+    return part as ChatContentItems;
+  });
+}
+
 function openAIMessageToSdkMessage(message: OpenAIMessage): ChatMessages {
+  if (message.role === "user") {
+    return {
+      content: toSdkUserContent(message.content),
+      role: "user",
+    };
+  }
+
   if (message.role === "assistant") {
     return {
       content: message.content,
--- a/apps/server/src/providers/openrouter/index.test.ts
+++ b/apps/server/src/providers/openrouter/index.test.ts
@@ -87,6 +87,110 @@ describe("createOpenRouterProvider", () => {
     expect(fetchMock).toHaveBeenCalledTimes(1);
   });
 
+  test("serializes user image parts with the OpenRouter SDK shape", async () => {
+    const fetchMock = mock(
+      async (input: RequestInfo | URL, init?: RequestInit) => {
+        const request =
+          input instanceof Request ? input : new Request(input, init);
+        const body = (await request.json()) as {
+          messages: Array<{ content: unknown; role: string }>;
+        };
+
+        expect(body.messages[1]).toEqual({
+          content: [
+            { text: "What is this?", type: "text" },
+            {
+              image_url: {
+                url: "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAAB",
+              },
+              type: "image_url",
+            },
+          ],
+          role: "user",
+        });
+
+        return new Response(chatCompletionResponse("A tiny image"), {
+          headers: { "Content-Type": "application/json" },
+          status: 200,
+        });
+      }
+    );
+    const provider = createOpenRouterProvider({
+      apiKey: "test-key",
+      fetcher: fetchMock as typeof fetch,
+    });
+
+    await provider.generateChat({
+      messages: [
+        {
+          content: [
+            { text: "What is this?", type: "text" },
+            {
+              data: "iVBORw0KGgoAAAANSUhEUgAAAAEAAAAB",
+              mediaType: "image/png",
+              type: "image",
+            },
+          ],
+          role: "user",
+        },
+      ],
+      system: "You are helpful.",
+    });
+  });
+
+  test("serializes user document parts with the OpenRouter SDK shape", async () => {
+    const fetchMock = mock(
+      async (input: RequestInfo | URL, init?: RequestInit) => {
+        const request =
+          input instanceof Request ? input : new Request(input, init);
+        const body = (await request.json()) as {
+          messages: Array<{ content: unknown; role: string }>;
+        };
+
+        expect(body.messages[1]).toEqual({
+          content: [
+            { text: "Summarize this report", type: "text" },
+            {
+              file: {
+                file_data: "data:application/pdf;base64,JVBERi0=",
+                filename: "report.pdf",
+              },
+              type: "file",
+            },
+          ],
+          role: "user",
+        });
+
+        return new Response(chatCompletionResponse("A short report"), {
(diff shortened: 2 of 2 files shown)


yusupsupriyadi (commented):
Thanks for taking this. I went through the change against the SDK models and it fixes #844 correctly, but I hit two new TypeScript errors in the added code, so I am commenting rather than approving yet.

**What I verified**

- The wire shapes match the SDK input models. `ChatContentImage` is `{ imageUrl: { url }, type: "image_url" }` and its outbound schema serializes back to `image_url`; `ChatContentFile` is `{ file: { fileData, filename }, type: "file" }` and serializes `fileData` to `file_data`. Both new tests assert on the parsed request body, so they check the real payload rather than the intermediate object.
- The tests genuinely catch the bug. Running this PR's `index.test.ts` against the `main` copy of `index.ts`: 7 pass, 2 fail with `SDKValidationError`. Against this PR's `index.ts`: 9 pass, 0 fail.
- Coverage of the path is complete. `toOpenAIChatUserContent` (packages/core/src/message-content.ts:460) only ever emits `text`, `image_url` and `input_file` parts, the last via `toOpenAIResponsesDocumentBlock` (packages/core/src/document-content.ts:173). All three are handled.
- Text-only turns are unaffected: `content` stays a string and is returned as is.
- Dropping `filename` when it is not a string is right, since `toOpenAIResponsesDocumentBlock` sets `filename: part.filename` unconditionally and `ChatContentFileFile.filename` is optional.

**Two new type errors**

`part` is `Record<string, unknown>`, so `part.image_url` is `unknown`. The `typeof ... === "object" && ... !== null` guard narrows it to `object`, and `object` has no `url` property:

```
apps/server/src/providers/openrouter/index.ts(123,29): error TS2339: Property 'url' does not exist on type 'object'.
apps/server/src/providers/openrouter/index.ts(126,41): error TS2339: Property 'url' does not exist on type 'object'.
```

These are new. Type-checking the `main` version of the same file with the same config does not report them; the three other tsc errors in this file, around `SendChatCompletionRequestResponse`, are pre-existing. CI does not run `tsc`, so this will not turn the build red, but it surfaces in an editor.

A type predicate keeps the runtime guard identical and compiles clean:

```ts
function isImageUrl(value: unknown): value is { url: string } {
  return (
    typeof value === "object" &&
    value !== null &&
    typeof (value as { url?: unknown }).url === "string"
  );
}
```

used as:

```ts
if (part.type === "image_url" && isImageUrl(part.image_url)) {
  return { imageUrl: { url: part.image_url.url }, type: "image_url" };
}
```

I checked that this version type-checks with no errors in the file.

**Minor, not blocking**

Taking only `url` drops `detail`, which `ChatContentImageImageUrl` supports. Nothing emits `detail` today, so it is not a bug, just worth knowing if an image-detail option is ever added.

Happy to approve once the type errors are addressed.

apps/server/src/providers/openrouter/index.ts
```diff
@@ -107,7 +108,50 @@ function toSdkTools(
...
+      part.type === "image_url" &&
+      typeof part.image_url === "object" &&
+      part.image_url !== null &&
+      typeof part.image_url.url === "string"
+    ) {
+      return {
+        imageUrl: { url: part.image_url.url },
+        type: "image_url",
+      };
+    }
```
yusupsupriyadi, line 132:
```suggestion
function isImageUrl(value: unknown): value is { url: string } {
  return (
    typeof value === "object" &&
    value !== null &&
    typeof (value as { url?: unknown }).url === "string"
  );
}

function toSdkUserContent(
  content: Extract<OpenAIMessage, { role: "user" }>["content"]
): string | ChatContentItems[] {
  if (typeof content === "string") {
    return content;
  }

  return content.map((part): ChatContentItems => {
    if (part.type === "image_url" && isImageUrl(part.image_url)) {
      return {
        imageUrl: { url: part.image_url.url },
        type: "image_url",
      };
    }
```

This clears the two `TS2339: Property 'url' does not exist on type 'object'` errors I mentioned above.

`part` is `Record<string, unknown>`, so `part.image_url` is `unknown`. Narrowing it with `typeof ... === "object" && ... !== null` lands on `object`, and `object` has no `url` property, so both `typeof part.image_url.url === "string"` and `url: part.image_url.url` fail to compile. A type predicate performs the same three runtime checks but carries `{ url: string }` into the branch, so the property access is typed.

I applied this exact suggestion to your branch and checked it:

- `tsc --noEmit`: the two TS2339 errors are gone. The three remaining errors in the file, around `SendChatCompletionRequestResponse` at the `parseChatResult` and stream call sites, are pre-existing on `main` and unrelated.
- `bun test` on your `index.test.ts`: 9 pass, 0 fail, 23 expect() calls.
- `ultracite check` on the patched file: clean, no fixes applied.

Runtime behaviour is unchanged, so both new tests keep passing.

yusupsupriyadi (commented):
Follow-up on my review above: here is the type fix as an applyable suggestion, so it is easy to take or discard.

apps/server/src/providers/openrouter/index.ts
puspoaditya: Implemented in e7500843. I added the `isImageUrl` type predicate and now use it before reading `part.image_url.url`; runtime behavior is unchanged.

Validated after the change:
- `bun test apps/server/src/providers/openrouter/index.test.ts` — 9 pass, 0 fail
- `ultracite check` on the provider and its test — clean
- focused TypeScript check — 0 image-url `TS2339` diagnostics

Thanks for the precise suggestion and verification.

yusupsupriyadi (approved):
LGTM from me

PR #864, opened 2026-09-06
fix(web): validate public artifact-share fetch origins

### LINKED ISSUE #576: [frontend] use-public-artifact-share fetches without auth using client.baseUrl
## Bug
Builds the URL as `${client.baseUrl}/v1/public/artifact-shares/...`. `client.baseUrl` is configured at runtime, not an env var read at build time. If the operator misconfigures `NAKAMA_PUBLIC_URL` to point at a private endpoint, the share fetch hits it.

## Files
- `apps/web/src/hooks/use-public-artifact-share.ts:27-30, 54-56`

## Fix
Validate `client.baseUrl` against a known allow-list (http/https only, no `localhost` in prod).

## Severity
Medium — config-driven SSRF-adjacent.

akmalsyrf:
## Summary

Public artifact-share fetches validate `client.baseUrl` before hitting `/v1/public/artifact-shares/...`. Fixes #576.

**Before:** any absolute `client.baseUrl` (including `file:` / prod `localhost`) was interpolated straight into the unauthenticated share fetch URL.
**After:** `buildPublicArtifactShareUrl` allows empty/relative same-origin, http(s) only, rejects localhost in production; `PublicArtifactSharePage` try/catch keeps download UI stable when build rejects.

Why safe:
1. Default web client still uses `baseUrl: ""` (relative) — happy path unchanged
2. Rejected origins throw into the existing “unavailable” share error; page catch → null download link (same as before)
3. One exported builder + module-local origin check — no wrapper export

Only re-add localhost in prod if a documented reverse-proxy setup requires it. Private LAN IPs are still allowed if misconfigured as an absolute http(s) origin.

## Test plan
- [x] `bun test apps/web/src/lib/public-artifact-share-url.test.ts` (5 pass)
- [x] CI `test` / `knip` / `sync` / `react-doctor` green on prior head; re-run on `83417fbf`
- [ ] Smoke: relative + https OK; `file:` / prod localhost → page download link null (no crash)

### DIFF
--- a/apps/web/src/lib/public-artifact-share-url.ts
+++ b/apps/web/src/lib/public-artifact-share-url.ts
@@ -49,3 +49,17 @@ export function buildPublicArtifactShareUrl(
   const suffix = query ? `?${query}` : "";
   return `${origin}${path}${suffix}`;
 }
+
+/** Same as buildPublicArtifactShareUrl, or null when the origin is rejected. */
+export function tryBuildPublicArtifactShareUrl(
+  baseUrl: string,
+  token: string,
+  query?: string,
+  options?: { isProd?: boolean }
+): string | null {
+  try {
+    return buildPublicArtifactShareUrl(baseUrl, token, query, options);
+  } catch {
+    return null;
+  }
+}
--- a/apps/web/src/lib/public-artifact-share-url.ts
+++ b/apps/web/src/lib/public-artifact-share-url.ts
@@ -1,0 +1,51 @@
+const LOCAL_HOSTS = new Set(["localhost", "127.0.0.1", "::1"]);
+
+function isLocalHostname(hostname: string): boolean {
+  const host = hostname.toLowerCase();
+  return LOCAL_HOSTS.has(host) || host.endsWith(".localhost");
+}
+
+/**
+ * Resolve a safe API origin for unauthenticated public artifact-share fetches.
+ * Empty / relative baseUrl stays same-origin. Absolute URLs must be http(s);
+ * localhost is rejected in production builds.
+ */
+export function resolvePublicArtifactShareOrigin(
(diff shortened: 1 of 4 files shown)


apps/web/src/lib/public-artifact-share-url.ts
```diff
@@ -0,0 +1,65 @@
...
+  options?: { isProd?: boolean }
+): string {
+  const origin = resolvePublicArtifactShareOrigin(baseUrl, options);
+  const path = `/v1/public/artifact-shares/${encodeURIComponent(token)}`;
+  const suffix = query ? `?${query}` : "";
+  return `${origin}${path}${suffix}`;
+}
+
+/** Same as buildPublicArtifactShareUrl, or null when the origin is rejected. */
+export function tryBuildPublicArtifactShareUrl(
```
ahmadrosid: `tryBuildPublicArtifactShareUrl` is only used once, on the share page, so the render path does not throw. Drop this wrapper and wrap that one call in try/catch.

Same with exporting `resolvePublicArtifactShareOrigin` — keep it private and drive the reject cases through `buildPublicArtifactShareUrl` in the tests. Then this file is one throwing builder plus an unexported origin check.

ahmadrosid (commented):
Origin check is the right fix. One extra export layer you can drop.

apps/web/src/lib/public-artifact-share-url.ts
akmalsyrf: Good call — dropped `tryBuildPublicArtifactShareUrl`; the share page catches instead. `resolvePublicArtifactShareOrigin` is module-local now, and the reject/allow cases go through `buildPublicArtifactShareUrl` in the tests.

PR #865, opened 2026-09-06
fix(web): mask notification webhook secrets by default

### LINKED ISSUE #521: [security] NotificationDestinationsCard renders webhook secret in plaintext
## Bug
`latestSecret.apiKey` is fetched, displayed in a `<code>` block, and built into a `curl` example. The card never masks it, never auto-clears from memory after a timeout, and the "copy" button copies the raw key. If the user walks away with the card open, the secret sits in the DOM.

## Files
- `apps/web/src/components/NotificationDestinationsCard.tsx:69, 78, 98, 156`

## Fix
Mask by default with a "Reveal" toggle; clear `latestSecret` from state after 60s.

## Severity
High — secret at rest in DOM.

akmalsyrf:
## Summary

Newly minted notification webhook API keys stay masked until Reveal, then clear from React state after 60s. Fixes #521.

**Before:** `latestSecret.apiKey` sat in plaintext in the details panel (and curl example) for as long as the card stayed mounted.
**After:** default mask (`••••` + last 4), Reveal/Hide toggle, Copy still uses the real key, inline 60s clear, and rotate remounts so Reveal resets.

Why safe:
1. Copy curl / Copy API key still paste the real secret for the operator who just created it
2. Webhook URL stays visible (not secret); only the key is masked
3. Timer resets when a new secret is minted (rotate/create)

Only lengthen the TTL if operators report keys vanishing before they finish wiring Telegram.

## Test plan
- [x] `bun test apps/web/src/lib/notification-destinations.test.ts` (5 pass)
- [x] Owner YAGNI: single curl string + replace, direct `<LatestSecret />`, inline `60_000`
- [ ] Create/rotate destination → masked key + curl; Reveal; Copy still real; clears ~60s

### DIFF
--- a/apps/web/src/components/NotificationDestinationsCard.tsx
+++ b/apps/web/src/components/NotificationDestinationsCard.tsx
@@ -554,7 +554,12 @@ function NotificationDestinationItem({
       ) : null}
 
       {latestSecret?.destination.id === destination.id ? (
-        <LatestSecret latestSecret={latestSecret} />
+        {latestSecret ? (
+          <LatestSecret
+            key={latestSecret.apiKey}
+            latestSecret={latestSecret}
+          />
+        ) : null}
       ) : null}
     </div>
   );
--- a/apps/web/src/components/NotificationDestinationsCard.tsx
+++ b/apps/web/src/components/NotificationDestinationsCard.tsx
@@ -25,7 +25,6 @@ import { formatError } from "@/lib/client";
 import {
   buildNotificationWebhookUrl,
   formatTelegramDestinationLabel,
-  LATEST_WEBHOOK_SECRET_TTL_MS,
   maskWebhookApiKey,
   parseTelegramTopicLink,
 } from "@/lib/notification-destinations";
@@ -88,18 +87,6 @@ function LatestSecret({
     `    "level": "info"`,
     `  }'`,
   ].join("\n");
-  const displayCurlExample = revealed
-    ? curlExample
-    : [
-        `curl -X POST '${webhookUrl}' \\`,
-        `  -H 'Content-Type: application/json' \\`,
-        `  -H 'X-API-Key: ${maskWebhookApiKey(apiKey)}' \\`,
-        `  -d '{`,
-        `    "title": "New notification",`,
-        `    "body": "Hello from Nakama",`,
-        `    "level": "info"`,
-        `  }'`,
-      ].join("\n");
 
   async function copyCurlExample() {
     try {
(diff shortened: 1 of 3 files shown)


apps/web/src/components/NotificationDestinationsCard.tsx
```diff
@@ -82,6 +88,18 @@ function LatestSecret({
     `    "level": "info"`,
     `  }'`,
   ].join("\n");
+  const displayCurlExample = revealed
```
ahmadrosid: Same curl template twice. Copy still needs the real key, but the <pre> can reuse one string with `displayApiKey` (or `curlExample.replace(apiKey, displayApiKey)`). Then drop `displayCurlExample`.

apps/web/src/components/NotificationDestinationsCard.tsx
```diff
@@ -511,7 +554,12 @@ function NotificationDestinationItem({
       ) : null}
 
       {latestSecret?.destination.id === destination.id ? (
-        <LatestSecret latestSecret={latestSecret} />
+        {latestSecret ? (
```
ahmadrosid: The outer `latestSecret?.destination.id === destination.id` already means the secret is set. Render `<LatestSecret key={latestSecret.apiKey} … />` directly and drop the inner ternary.

apps/web/src/lib/notification-destinations.test.ts
```diff
@@ -23,3 +25,17 @@ describe("formatTelegramDestinationLabel", () => {
...
+describe("maskWebhookApiKey", () => {
+  test("masks long keys and keeps a short suffix", () => {
+    expect(maskWebhookApiKey("nk_live_abcdefghij")).toBe("••••ghij");
+  });
+
+  test("fully masks short keys", () => {
+    expect(maskWebhookApiKey("abc")).toBe("••••");
+  });
+
+  test("exports a one-minute clear window", () => {
```
ahmadrosid: This only asserts the constant equals itself. The timer has one call site — inline `60_000` in the effect and drop the export plus this test. The two mask cases already cover the real behavior.

ahmadrosid (commented):
Mask + Reveal + TTL is the right shape. Three spots you can shrink without changing behavior.

apps/web/src/components/NotificationDestinationsCard.tsx
akmalsyrf: Good call — one curl string with the masked key now, dropped the inner LatestSecret ternary, and inlined the 60s TTL (no exported constant or self-test).

PR #866, opened 2026-09-06
fix(platform): gate per-message worker logs behind NAKAMA_CH_DEBUG

### LINKED ISSUE #566: [platform] console.log structural info on by default in all workers
## Bug
Prints structural info every message. Combined with secret-leak logging in H9, this amplifies that problem.

## Files
- `apps/platform/{telegram,discord,whatsapp}/src/chat-handler.ts`, `socket.ts`, `bot.ts`

## Fix
Gate behind `NAKAMA_CH_DEBUG` or use a leveled logger.

## Severity
Medium — log hygiene.

akmalsyrf:
## Summary

Telegram / Discord / WhatsApp workers keep per-message structural `console.log` quiet unless `NAKAMA_CH_DEBUG=1`. Fixes #566.

**Before:** every inbound/skip/handle path printed structural lines (message ids, sizes, sometimes channel context) by default.
**After:** those paths no-op without the env flag; Discord inbound formatter always includes ids when the bot gate is on; Telegram ignore-log inlines the env check (no one-callsite helper).

Why safe:
1. Quiet vs loud stays at the call sites; Discord `bot.ts` / chat-handler gates unchanged
2. Operators can still turn verbose channel logs on with `NAKAMA_CH_DEBUG=1`
3. Lifecycle logs and `console.error` paths stay for ops

Only restore default per-message logs if a support playbook depends on them in production.

## Test plan
- [x] `bun test` discord `channel-log` + discord/telegram/whatsapp `chat-handler` + whatsapp `channel-log` (151 pass)
- [ ] Quiet without `NAKAMA_CH_DEBUG`; logs return with `=1` on one Discord inbound + one Telegram group ignore


apps/platform/discord/src/bot.ts
```diff
@@ -50,7 +50,9 @@ export async function createBot(
   });
 
   client.on(Events.MessageCreate, async (message: Message) => {
-    console.log(formatDiscordInboundMessageLog(message));
+    if (isChannelDebugEnabled()) {
```
ahmadrosid, line 53: This `if` already keeps inbound logs quiet, so `formatDiscordInboundMessageLog` still branching on the same flag (omit ids when off) never runs in production. Always include `authorId` / `channelId` in the formatter and drop the default-omit test — quiet vs loud lives here.

apps/platform/telegram/src/chat-handler.ts
```diff
@@ -128,15 +129,16 @@ export function createChatHandler(deps: ChatHandlerDeps) {
...
       ];
-      if (process.env.NAKAMA_CH_DEBUG === "1") {
-        parts.splice(
-          3,
-          0,
-          `botId=${botInfo?.id ?? "unknown"}`,
-          `chatId=${chatId}`,
-          `userId=${userId}`
-        );
+      if (!isChannelDebugEnabled()) {
```
ahmadrosid: Debug-off returns before any log, so the `parts.splice` below is leftover from the old always-log path. Check the flag first and put `botId` / `chatId` / `userId` in the array. After that, `isChannelDebugEnabled` is only used here — inline `process.env.NAKAMA_CH_DEBUG === "1"` like this file already did and skip the new `channel-log.ts` + clone test.

ahmadrosid (commented):
Gating the per-message logs is the right shape. Two leftover bits from the old always-log / maybe-ids path.

apps/platform/discord/src/bot.ts
akmalsyrf: Good call — Discord formatter always includes the ids now; quiet stays at the if. Telegram checks NAKAMA_CH_DEBUG up front, drops the splice leftover, and inlines the env check — no more channel-log helper.

PR #867, opened 2026-09-06
fix(web): make successful turn stamps monotonic

### LINKED ISSUE #572: [frontend] saveDraft closure race after profile switch
## Bug
`setLastSuccessfulTurnAt(Date.now())` uses `Date.now()` for an ID-equivalent timestamp; combined with the closure-based queue drain at `:839-848`, two concurrent `executeSend` invocations on the same session can both set the same timestamp if they finish in the same ms.

## Files
- `apps/web/src/pages/chat/use-chat-page.ts:786-789`

## Fix
Bump a per-turn counter rather than `Date.now()`, or use `performance.now()` with ms precision + a counter.

## Severity
Medium — state race.

akmalsyrf:
## Summary

`lastSuccessfulTurnAt` stays a wall-clock poll window start, but never collides when two turns finish in the same ms. Fixes #572.

**Before:** `setLastSuccessfulTurnAt(Date.now())` could assign the same stamp twice in one millisecond.
**After:** `nextSuccessfulTurnAt(previous)` returns `now`, or `previous + 1` when `now <= previous`, used from both send + reconnect success paths.

Why safe:
1. Post-turn skill-review polling still uses a real timestamp + 45s window
2. Pure helper with unit coverage; no storage / API shape change
3. Same setter site count as before (reconnect + successful send)

Only switch to a pure counter if polling stops depending on wall clock.

## Test plan
- [x] `bun test apps/web/src/pages/chat/chat-page.shared.test.ts` (12 pass)
- [x] CI `test` / `knip` / `sync` / `react-doctor` green on `99df3dd6`
- [x] Monotonic across same-ms + clock-skew; wired on send + reconnect success paths

### DIFF
--- a/apps/web/src/pages/chat/chat-page.shared.test.ts
+++ b/apps/web/src/pages/chat/chat-page.shared.test.ts
@@ -190,4 +190,8 @@ describe("nextSuccessfulTurnAt", () => {
   test("bumps by 1ms when two turns finish in the same millisecond", () => {
     expect(nextSuccessfulTurnAt(1000, 1000)).toBe(1001);
   });
+
+  test("stays monotonic when wall clock moves backwards", () => {
+    expect(nextSuccessfulTurnAt(1000, 999)).toBe(1001);
+  });
 });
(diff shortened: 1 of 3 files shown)


ahmadrosid (commented):
Looks lean — ship.

PR #868, opened 2026-09-06
feat(web): mobile-ready dashboard shell and installable PWA

### LINKED ISSUE #359: Security small-fixes batch (tool catalog leak, CSP hash, password trim, worker log clamp)
# Summary

Four small, independent security fixes found by the audit. Each is a few lines.

## a) Public tool catalog leaks custom-tool `modulePath`

`GET /v1/tools` is intentionally public (`public-routes.ts:14`, test-asserted in `public-routes.test.ts:24-28`), but `listTools` maps through `toToolDetail` (`tools.ts:308-310` → `profile-service.ts:809-816`), which spreads `handlerConfig` — including custom-tool `modulePath`, disclosing the `~/.nakama/tools/` layout unauthenticated.

**Fix:** use `toToolSummary` in `listTools`.

## b) CSP `script-src 'unsafe-inline'`

`app.ts:73` allows inline scripts because of the inline theme-bootstrap script in `apps/web/index.html:8-34` (Vite keeps it inline in builds). CSP therefore provides no XSS defense-in-depth while the app renders model output constantly.

**Fix:** compute a `sha256-` hash for that single script and drop `'unsafe-inline'`.

## c) Password trim inconsistency breaks login

Setup (`auth.ts:376`), accept-invite (`org-service.ts:632`), and change-password (`org-service.ts:700-701`) trim passwords before hashing; login compares raw input (`auth.ts:451-453`). A password set with leading/trailing whitespace can never be used to log in.

**Fix:** make trimming consistent (recommend trimming at login too, or not trimming anywhere).

## d) Worker-log clamp fails open on NaN

`workers.ts:157-161`: `parseInt("abc")` → NaN; `Math.max(1, NaN)` stays NaN; downstream `slice(-NaN)` equals `slice(0)` — the **entire log file** is returned, defeating the intended 1..2000 clamp.

**Fix:** guard with `Number.isNaN` and default (e.g. 200 lines).

## Verification notes

All citations spot-verified against the local checkout — every file:line is accurate as of 2026-08-21; no corrections needed.

- a) Confirmed: `/v1/tools` in `PUBLIC_ROUTES` at `apps/server/src/http/public-routes.ts:14`; test at `apps/server/src/http/public-routes.test.ts:24-28`; route handler at `apps/server/src/http/routes/tools.ts:308-310`; `profileService.listTools()` maps via `tools.map(toToolDetail)` (`apps/server/src/services/profile-service.ts:361-364`); `toToolDetail` includes `handlerConfig` (`profile-service.ts:813`). `modulePath` is indeed stored inside `handlerConfig` for JS tools (see `readJavascriptModulePath` / `validateJavascriptToolModule` in `profile-service.ts:879-895`), so the leak claim holds.
- b) Confirmed: CSP string at `apps/server/src/http/app.ts:73` contains `script-src 'self' 'unsafe-inline'`; inline theme-bootstrap `<script>` spans exactly `apps/web/index.html:8-34`. The "Vite keeps it inline in builds" premise is consistent with the source (non-module inline script in the HTML template) but was not verified by running a build.
- c) Confirmed: setup trims at `auth.ts:376`, accept-invite trims at `org-service.ts:632`, change-password trims both current and new password at `org-service.ts:700-701`, while login passes raw `body.password` to `verifyPassword` at `auth.ts:451-453`.
- d) Confirmed end-to-end: clamp at `apps/server/src/http/routes/workers.ts:157-161` lets NaN survive `Math.max`/`Math.min`; downstream `readLastLines` does `allLines.slice(-lineCount)` (`apps/server/src/services/worker-manager-service.ts:317`), and `slice(-NaN)` behaves as `slice(0)` — the whole file is returned.

## Acceptance criteria

- [ ] Unauthenticated `GET /v1/tools` returns tool summaries that contain no `handlerConfig` field — no `modulePath` or other handler-config keys appear anywhere in the response.
- [ ] Authenticated tool-detail endpoints still return full detail including `handlerConfig` (no regression for admin UI).
- [ ] Regression test: public `GET /v1/tools` response asserts absence of `modulePath`/`handlerConfig`.
- [ ] Served responses carry a CSP whose `script-src` no longer contains `'unsafe-inline'`.
- [ ] The web app loads with zero CSP violations: the theme-bootstrap inline script is allowed via its computed `sha256-` hash matching the built `index.html`.
- [ ] Theme bootstrap still applies the correct theme on first paint (no flash of wrong theme after the CSP change).
- [ ] A password containing leading/trailing whitespace set through any password-setting path (setup, accept-invite, change-password) can be used to log in — trimming policy is identical across all four flows.
- [ ] Regression test: create an account with a whitespace-padded password via setup, then log in with the same input and get a session.
- [ ] `GET /v1/workers/:name/logs?lines=abc` returns at most the default line count (200), never the entire log file.
- [ ] Non-numeric, zero, negative, and missing `lines` all resolve inside the intended 1..2000 clamp; values above 2000 are clamped to 2000.
- [ ] Regression test: non-numeric `lines` param response is capped at the default and does not equal the full log content.

## Related issues

- #355 — same credential-handling surface: #359(c) password-trim consistency belongs to the same auth-flow hygiene as revoking sessions on password change; both touch how passwords are normalized/validated across setup, invite acceptance, and change-password.
- #354 — same endpoint surface: both modify what `/v1/workers/:name/logs` returns (#354 redacts log content, #359(d) fixes the line-count clamp); independent fixes, but implementers should coordinate on the same route.

---
Found by internal security audit 2026-08-21. All four verdicts VALID / LOW after second-pass validation.

### LINKED ISSUE #841: feat(web): mobile-ready dashboard shell and installable PWA
### Summary

Make the web dashboard usable on a phone and installable as a PWA, so an operator running Nakama on a server behind a public domain can drive the whole console from the device they actually have with them.

Today the dashboard is switched off below 640px on purpose:

- `apps/web/src/components/Layout.tsx:65` wraps the entire shell in `max-sm:hidden`
- `apps/web/src/components/Layout.tsx:235-249` renders `NarrowViewportNotice` instead: *"This console needs a wider window … To chat with your agent from a phone, use the Telegram, WhatsApp or Discord bridge instead."*

There is also no manifest, no service worker and no install surface (`apps/web/public/` holds only the logo PNGs), so the app cannot be added to a home screen.

### Why

The bridges the notice points at are a chat surface, not the console. Over Telegram, WhatsApp or Discord you cannot choose which profile answers the turn, change the thinking/reasoning effort, switch the workspace folder, or reach settings, providers, skills, tools, MCP servers, automations or org memory. That is fine when the phone is a convenience and a laptop is nearby. It is not fine when Nakama is deployed to a server on a public domain specifically so it can be reached from anywhere. Then the phone is often the only device to hand, and the answer today is "open a laptop".

Installability matters for the same reason: a home-screen entry that opens standalone, with its own icon and a cached app shell, is what makes a self-hosted console feel reachable rather than a bookmark that shows a blank page on a flaky connection.

### Proposed scope and non-goals

Two phases, both shipping together in #868. The split below is still how the work is laid out, and the PR walks the whole thing in fourteen commits, but reviewing them in one place keeps `index.html` and the shell header from being touched twice.

**Phase 1: responsive shell**

- Drop `max-sm:hidden` and `NarrowViewportNotice`; below `sm` the profile rail and nav move into an off-canvas drawer opened from a header bar.
- Phone padding and `env(safe-area-inset-*)` for the notch and home indicator.
- 16px form controls below 640px, because iOS zooms in on anything smaller and never zooms back out.
- The attachment/artifact side panel goes fullscreen on phones; as a resizable column it clamps to roughly 195px there, which is unusable.

**Phase 2: PWA**

- `manifest.webmanifest` plus 192/512 `any` and maskable icons and an apple-touch icon.
- App-shell service worker precaching the built shell, with an explicit "new version ready" prompt rather than a silent `skipWaiting`.
- Server side: serve `.webmanifest` as `application/manifest+json`, and set `Cache-Control` per class (immutable for `/assets/`, `no-cache` for the shell, `sw.js` and the manifest).

**Non-goals for a first pass**

- Working offline. `AuthGuard` needs `/v1/auth/me`, so with no network the cached shell loads and redirects to `/login`. The shell is cached, the app is not usable.
- Push notifications, background sync, native wrappers.
- Redesigning dense settings tables beyond making them reachable and scrollable at phone width.

- `apps/web/src/components/Layout.tsx` for the shell, the drawer, and removing the notice
- `apps/web/index.html` for the manifest link, per-scheme `theme-color`, apple/mobile web-app metas, and `viewport-fit=cover`. The inlined theme script must stay byte-identical: its CSP hash is asserted in `apps/server/src/http/app.ts`.
- `apps/web/src/index.css`, `apps/web/src/pages/chat/*`, `apps/web/src/pages/LoginPage.tsx`
- `apps/web/public/` for the manifest and icons
- `apps/web/vite.config.ts` for `build.manifest` (the precache list), plus a build step for the worker
- `apps/server/src/static-web.ts` for the manifest MIME type and `Cache-Control`

A hand-written worker is worth considering over `vite-plugin-pwa`: no new dependency, and the routing/precache logic stays plain functions that can be unit-tested, which matters because `apps/web` has no DOM test harness.

### Acceptance criteria

- [ ] At 390x844, chat, profiles, settings and integrations render with no horizontal page scroll
- [ ] `NarrowViewportNotice` is gone and no part of the shell is hidden by `max-sm:hidden`
- [ ] Profile rail and navigation are reachable from a drawer on phones, and picking a profile dismisses it
- [ ] Desktop (1280) and tablet (768) layouts are unchanged, and sidebar collapse still works
- [ ] The manifest is served as `application/manifest+json` and the app installs with a standalone display mode and its own icon
- [ ] With the network off, a navigation is answered from the precached shell
- [ ] A new build surfaces an explicit update prompt; no silent `skipWaiting`
- [ ] The CSP-hash guard over `index.html` still passes (`apps/server/src/http/app.test.ts`)
- [ ] `bun test apps/web` and `bun run knip` stay clean

Both phases are implemented and up as #868, rebased onto 6f072a63 and re-verified there (web build including `sw.js`, 354 web tests, 49 server tests covering the manifest MIME, the Cache-Control classes and the CSP-hash guard, and a clean `bun run knip`). Driven end to end at 390x844 Chromium, plus the install, offline-shell and update paths. Not tested on real iOS or Android hardware yet. Happy to reshape the split, the breakpoint, or the worker approach if you would rather it went another way.

yusupsupriyadi:
## Summary

Open the Nakama console on a phone, install it to the home screen, and drive the whole thing from there.

**Before:** below 640px the shell was `max-sm:hidden` and `NarrowViewportNotice` sent you to the Telegram, WhatsApp or Discord bridge; no manifest, no service worker, no install surface.
**After:** the profile rail and nav open in an off-canvas drawer from the header (`MobileNavDrawer` + `AppSidebar`), and the built shell installs standalone and boots from a precached `sw.js`.

Why safe:
1. Desktop and tablet render the same tree as before; every phone-only path is gated behind `sm:` classes or one `matchMedia` hook
2. The worker precaches exactly what `precacheUrls` reads out of the build manifest, and a new build waits for an explicit prompt instead of `skipWaiting`
3. The CSP-hash guard over `index.html` still passes, so the inlined theme script is byte-identical

Only re-add a narrow-viewport notice if a dense settings table turns out to be unusable at 390px rather than merely tight.

## Test plan
- [x] `bun run --filter @nakama/web build` (tsc, vite, then `sw.js` precaching 5 files)
- [x] `bun test` in `apps/web` (354 pass)
- [x] `bun test src/static-web.test.ts src/http/app.test.ts` in `apps/server` (49 pass, CSP-hash guard included)
- [x] `bun run knip` (clean)
- [x] At 390x844 walk chat, profiles and settings, then install to the home screen and confirm it opens standalone

Fixes #841. Both phases described there land here rather than in two PRs.

Related, not fixed: #359, which added the CSP-hash guard over `index.html` that this change has to keep passing.


apps/web/src/context/chat-attachment-panel-context.tsx
```diff
@@ -125,20 +128,21 @@ export function ChatAttachmentPanelProvider({
     setConfig(null);
   }, []);
 
+  const fullscreen = (config?.fullscreen ?? false) || isPhone;
```
ahmadrosid: The rest of this PR already drives phone layout with `max-sm:`. Forcing fullscreen through `usePhoneViewport` is the one place that reimplements that cutoff in JS.

Collapse the chat column when the panel is open with the same `max-sm:` classes, and hide the resize handle the same way. Then `use-media-query.ts` can go — nothing else calls it.

apps/web/src/components/MobileNavDrawer.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+      <DialogPrimitive.Portal>
+        <DialogOverlay className="bg-black/40" />
+        <DialogPrimitive.Popup
+          className="data-open:slide-in-from-left data-closed:slide-out-to-left fixed inset-y-0 left-0 z-50 flex w-[18.5rem] max-w-[85vw] bg-background pl-[env(safe-area-inset-left)] outline-none duration-150 data-closed:animate-out data-open:animate-in"
+          data-slot="mobile-nav-drawer"
+        >
+          <DialogTitle className="sr-only">Navigation</DialogTitle>
+          <ProfileRail onNavigate={close} />
+          <AppSidebar onNavigate={close} variant="drawer" />
```
ahmadrosid: The pathname/search effect already closes this on any real navigation, including menus inside the drawer. Threading `onNavigate` through AppSidebar and every nav link is a second close path for the same clicks.

Keep `onNavigate` on ProfileRail — same-profile and `/chat` switches often don’t change the URL. Drop it from AppSidebar.

ahmadrosid (commented):
The shell and PWA pieces look right — drawer + precache-from-manifest is the shape I’d want. Two spots that can shrink without losing the phone/install behavior.

apps/web/src/context/chat-attachment-panel-context.tsx
yusupsupriyadi: Done in 2e9960f6. `usePhoneViewport` and `use-media-query.ts` are gone, and `chat-attachment-panel-context.tsx` is back to untouched, so the JS cutoff is out of the PR entirely.

The panel itself now carries `max-sm:left-0 max-sm:w-full! max-sm:max-w-none max-sm:min-w-0 max-sm:flex-1`, the resize handle takes `max-sm:hidden`, and `ChatPageColumn` collapses on `attachmentPanel.isOpen` with the `max-sm:` twin of the fullscreen classes it already had.

One note on the `!`: the panel sets its width through inline `style={{ width }}`, so a plain `max-sm:w-full` loses to it. `w-full!` is what wins, matching the trailing-`!` style already used in `ui/command.tsx`. Everything else is plain utilities.

apps/web/src/components/MobileNavDrawer.tsx
yusupsupriyadi: Done in 1c3446fc. `onNavigate` is off `AppSidebar`, `SidebarNavGroup`, `SidebarNavButton` and `SidebarNavLink`; only `ProfileRail` keeps it, and the pathname/search effect handles the rest.

Known gap I am fine with, flagging it so it is a choice and not a surprise: tapping the nav entry for the page you are already on leaves the drawer open, since the url does not change. Nothing happens on that tap anyway.

ahmadrosid:
Thanks for putting this together. The responsive shell and PWA setup match the goal of #841, and the CI checks are green. I reviewed the diff against the current `main` and found a few things to address before merging.

First, the branch is 17 commits behind `main` and conflicts with the recent composer draft changes in `apps/web/src/pages/chat/chat-page-content.tsx`. Please rebase before merging so the mobile composer safe-area changes and the draft persistence from #876 are both preserved.

In `apps/web/src/components/Layout.tsx` around the shell and header layout, the mobile shell only accounts for horizontal safe-area insets. There is no `safe-area-inset-top` padding for the authenticated app shell. On an iPhone running the app in standalone mode, the header and menu button can sit under the status bar or notch. Please add top safe-area handling to the shell or header and verify it on iOS.

In `apps/web/src/components/MobileNavDrawer.tsx`, the trigger is hidden at the `sm` breakpoint, but an already-open dialog can remain mounted when the viewport crosses that breakpoint. Rotating a phone or widening the browser can leave the mobile drawer over the desktop sidebar. Please close the drawer when `matchMedia("(min-width: 640px)")` becomes true, or otherwise ensure it cannot remain open above `sm`.

There is also a smaller performance concern: `ProfileRail` and `AppSidebar` remain mounted but hidden below `sm`, and opening the drawer mounts another navigation copy. This is not a blocker, but it may duplicate queries and state work on phones. It would be worth checking whether the desktop shell can be mounted only at the desktop breakpoint.

Please also check the header at 390px wide. The menu button, `AgentWorkTabs`, and page actions share the same row, so confirm there is no horizontal overflow.

The CSP hash, manifest, service-worker update flow, API cache behavior, and desktop layout look sound. After the rebase and fixes, please do a real-device check on iOS Safari and Android Chrome for install, standalone launch, offline shell behavior, and the update prompt. The core approach is good, but I would not merge the current revision yet.

yusupsupriyadi:
@ahmadrosid Rebased onto `main` (2f61d9c0) and pushed the five fixes as separate commits. The rebase was clean: `chat-page-content.tsx` keeps both `draftStorageKey={composerDraftKey}` from #876 and the composer's bottom safe-area padding.

**Top safe area** (c32019cc, 98f09396). The shell header now takes `pt-[env(safe-area-inset-top)]` with `h-[calc(3.5rem+env(safe-area-inset-top))]`, so `bg-card` paints under the status bar instead of the bar being squeezed. The drawer is a fixed portal, so the header's inset does not reach it and it needed the same padding of its own. With no insets the header still measures 56px.

**Drawer above sm** (42115501). A `matchMedia("(min-width: 640px)")` effect closes it on the crossing. Measured: open at 390px, widen to 900px, `[data-slot="mobile-nav-drawer"]` is gone and `body` carries no leftover inline scroll lock.

**Header at 390px** (86654c24, 0cc5ac4a). Confirmed, with the boundary a little lower than 390. The four-tab profile header is 309px, and `shrink-0` on the actions column meant its right edge landed at 357px inside a 320px viewport, clipped by the shell's `overflow-hidden` with no way to reach `Proposals`. The column is now `sm:shrink-0`, so below `sm` it gives way and the tab row scrolls, which is what the profile row's own `overflow-x-auto` was always waiting for. The system and automations rows got the same `overflow-x-auto`. Measured at 320px and 390px on `/profiles`, `/system` and `/automations`: `documentElement.scrollWidth` equals the viewport, every header child's right edge is inside it, and at 320px the profile row scrolls 309 in 260.

**Duplicate mount.** Left as is. `ProfileRail` and `AppSidebar` fetch through `useProfilesQuery` and `useAutomationUnreadTotal`, both TanStack Query, so the second copy reads the same cache entries and issues no extra requests. Gating the desktop copy on a breakpoint would mean putting back the JS media query you asked me to remove in the attachment-panel thread, and paying a remount on every rotation, to save two subscriptions.

Still open: a real notch. `env()` insets are 0 in headless Chrome, so what it can confirm is that the rule is emitted and the header is unchanged at 56px without insets. The real-device pass on iOS Safari and Android Chrome for install, standalone launch, offline shell and the update prompt is still open, and I will report back rather than assume it.

Checks on the rebased branch: `bun run --filter @nakama/web build` (tsc, vite, `sw.js` precaching 5 files), `bun test` in `apps/web` (373 pass), `bun test src/static-web.test.ts src/http/app.test.ts` in `apps/server` (49 pass), `bun run knip` clean, `ultracite check` clean.

apps/web/scripts/build-sw.ts
```diff
@@ -0,0 +1,75 @@
...
+  process.exit(1);
+}
+
+const manifest = (await manifestFile.json()) as ViteManifest;
+const precache = precacheUrls(manifest, EXTRA_PRECACHE_URLS);
+const version = buildVersion(precache);
+
+const result = await Bun.build({
+  define: {
+    __SW_PRECACHE__: JSON.stringify(JSON.stringify(precache)),
```
ahmadrosid: This can use a single `JSON.stringify(precache)`: Bun can substitute the array literal directly. Declare `__SW_PRECACHE__` as `string[]` in the worker and use it directly, which removes the extra encoding and runtime `JSON.parse`.

apps/web/src/lib/sw-precache.test.ts
```diff
@@ -0,0 +1,84 @@
...
+    expect(precacheUrls(MANIFEST, [])).toEqual([
+      "/",
+      "/assets/index-css111.css",
+      "/assets/index-js111.js",
+      "/assets/vendor-css222.css",
+      "/assets/vendor-js222.js",
+    ]);
+  });
+
+  test("leaves lazily loaded page chunks to the runtime cache", () => {
```
ahmadrosid: The test immediately above checks the entire returned array using the same manifest, so it already fails if the lazy SettingsPage chunk is included. You can drop this separate exclusion test and keep the same coverage.

ahmadrosid (commented):
Two small simplifications in the precache setup; the shared mobile sidebar looks reasonable.

apps/web/scripts/build-sw.ts
yusupsupriyadi: @ahmadrosid Done in 2521fe40. `define.__SW_PRECACHE__` is now `JSON.stringify(precache)`, the worker declares it as `string[]` and uses it directly, so the `JSON.parse` is gone. Checked Bun really substitutes an array literal before changing it, and the built `dist/sw.js` carries `["/","/assets/index-C8Rtkqee.js",...]` inline with zero `JSON.parse` occurrences left in the bundle.

Kept the `PRECACHE_URLS` const instead of using `__SW_PRECACHE__` at the call site, to stay symmetric with `VERSION = __SW_VERSION__` on the line above.

apps/web/src/lib/sw-precache.test.ts
yusupsupriyadi: @ahmadrosid Done in 2521fe40. Dropped it. The `toEqual` test above already fails if the lazy `SettingsPage` chunk appears, so its name now says so: "precaches the shell, the entry chunk and its static imports, nothing lazy". That keeps the reason the fixture has a lazy chunk at all visible without the second assertion.

PR #871, opened 2026-09-06
feat(plugins): install npm plugins for dashboard and agents

ahmadrosid:
## Summary

Admins install plugins that add native dashboard pages, agent tools, skills, and per-organization storage. Workflows ships as an official plugin.

**Before:** custom tools and skills had no installable package combining UI and per-org storage.
**After:** approved npm packages and official plugins use the organization lifecycle; pages render as native React components through DeepSeek Harness-inspired slots.

Browser modules export `inject` and `apply(ctx)`, register a page with `ctx.slots.register`, and use the host's React, actions, and styles. Effects clean up on unload or failed startup. Workflows now uses this architecture. `ui.entryModule` replaces the pre-release `entryHtml`/iframe contract. Nakama keeps its backend lifecycle; this does not adopt the full DeepSeek runtime.

Why safe:
1. Package preview pins digest/integrity; installation validates archive paths and runs no npm scripts or dependency installation.
2. Authorization, CSRF, org isolation, migration rollback, and backup protections remain tested; native UI tests cover failed startup, cleanup, cancellation, and late org responses.

Plugins remain trusted code without a sandbox. Third-party installs require an exact public npm version; local/Git/ZIP sources and private registries remain unsupported.

## Test plan

- [x] `bun run test` (3,121 pass); final `bun test plugin plugins` (127 pass) and `bun run --cwd apps/web test` (413 pass), including a render test of the shipped Workflows module.
- [x] Web/server/Workflows builds, strict Workflows UI typecheck, `bun run check`, `bun run knip`, and React Doctor (100/100).
- [ ] System → Plugins: install and enable Workflows, open its page, then switch organization and theme. Browser interaction remains unverified.

Known limits: backend lifecycle hooks and profile-pack plugin reassignment remain deferred. Shutdown does not await the full kill grace window; disabling mid-reply can rebuild the cached chat session. Some OpenAPI errors remain under-documented and use machine codes. The standalone root TypeScript check failed on the prior PR head; the web typecheck passes.


huanglinfei091-cmd:
Hi @ahmadrosid — I noticed the `knip` job failed at `Run knip` in [this GitHub Actions run](https://github.com/ahmadrosid/nakama/actions/runs/34055761725) for this PR.

I’m testing [RunBack](https://github.com/huanglinfei091-cmd/runback), an early alpha CLI that reconstructs a failed public Actions run locally with `act`. For this run, the useful question is whether the same Knip unused-code/dependency finding appears locally, or whether reconstruction stops with an honest blocked result.

```bash
runback doctor
runback https://github.com/ahmadrosid/nakama/actions/runs/34055761725
```

The Linux amd64 binary and checksum are in the [v0.1.0-alpha release](https://github.com/huanglinfei091-cmd/runback/releases/tag/v0.1.0-alpha). If you try it, the verdict (`SAME_FAILURE`, `DIFFERENT_FAILURE`, or `REPLAY_BLOCKED`), Stage/Cause, and any install friction would be useful. Please do not post tokens or secrets. I’m asking for candid alpha evidence rather than a star or promotion; a failed reproduction is useful too.

fajarhide:
@ahmadrosid the host shape reads right to me: install never executes plugin code, the org context is host-derived rather than taken from plugin input, and `actorMayInvoke` is default-closed on an unknown role. One correctness thing, then three shrinks.

**The `orgRole` rewrite at `apps/server/src/http/routes/sessions.ts:393` costs a real org admin their Super Bot access.** `createSession` runs `canAccessSuperBotProfile({ orgRole: options.orgRole })` at `agent-service.ts:1650`, before `buildChatSession` ever sees the value, so hardcoding `"member"` for telegram/whatsapp/discord makes that check fail for an admin who is not also a platform admin. I stubbed the agent and captured what the route hands it:

```
orgRole handed to agent.createSession: member
canAccessSuperBotProfile with that role: false
```

Reverting only that hunk gives `admin` and `true`. The plugin side does not need the route copy: `buildChatSession:3441` already coerces the same three channels into `pluginOrgRole`, and `pluginActorFromContext` (`tool-resolver.ts:232`) does it a third time at invoke. Dropping it leaves the rule in the two places that see every path, including the channel workers that never go through `POST /v1/sessions`.

`extraEnv` on `SpawnJsonToolTransport` (`custom-tool-subprocess.ts:43`) has no caller. It is also the one field that punches through the env allowlist that file exists for, so I would delete it now rather than leave it. `detached: false` in the same spawn is the default already.

`capabilityRevisionForOrg` reads `listOrgPlugins()` for every org and filters in JS, and it runs twice per message (`resolveSession`, `beginSessionTurn`). The migration creates `org_plugins_org_id` and nothing queries by it. An `orgId` argument on `listOrgPlugins` uses the index and drops the scan.

`pluginRowIdentity` and `pluginHasRetainedData` (`use-plugins.ts:397` and `:297`) are exported, but only `plugins.test.tsx` calls them. The page never does.

What I checked and could not fault: traversal on `/v1/plugins/ui/*` (`resolvePluginReleaseEntry` plus the `isUiReleasePath` allowlist), the path org id compared against the session org, the CSP and `frame-ancestors` split for plugin documents, and the abort rewrite in `spawnJsonTool`, where a cancelled turn rejects in 210ms and a pre-aborted one in 0ms, both after the child closed. `bun run test` is green in a fresh worktree here, 3004 tests, and knip is clean. I did not install a real plugin ZIP, so the one unticked box in your test plan is still unticked.

apps/server/src/http/routes/sessions.ts
```diff
@@ -390,7 +390,12 @@ export function registerSessionRoutes(
         excludeSuperBot: auth.mode === "local-token" && channel !== "cli",
         isPlatformAdmin: auth.isPlatformAdmin,
         model: body.model,
-        orgRole: auth.orgRole,
+        orgRole:
+          channel === "telegram" ||
+          channel === "whatsapp" ||
+          channel === "discord"
+            ? "member"
+            : auth.orgRole,
```
fajarhide: `buildChatSession:3441` already coerces these three channels into `pluginOrgRole`, and `pluginActorFromContext` does it a third time at invoke. This copy runs before `canAccessSuperBotProfile` at `agent-service.ts:1650`, so a plain org admin loses Super Bot here.

```ts
        orgRole: auth.orgRole,
```

apps/server/src/services/custom-tool-subprocess.ts
```diff
@@ -37,6 +38,14 @@ function buildAllowlistedSubprocessEnv(
   return env;
 }
 
+interface SpawnJsonToolTransport {
+  extraArgs?: string[];
+  extraEnv?: NodeJS.ProcessEnv;
```
fajarhide: `extraEnv` has no caller, and it is the one field that reopens the env allowlist this file exists for. `detached: false` in the same spawn is the default already.

```ts
interface SpawnJsonToolTransport {
  extraArgs?: string[];
  includeConfigDir?: boolean;
  /** When set, written to stdin instead of `input`. Legacy callers omit this. */
  stdin?: unknown;
}
```

and at :74:

```ts
  const env = buildAllowlistedSubprocessEnv(
    workspaceRoot,
    transport?.includeConfigDir ?? true
  );
```

apps/server/src/services/plugin-service.ts
```diff
@@ -0,0 +1,2571 @@
...
+      };
+    } finally {
+      handle.release();
+    }
+  }
+
+  async capabilityRevisionForOrg(orgId: string): Promise<string> {
+    const installs = (await this.db.listOrgPlugins()).filter(
+      (install) => install.orgId === orgId
+    );
```
fajarhide: This reads every org's installs and filters in JS, twice per message through `resolveSession` and `beginSessionTurn`. `listPluginReleases(pluginId?)` in the same adapter already has the shape, and the migration creates `org_plugins_org_id` that nothing queries by.

```ts
    const installs = await this.db.listOrgPlugins(orgId);
```

with `listOrgPlugins(orgId?: string)` in `types.ts` and, in `adapters/sqlite.ts`:

```ts
  const listOrgPluginsForOrgStmt = db.prepare(
    "SELECT * FROM org_plugins WHERE org_id = ? ORDER BY plugin_id"
  );

    async listOrgPlugins(orgId) {
      const rows = orgId
        ? listOrgPluginsForOrgStmt.all(orgId)
        : listOrgPluginsStmt.all();
      return rows.map((row) => toOrgPluginRecord(row as OrgPluginRow));
    },
```

`listOrgPluginDetails:935` takes the same argument; the `pluginId` filter at :1005 and the recovery sweep at :1108 want every org and stay as they are.

apps/web/src/hooks/use-plugins.ts
```diff
@@ -0,0 +1,425 @@
...
+    plugin.pendingOperation !== null ||
+    plugin.lifecycleState === "enabling" ||
+    plugin.lifecycleState === "disabling" ||
+    plugin.lifecycleState === "updating"
+  );
+}
+
+export function pluginHasRetainedData(plugin: OrgPluginDetail): boolean {
+  return (
+    plugin.lifecycleState === "retained" || plugin.databaseGeneration !== null
```
fajarhide: Only `plugins.test.tsx:430` calls this, the page never does. Same for `pluginRowIdentity` at :397, whose test pins the `name pluginId` string. Deleting both helpers and those two assertions is 24 lines.

packages/db/src/migrate.ts
```diff
@@ -1519,6 +1520,71 @@ function migrateProfileChangeEventsTable(db: Database): void {
...
+      pending_operation TEXT,
+      last_lifecycle_error TEXT,
+      created_at TEXT NOT NULL,
+      updated_at TEXT NOT NULL,
+      PRIMARY KEY (org_id, plugin_id),
+      FOREIGN KEY (org_id) REFERENCES organizations (id) ON DELETE CASCADE
+    );
+  `);
+
+  addNullableTextColumnIfMissing(db, "plugin_releases", "digest");
```
fajarhide: `plugin_releases` is created with `digest` twelve lines above this, so the guard can never fire. The four `tools` and `skills` calls under it are the real ones.

fajarhide (commented):
The points from my comment above with the patch I ran for each, plus one dead migration line I missed there. Together they are 16 insertions and 51 deletions across eight files, `bun run test` green at 3004 tests and knip clean on my worktree.

apps/server/src/http/routes/sessions.ts
ahmadrosid: > so a plain org admin loses Super Bot here.

Fixed in bd20d245. Session creation now receives the authenticated org role. Added HTTP regression coverage for org admins and members across Telegram, WhatsApp, and Discord; plugin tool role restrictions stay in the dispatcher.

apps/server/src/services/custom-tool-subprocess.ts
ahmadrosid: > `extraEnv` has no caller

Removed `extraEnv` and the redundant `detached: false` in bd20d245. Subprocess environments now come directly from the allowlist. JavaScript, Python, and plugin runtime tests pass.

apps/server/src/services/plugin-service.ts
ahmadrosid: > This reads every org's installs and filters in JS

Fixed in bd20d245. Both organization-scoped service calls now use `listOrgPlugins(orgId)` with a SQL WHERE clause. Recovery and release removal retain the all-org query. Tests cover both organizations, unfiltered reads, and empty or unknown org IDs.

apps/web/src/hooks/use-plugins.ts
ahmadrosid: > the page never does.

Removed both unused helpers and their implementation-only assertions in bd20d245. Plugin page tests, Knip, the web build, and React Doctor pass.

packages/db/src/migrate.ts
ahmadrosid: > `plugin_releases` is created with `digest` twelve lines above this

Removed the redundant digest-column migration in bd20d245. The tools/skills upgrade guards remain. The legacy-database migration test still verifies digest creation and repeated migration.

ahmadrosid:
> One correctness thing, then three shrinks.

Addressed these and the extra migration comment from the review in bd20d245: restored authenticated org roles for Super Bot session creation, scoped plugin lookups in SQL, and removed the unused environment override, UI helpers, and redundant digest migration.

104 targeted tests pass, including real HTTP checks that org admins can create messaging Super Bot sessions while members cannot. Web build, Knip, and lint pass; React Doctor is 100/100. The earlier Knip failure mentioned above was already green on the reviewed head, and it also passes locally after these fixes.

fajarhide:
@ahmadrosid re-checked at 223a86c8, all five are in. Your `orgId === undefined` is better than the truthiness check I proposed: on mine an empty string would have listed every org.

I put the coercion back to see whether the new `rbac-mutations` case bites, and it does.

```
Expected: 201
Received: 403
(fail) messaging Super Bot sessions enforce the authenticated admin role
```

Full suite green on that head here, 3025 tests.

One small thing from 5ba07177, not from the review: `retryable` is set at three call sites and nothing in production reads it. `throwPluginHttpError` passes only `error.code`, so the single reader is `plugin-runtime.test.ts:462`.

PR #876, opened 2026-09-07
fix(chat): persist user-scoped composer drafts

### LINKED ISSUE #861: Persist chat composer draft across refresh (per user)
## Summary

Keep the chat composer text when the user types, so a page refresh does not wipe the draft. Store it per logged-in user so accounts on the same browser do not share drafts.

## Why it matters

The composer at `/chat` (InputGroup) is empty after refresh. Users lose mid-thought text. Existing `storeChatDraft` / `consumeStoredChatDraft` in `apps/web/src/lib/chat-history.ts` is a one-shot `sessionStorage` handoff for `?draftKey=` navigation — it is not a live typing cache and it is not user-scoped.

- `apps/web/src/components/chat/chat-composer.tsx` — persist while typing, restore on mount
- `apps/web/src/pages/chat/use-chat-page.ts` — composer state / new-chat draft restore
- `apps/web/src/lib/chat-history.ts` — do not reuse consume-once `draftKey` for this
- Auth identity (user id) + active org — key prefix so users (and orgs) stay isolated

## Acceptance criteria

- [ ] Typing in the chat composer writes a draft to browser storage (debounced is fine)
- [ ] Refresh on the same chat/new-chat surface restores the last typed text
- [ ] Sending a message (or explicitly clearing the composer) removes that draft
- [ ] Draft keys include the current user id; switching accounts on the same browser does not show the other user's draft
- [ ] Drafts do not leak across orgs (include `orgId` in the key)
- [ ] Existing `?draftKey=` / `storeChatDraft` navigation still works
- [ ] Tests cover restore, clear-on-send, and user/org isolation

mamenesia:
## Summary

Keep unfinished chat text across refresh without mixing accounts or organizations. Fixes #861.

**Before:** refresh erased typed text; `draftKey` only supported a consume-once navigation handoff.
**After:** live composer text restores per user, org, profile, and chat; sending or deleting text removes the draft.

Why safe:
1. Auth responses expose the stable user ID used in draft keys; no email-based identity or database migration.
2. The existing input provider owns live text; navigation prefills are applied and acknowledged once.
3. Draft-storage failures do not block editing; existing consume-once handoffs remain separate.

Residual: drafts live in browser storage, not encrypted or synced across devices. Storage denial means refresh persistence is unavailable.

## Test plan
- [x] `bun test apps/web/src apps/server/src/app.test.ts apps/server/src/services/org-service.test.ts` (404 pass after rebasing onto current main).
- [x] `bun x tsc --noEmit -p apps/web/tsconfig.json` and targeted `bun x ultracite check` on all 10 changed files.
- [x] Local browser checks before rebase: new/existing-chat refresh, send/delete clearing, account/org isolation, editable consume-once prefills, and denied draft storage. Send requests were blocked before the model call; this verifies clearing, not successful AI replies.
- [ ] CI and fresh draft review before marking ready for review.
- [ ] Type a draft, refresh, then clear it and refresh again: text restores once and stays cleared.

### DIFF
--- a/apps/web/src/components/chat/chat-composer.tsx
+++ b/apps/web/src/components/chat/chat-composer.tsx
@@ -17,14 +17,7 @@ import {
   Image01Icon,
   WifiOff01Icon,
 } from "hugeicons-react";
-import {
-  useCallback,
-  useEffect,
-  useLayoutEffect,
-  useMemo,
-  useRef,
-  useState,
-} from "react";
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import {
   PromptInput,
   PromptInputBody,
@@ -75,7 +68,7 @@ import {
   type SkillSlashRange,
 } from "@/lib/chat-composer-skills";
 import type { ChatContextUsage } from "@/lib/chat-context-usage";
-import { type ComposerPrefill, storeComposerDraft } from "@/lib/chat-history";
+import { storeComposerDraft } from "@/lib/chat-history";
 import {
   ALL_ATTACHMENT_ACCEPT,
   DOCUMENT_ACCEPT,
@@ -110,12 +103,10 @@ interface ChatComposerBaseProps {
   draftStorageKey?: string | null;
   error: string | null;
   footerClassName?: string;
-  onPrefillConsumed?: (prefill: ComposerPrefill) => void;
   onStop?: () => void;
   onSubmit: (text: string, files: FileUIPart[]) => void;
   onSubmitQuestionnaire?: (answers: AgentQuestionAnswer[]) => void;
   placeholder?: string;
-  prefill?: ComposerPrefill | null;
   questionnaire?: AgentQuestionnaire | null;
   queuedMessages?: QueuedComposerMessage[];
   todos?: AgentTodo[];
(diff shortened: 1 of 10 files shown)


apps/web/src/components/chat/chat-composer.tsx
```diff
@@ -536,6 +547,41 @@ function ChatComposerMain({
 }
 
 export function ChatComposer(props: ChatComposerProps) {
+  const { textInput } = usePromptInputController();
+  const { prefill, onPrefillConsumed } = props;
+  useLayoutEffect(() => {
```
ahmadrosid: Drop `prefill` / `onPrefillConsumed` and this layout effect. Persist `textInput.value` and clear on send is enough here.

The consume-once navigation draft can land by writing it with `storeComposerDraft` and remounting `PromptInputProvider` (bump the `key` when a handoff arrives). Same apply-once behavior, without a second input channel.

apps/web/src/pages/chat/use-chat-page.ts
```diff
@@ -154,7 +159,24 @@ export function useChatPage() {
   );
   const [canStop, setCanStop] = useState(false);
   const [error, setError] = useState<string | null>(null);
-  const [composerDraft, setComposerDraft] = useState("");
+  const [composerPrefill, setComposerPrefill] =
```
ahmadrosid: `composerPrefill`, `consumeComposerPrefill`, and the scope-key effect can go with that remount. At the `?new=1` handoff, `storeComposerDraft(targetDraftKey, requestedDraft)` then bump a nonce so the provider remounts and reads storage. `ComposerPrefill` is unused after that.

ahmadrosid (commented):
Refresh persistence is solid. The consume-once prefill handshake is extra — write that draft into the same storage and remount the input provider.

apps/web/src/components/chat/chat-composer.tsx
mamenesia: Updated in https://github.com/ahmadrosid/nakama/commit/43bcdf7e: removed prefill/onPrefillConsumed, the layout effect, and ComposerPrefill. Navigation handoffs now persist the draft and bump the provider key.

One small difference: the remount also receives the incoming text in memory. mamenesia verified that reading only storage loses the consumed handoff when localStorage is full (restores the old draft) or denied (restores empty). The direct initialInput keeps that path working without an acknowledgment channel.

Local browser checks passed for handoffs, repeated/empty handoffs, quota/denied storage, refresh, clear/send, and account/org/profile isolation. All 353 web tests, web TypeScript, and targeted lint pass. Net 26 lines removed.

PR #880, opened 2026-09-07
feat(web): refresh a workflow run while it is still running

### LINKED ISSUE #874: Show step status while a workflow run is in progress
## Summary

The run history panel renders step receipts once a run has finished. While a run
is in flight it shows nothing new, so pressing **Run now** gives no feedback until
the whole recipe settles.

Refresh the open run while its status is `running`, and stop refreshing when it
settles.

## Why it matters

Every step already records its own status, input, output and error in
`workflow_run_steps` as it goes, so the data is written and never read back
until the end. With a slow `tool` step that means a blank panel for as long
as the step takes, and no way to tell a slow run from a stuck one. That is also
the only signal a user has for the orphaned runs in the bug filed alongside this.

- `apps/web/src/pages/workflows/workflow-builder.tsx`, `WorkflowRunHistory` and
  `WorkflowRunHistoryDetail`
- `apps/web/src/hooks/use-workflows.ts`
- `GET /v1/workflows/:workflowId/runs/:runId` already returns the steps

No new dependency, and no second streaming surface: chat already owns SSE here,
and polling a single run while it is open costs less than adding another stream.

## Acceptance criteria

- [ ] an open run whose status is `running` refreshes on its own
- [ ] each step shows pending, running, completed or failed as the run proceeds
- [x] refreshing stops once the run reaches `completed` or `failed`
- [ ] a finished run is not polled at all, proven by the request count in the
      network panel
- [ ] leaving the page stops the refresh

fajarhide:
Closes #874.

`WorkflowRunHistoryDetail` already renders `runStatusLabel(step.status)` for every step, so the panel could always show a run progressing. The query behind it never refetched, which is why Run now looks dead until the recipe settles.

The whole change is one option on the existing query: poll while any run reads `running`, return `false` the moment none does. A finished workflow makes zero requests, `refetchIntervalInBackground` stays off so a hidden tab pauses, and leaving the page unmounts the query. No new dependency, and no second streaming surface.

`workflowRunsRefetchInterval` is exported so the decision is testable without a DOM. The test bites: with the predicate stubbed to `false` it fails on `Expected and actual values must be numbers`, and passes with it back. `bun run test` is green across 11 workspaces, 2916 pass 0 fail.

What I could not capture on this machine, so you can judge it rather than take my word: the request count in a real browser. Playwright is not installed here, and a bare `query-core` harness does not run interval timers at all, a static `refetchInterval: 500` fetched once in three seconds there too. So the polling decision is proven and the one line that hands it to `useQuery` is only proven by types.

### DIFF
--- a/apps/web/src/hooks/use-workflows.test.ts
+++ b/apps/web/src/hooks/use-workflows.test.ts
@@ -1,0 +1,41 @@
+import { describe, expect, test } from "bun:test";
+import type { WorkflowRunRecord } from "@nakama/core/contract";
+import { workflowRunsRefetchInterval } from "./use-workflows";
+
+function run(
+  id: string,
+  status: WorkflowRunRecord["status"]
+): WorkflowRunRecord {
+  return {
+    completedAt: status === "running" ? null : "2026-09-07T00:00:01.000Z",
+    error: null,
+    id,
+    input: null,
+    output: null,
+    startedAt: "2026-09-07T00:00:00.000Z",
+    status,
+    workflowId: "wf_a",
+  };
+}
+
+describe("workflowRunsRefetchInterval", () => {
+  test("polls only while a run is in flight", () => {
+    const interval = workflowRunsRefetchInterval([
+      run("wfrun_live", "running"),
+      run("wfrun_done", "completed"),
+    ]);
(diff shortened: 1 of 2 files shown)


yusupsupriyadi (approved):
LGTM

Issue #217, opened 2026-08-09
Support first-boot seed (env/mounted config) so managed instances skip the setup wizard


ahmadrosid:
## Summary

Nakama Cloud (ahmadrosid/nakama-cloud, issue #7) wants managed instances to deploy pre-configured and skip the setup wizard entirely. The nakama app needs a small, env-gated first-boot seed to make that possible: provision the account, org (`Personal`), and default provider (OpenCode Zen) through the app's own config layer before the first request is served. Self-hosted installs must be completely unaffected.

## Why the app must support this (no cloud-side-only fix)

- `apps/web/src/components/SetupGuard.tsx` redirects to `/setup` unless `health.userConfigured === true && health.providerConfigured === true`.
- `userConfigured` is `humanUserCount > 0` (`apps/server/src/services/system-status-service.ts:162`) — the wizard's account step (SetupStepAccount) creates the admin user. A seed that only writes the provider leaves the wizard showing.
- Existing env hooks are insufficient: `NAKAMA_PROVIDER` (`packages/core/src/provider-resolution.ts`) selects a provider by name and requires a matching real API-key env var. It cannot express `openai_compatible` + `base_url` + `models_json`, and it never creates an org or user.
- `POST /v1/auth/setup` (`apps/server/src/http/routes/auth.ts`) does create the first org + admin account + session, but it is built for a human-in-the-loop browser flow; driving it from the control plane after readiness adds a bootstrap-credential handoff and a window where the user still sees the wizard.

## Proposed design

Add a first-boot seed module, enabled only when explicit env config is present (e.g. `NAKAMA_SEED_*` vars or a seed file path; the exact transport is an open question below). On boot:

1. If the app is already configured, skip entirely (idempotent, no clobbering on restart).
2. Otherwise provision, through existing services/config layer:
   - admin account (so `humanUserCount > 0` -> `userConfigured`)
   - org named `Personal`
   - provider `openai_compatible`, label `OpenCode Zen`, base_url `https://opencode.ai/zen/v1`, api_key `public`, models `big-pickle` ("Big Pickle", supportsThinking) and `hy3-free` ("Hy3 Free", supportsThinking) — shape matches `ProviderInstance` in `packages/core/src/user-config.ts`
3. Server then serves with `userConfigured && providerConfigured` true, so SetupGuard passes and the user lands in the app.

[...]

PR #882, opened 2026-09-07
feat(web): guided onboarding that fills USER.md from answers, re-runnable after setup

### LINKED ISSUE #398: Extract Agent Swarm (Tasks) into a Nakama extension
## Summary

Take **Agent Swarm** — the Tasks board at **Agent work → Tasks** (`/automations?tab=tasks`, also `/tasks`) — out of core Nakama and ship it as an **optional extension**.

Today this is a first-class product surface: a five-column kanban (Backlog → To Do → In Progress → Done → Failed), create/edit, LLM prompt draft, drag-and-drop, run-on-a-profile, and run/chat history. It is baked into the dashboard, API, and SQLite schema. Teams that only want chat, automations, or channels still pay the complexity.

Automations stay in core. This issue is only the Tasks / Agent Swarm product.

## Why

- **Core stays smaller.** Agent Swarm is a work-management product on top of profiles and chat, not a prerequisite for running Nakama.
- **Optional value.** Orgs that want a swarm board install the extension. Orgs that do not never see the tab, routes, or tables as a required surface.
- **Proof of the extension model.** This is the first real feature to extract, so it should force a real extension host (UI + HTTP + data), not a docs-only plugin story.

## What exists today (must keep working)

| Piece | Behavior |
| --- | --- |
| Board | Kanban titled **Agent Swarm**. Columns: `backlog`, `todo`, `in_progress`, `done`, `failed`. Cards keep an explicit `position`. |
| Create / edit | Title, description, prompt, bound `profileId`. |
| Draft prompt | `POST /v1/tasks/draft-prompt` turns title + description into a runnable prompt. |
| Run | `POST /v1/tasks/:taskId/run` starts the bound profile, writes a `task_runs` row, binds a chat session, moves the card to `in_progress` → `done` / `failed`. |
| Review | Run history + session messages on Done / Failed cards. |
| Auth | Org-scoped (`org_id`). Viewers read only. Non-viewers cannot bind Super Bot unless they may use that profile. |

### Current code (for extract, not rewrite)

- UI: `apps/web/src/pages/TasksPage.tsx`, `apps/web/src/components/tasks/*`, tab switch in `apps/web/src/pages/automations/agent-work-tabs.tsx`
- API: `apps/server/src/http/routes/tasks.ts` — `/v1/tasks`, `/v1/tasks/:taskId`, `/run`, `/runs`, `/messages`, `/draft-prompt`
- Services: `apps/server/src/services/task-service.ts`, `task-runner.ts`, `task-validate.ts`
- Schema: `packages/db/sql/schema.sql` — `tasks`, `task_runs`
- Contracts: `StoredTask`, `TASK_STATUSES` in `packages/core/src/contract.ts`
- Docs: [Automations and tasks](https://github.com/ahmadrosid/nakama/blob/main/docs/website/content/docs/automations.mdx)

## Proposed outcome

1. **Extension host (prerequisite or same epic).** A Nakama extension can register dashboard pages/nav, HTTP routes, and its own data — enabled per org (or per instance). Until that host exists, this feature cannot leave core without becoming a hard fork.
2. **`tasks` (or `agent-swarm`) first-party extension.** Same board, same APIs, same permissions. Core dashboard no longer ships the Tasks tab when the extension is off.
3. **Clean core.** No Agent work → Tasks tab, no `/tasks` redirect, no `/v1/tasks*` in the core OpenAPI, no required `tasks` / `task_runs` tables for a core-only install.
4. **Migration.** Existing `tasks` / `task_runs` rows survive enable/disable. Turning the extension off hides the UI; it does not delete work.

## Open questions (this is why the issue is `stage: later`)

These need a decision before anyone can size a PR:

1. What is a Nakama extension — first-party in-repo package, installable tarball, or both?
2. Enable per org, per instance, or both?
3. Does the extension host ship in this issue, or as a separate prerequisite issue?
4. Do `/v1/tasks*` stay as compatibility aliases, or move under an extension prefix?
5. When the extension is off, do we hide the tables or keep them inert?

## Acceptance criteria

- [ ] Extension host can register a dashboard page and HTTP routes
- [ ] Enabling the extension restores Agent Swarm at a stable path (today: `/automations?tab=tasks` or `/tasks`)
- [ ] Disabling it removes the Tasks tab and `/v1/tasks*` from a core-only install
- [ ] Create, draft prompt, drag, run, and review history still work for a non-viewer
- [ ] Viewers still cannot mutate; Super Bot profile binding stays gated
- [ ] Existing org task rows migrate / remain readable after the extract
- [ ] User docs describe Tasks as an extension, not a core Agent work tab

## Out of scope

- Reworking Automations (schedules, delivery, run badges)
- A browser / VS Code extension
- A public third-party marketplace (later)

## Gating

Do not start the extract until the extension host questions above have an answer — even a one-pager. The first PR should be the host + this feature moving onto it, or the host alone if that is split out.

### LINKED ISSUE #869: feat(web): guided onboarding that fills USER.md from answers, re-runnable after setup
## Summary

A guided onboarding flow that asks a new user a short list of questions (name, what they do, current projects, stack, how they want replies, hard always/never rules) and writes the answers into `USER.md`, the personalisation blob Nakama already injects into every system prompt.

It replaces the blank markdown textarea that setup step 4 shows today, and unlike that step it can be run again after setup is complete.

Who uses it: every operator on a fresh install, plus everyone on an install seeded past the wizard with `NAKAMA_SEED_ADMIN_*`, who today is never asked at all.

This is phase 1 of two, on purpose. Deciding which features to enable from the same answers is phase 2 and is out of scope here; see "Proposed scope and non-goals".

## Why

Personalisation already exists, and almost nobody fills it in.

- Step 4 "About You" (`apps/web/src/components/setup-wizard/setup-wizard.shared.ts:13-18`, `required: false`) renders `UserContextSettings` (`apps/web/src/components/setup-wizard/SetupStepUserContext.tsx:32`), which scaffolds `USER_CONTEXT_TEMPLATE` (`packages/core/src/user-context.ts:1-10`) into a `<Textarea>`. On their first minute with the product the user sees seven empty markdown bullets next to a "Set up later" link. Skipping is the path of least resistance.
- Skipping is unrecoverable from the UI. Once `userConfigured && providerConfigured` are both true, `SetupGuard` (`apps/web/src/components/SetupGuard.tsx:21-23`) stops redirecting, so `/setup` is a dead route and step 4 never returns. The only remaining entry point is the sidebar avatar menu, "Personalisation" (`apps/web/src/components/SidebarUserMenu.tsx:101-105`), which a new user has no reason to open. `SettingsPage.tsx` never imports `UserContextSettings`, so it is absent from `/settings` entirely.
- Docker and managed installs using `NAKAMA_SEED_ADMIN_EMAIL/NAME/PASSWORD` skip the wizard completely (`docs/website/content/docs/first-time-setup.mdx:17`), so those users are never asked once.
- Nothing records that the step was skipped, so Nakama cannot ask again later.

The cost is concrete. `buildChatSystemPrompt` injects `USER.md` under `# Personalisation (USER.md)` (`packages/agent/src/chat-prompt.ts:120-126`) for web chat, the CLI, and every channel. Empty means the assistant answers every user the same way, and the operator never learns the knob existed.

The fields are already the right ones. What is missing is asking them as questions instead of handing someone a markdown file to fill in.

## Proposed scope and non-goals

**Phase 1, this issue**

- A guided step: each question maps to one line of `USER_CONTEXT_TEMPLATE`, asked as a labelled input with an example, not as raw markdown.
- Answers render deterministically into the same markdown shape and save through the existing `PUT /v1/user/context`. No LLM call, so it works regardless of provider health and produces a file the user can read and diff.
- The raw editor stays. The last screen shows the generated `USER.md` and lets the user edit it before saving; `UserContextEditorDialog` is unchanged for later edits.
- Reachable after setup: an entry point on `/settings` and from the existing sidebar "Personalisation" item, so seeded installs and users who skipped can still run it.
- Still skippable, and it never blocks chat.

**Non-goals in phase 1**

- **No feature enabling or disabling.** Nakama has no feature on/off system today, so onboarding answers must not gate any capability. Wiring answers to capabilities before that system exists would just hard-code a second configuration path.
- No provider, model, or channel selection. That is step 3 and `/integrations`.
- No new table and no migration. `org_members.user_context` (`packages/db/sql/schema.sql:293`) already stores this. A "skipped/completed" flag so Nakama can re-prompt is the one piece that would need storage, and it is deliberately left out of phase 1.
- No change to profile soul files (`SOUL.md` / `STYLE.md`). A profile's voice is a separate concern from a user's personalisation.

**Phase 2, a separate issue once feature toggles exist**

The same answers pick a sensible default set of enabled features, so an install stays lean instead of shipping every surface to everyone. This is blocked until a toggle or extension system lands; #398 is the closest existing track and states an extension host is a prerequisite.

- `apps/web/src/components/setup-wizard/SetupStepUserContext.tsx:29-56` is the step body.
- `apps/web/src/components/setup-wizard/setup-wizard.shared.ts:13-18` holds the step list. Note that `SetupWizard.tsx` hardcodes the upper bound (`>= 4` at `:34`, `:43`, `:68`, `:76`, and `case 4:` at `:125`), which matters if a step is ever added.
- `apps/web/src/components/UserContextCard.tsx:40` (`UserContextEditorDialog`) can be reused for the review-and-edit screen; `:512` (`UserContextSettings`) is the settings row.
- `apps/web/src/pages/SettingsPage.tsx` gains the post-setup entry point.
- `packages/core/src/user-context.ts:1-10`: `USER_CONTEXT_TEMPLATE` becomes the source of the question list plus a render function.
- `apps/server/src/http/routes/user-context.ts:98`, `:107`, `:115` already expose `GET /v1/user/context`, `PUT /v1/user/context`, and `POST /v1/user/context/init`. No new endpoint expected.
- `docs/website/content/docs/first-time-setup.mdx:60-62`: rewrite "Step 4: About you" and refresh the screenshot.
- `docs/website/content/docs/agent-prompt.mdx`: note where the answers land.

## Acceptance criteria

- [ ] Setup step 4 asks the `USER_CONTEXT_TEMPLATE` fields as labelled inputs, not as raw markdown
- [ ] Answering and continuing produces a `USER.md` whose content is the rendered answers, readable at `GET /v1/user/context?content=true`
- [ ] Leaving every field blank saves nothing: `GET /v1/user/context` still reports `active: false`
- [ ] The generated markdown is shown for review and is editable by hand before saving
- [ ] "Set up later" still skips the step and lands the user in chat
- [ ] The same flow can be started from `/settings` on an install where `userConfigured` and `providerConfigured` are both already true
- [ ] Re-running it for a user who already has `USER.md` pre-fills from the saved content and never silently discards hand-written text
- [ ] A chat afterwards carries the answers under `# Personalisation (USER.md)` in the system prompt built by `buildChatSystemPrompt`
- [ ] No answer enables or disables any feature
- [ ] No new table or column; `packages/db/sql/schema.sql` is unchanged
- [ ] `bun test`, `bun run check`, and `bun run knip` pass
- [ ] `docs/website/content/docs/first-time-setup.mdx` step 4 matches the new flow

## Related issues

Searched open and closed for `onboarding`, `personalization`, `persona`, `first run`, `wizard`, `settings`, `toggle`, `disable`, `modular`, `plugin`, `opt-in`, `capabilities`, `lightweight`, `soul`, `user profile`, and `system prompt`. Nothing reports this, so there is no closing keyword on this issue.

Related, not fixed:

- #398, the extension and modularity track. Phase 2 of this issue depends on a system like it.
- #217 (closed). First-boot seed skips the wizard, which is exactly why this flow has to be reachable from `/settings` too.

Measured on `main` at `d9cf5c26`.

yusupsupriyadi:
## Summary

Setup step 4 asks a new user who they are, one question per screen, and writes the answers into `USER.md`. Reachable again later from Settings → Personalisation.

**Before:** step 4 dropped a blank `USER.md` textarea in front of a first-time user, next to "Set up later"; once setup finished the step was unreachable and `/settings` never showed it.
**After:** role cards with icons, role-aware stack quick picks, reply language, length and tone cards with sample sentences, then the exact Markdown for review; `renderUserContext` writes it through the existing `PUT /v1/user/context`.

Why safe:
1. The file format is unchanged: `USER_CONTEXT_TEMPLATE` is now derived from the same seven labels and a test pins the exact bytes, so the server and CLI paths read what they always read
2. No new storage, endpoint, or LLM call; the form is a view over the raw `USER.md` string, so hand-written notes outside the bullets survive a round trip (tested)
3. Untouched or fully blank answers still save nothing (`active: false`); the old auto `/init` on mount, which made an empty template report active, is gone

Only revisit if an existing install turns out to rely on the dashboard scaffolding `USER.md` through `/init` on open; the CLI `/user init` path is unchanged.

Fixes #869

Related, not fixed: #398 (extension host; phase 2 of #869 depends on a toggle system like it), #217 (seeded installs skip the wizard, which is why the flow is reachable from Settings too).

## Test plan
- [x] `bun test packages/core/src/user-context.test.ts` (13 pass)
- [x] `bun test packages/db/src/user-context.test.ts apps/server/src/http/user-context.test.ts` (3 pass)
- [x] `cd apps/web && tsc --noEmit`, `bun x ultracite check` on the changed files, `bun run knip` (clean)
- [x] Fresh install in the browser: walked account → org → provider → step 4, Finish wrote the rendered file to `org_members.user_context`; Settings → Personalisation pre-filled it, a planted draft restored, Save cleared the draft
- [x] On a fresh install pick a role, skip with "Set up later", then open Settings → Personalisation: the typed answers should be there as a draft


ahmadrosid:
## Lean review

No functional blocker found.

- The guided onboarding flow matches #869: answers populate `USER.md`, can be rerun, and supports “Set up later”.
- The inactive state for an empty document is correct.
- Tests cover the formatting and user-context behavior.

Minor points:

- `UserContextForm.tsx` is over 1,000 lines. Consider splitting the preset data from the form logic.
- Removing automatic `USER.md` creation on page open may affect installs that relied on that behavior. Please confirm this is intentional.
- This PR touches `SettingsPage.tsx`, which may conflict with PR #886.

**Verdict:** acceptable with minor cleanup.

yusupsupriyadi:
@ahmadrosid Thanks. All three points handled, pushed as b163d992, d45958ea, 546cb173.

**Preset split.** `UserContextForm.tsx` is 737 lines now, down from 1051. The role, work and reply-style tables moved to `apps/web/src/components/user-context-presets.ts` (352 lines) together with the pure string codecs that read them: `splitReplies`/`joinReplies`, `splitAlways`/`joinAlways`, `browserLanguageLabel`, `isPresetRole`. `WORK_BY_ROLE` and `GENERIC_WORK` stay private behind `workHintsForRole(role)`, so the form no longer does the table lookup and fallback itself. Placement follows `theme-options.ts` next to its component. The form file is now only the React sections.

Those codecs were the untested part of the change, so the move came with `user-context-presets.test.ts` (10 tests): both parsers round trip through their join counterparts, leading known words are consumed while free text is kept verbatim, and the generic work fallback is pinned.

**Automatic `USER.md` creation.** Intentional, and it was narrower than it looks. `autoInit` was only ever `true` from `SetupStepUserContext.tsx`; `UserContextSettings` defaulted it to `false`, so opening Settings never created the file. Only setup step 4 did, and only when no file existed yet, which means an existing install always took the `{ created: false }` path. The endpoint, `agent.initUserContext`, `client.initUserContext` and the CLI `/user init` are all untouched; the removed commit only dropped the React mutation wrapper that no longer had a caller.

The behavior worth losing: `initUserContext` stored `USER_CONTEXT_TEMPLATE`, which is a heading plus `- Label:` bullets with nothing after the colons. `normalizeUserContextContent` only trims, so that skeleton counted as content, `buildUserContextStatus` reported `active: true`, and `chat-prompt.ts:120` pushed a `# Personalisation (USER.md)` section of empty labels into the system prompt of every chat. Now the file is written when someone actually answers, and a missing file was already a supported state on that code path.

**#886.** Real overlap, and both are additive. Both branches insert into the same `Card` in `SettingsPage.tsx` right after the `ThemeToggle` row and before the `{version ? ...}` block, and both add imports in the same region, so whichever merges second gets a both-added conflict that resolves by keeping both rows. Both PRs are mine, so I will rebase the second one and post the result rather than leave it to the merge. Nothing in either change depends on the other.

Not rebased onto `main`: the 13 commits since this branch point touch none of the files in this PR (`git diff --name-only` intersection is empty) and the PR still reports `MERGEABLE` / `CLEAN`, so a rewrite would only churn the history.

Checks: `cd apps/web && tsc --noEmit` clean, `bun test` in `apps/web` 362 pass (10 new), `bun test packages/core/src/user-context.test.ts` 13 pass, `bun test packages/db/src/user-context.test.ts apps/server/src/http/user-context.test.ts` 3 pass, `bun run knip` clean, `bun x ultracite check` clean on the changed files.

ahmadrosid:
image
@yusupsupriyadi can you please hide the scrollbar here

ahmadrosid:
@yusupsupriyadi also please remove narrative ui, meaning too much description copy

yusupsupriyadi:
@ahmadrosid Both done, pushed as 150373c7 and 813a0c80.

image

**Scrollbar.** The dialog body is the scroller, so it got the `no-scrollbar` utility that already lives in `index.css` and is used by the sidebar, `KnowledgeTab` and `FilesPage`. No new CSS. It still scrolls: measured in the browser, `scrollHeight` 1304 against `clientHeight` 391 with a 0px gutter.

**Narrative copy.** Removed across the whole flow, not just the dialog:

- the dialog description under the title
- the four section subtitles, dropped from `USER_CONTEXT_SECTIONS` itself so the setup step trims with it
- the blurb under each of the 11 role cards and under "Something else"
- field hints: name, current projects, tech stack (the per-role `stackHint`), Always, Never, plus the "Saved as an Always rule" footnote
- the review step, now just the textarea with no intro paragraph and no footnote

Length and Tone were two-column cards only to carry a sample sentence, so with those gone they render as chips like the Reply language row above them. Placeholders stayed, they guide the answer without narrating. The one line I did keep is the settings row subtitle under "Personalisation", since every sibling row on that page has one and it would read as broken without it.

Net 28 insertions, 123 deletions across 4 files.

Checks: `cd apps/web && tsc --noEmit` clean, `bun test` in `apps/web` 362 pass, `bun run knip` clean, `bun x ultracite check` clean on the changed files. Also walked a fresh install in the browser: setup step 4 and Settings -> Personalisation, scrollbar gone in both light and dark, the 12 role cards now fit the space that used to hold 6.

yusupsupriyadi:
Rebased onto `main` after #886 landed, force-pushed to `9a8074b8`. `MERGEABLE` / `CLEAN` again, all three checks green.

As expected the only conflict was `SettingsPage.tsx`, and only the row block: both branches insert after the `ThemeToggle` row and before `{version ? ...}`. Resolved by keeping both, Personalisation first and then the "Token usage in chat" toggle. The imports merged on their own. Against `main` this PR now touches that file with exactly one import and one line:

```diff
+import { UserContextSettings } from "@/components/UserContextCard";
...
+          <UserContextSettings />
```

Nothing else intersected: the 24 commits that landed since the old base touch none of the other files here.

Checks after the rebase: `cd apps/web && tsc --noEmit` clean, `bun test` in `apps/web` 385 pass (was 362, the extra 23 come from #886), `bun test` on the user-context tests in core, db and server 16 pass, `bun run knip` clean, `bun x ultracite check` clean on the 10 changed files. Walked a fresh install in the browser too: Settings shows Appearance, Personalisation, Token usage in chat, Version in one card with no spacing or divider oddity, and the dialog still measures `scrollHeight` 1304 against `clientHeight` 576 with a 0px gutter.

ahmadrosid:
@yusupsupriyadi we have personalization step in setup page can you bring this there as well?


image

packages/core/src/user-context.ts
```diff
@@ -1,14 +1,128 @@
...
+  const extra: string[] = [];
+
+  for (const line of (content ?? "").split("\n")) {
+    const prefix = USER_CONTEXT_FIELDS.find((field) =>
+      line.startsWith(bulletPrefix(field))
+    );
+
+    if (prefix) {
+      // Drop the single separating space, so a trailing space survives a round trip.
+      answers[prefix.key] = line
```
fajarhide, line 86: A repeated bullet loses the earlier line. The key is overwritten here and the first line never reaches `extra`, so it is gone on the next save. USER.md is hand-editable, and two `Always` rules on two lines is a reasonable thing for someone to write:

```
in       : - Always: Answer in English
           - Always: Use metric units
parsed   : {"answers":{"always":"Use metric units"},"extra":""}
rendered : - Always: Use metric units
```

Pushing the line into `extra` when the key is already set keeps it, at the cost of it moving below the bullets. Free text under its own heading round-trips byte for byte, I checked that separately and it passes.

fajarhide (commented):
@yusupsupriyadi I ran this one rather than read it. Fresh config dir, its own server and Vite, walked the wizard from an empty install to the chat page, then reopened Settings. The step reads well and the flow is the one #869 asked for: name pre-filled from the account, blank answers left out of the file entirely, the exact Markdown shown before saving, and the same form reachable again from Settings with the saved answers in place. Twenty seconds of it:

https://github.com/user-attachments/assets/7b589794-71a2-47c9-9820-bb967755daec

One thing to fix before merge, inline. Two smaller notes: the subtitle carries an em dash (`Help the agent understand your preferences - optional`), and the Timezone control on the last panel is not part of USER.md, so it reads as a second concern riding along.

Not checked: the "Set up later" draft path and a seeded install (`NAKAMA_SEED_ADMIN_*`), and no test of mine touches the React components, since `apps/web` has no DOM harness.

packages/core/src/user-context.ts
yusupsupriyadi: @fajarhide Fixed in 1e9cfd09. `parseUserContext` now takes the first bullet per key and lets a repeat fall through to `extra`, exactly as you described:

```
in       : - Always: Answer in English
           - Always: Use metric units
parsed   : {"answers":{"always":"Answer in English"},"extra":"- Always: Use metric units"}
rendered : - Always: Answer in English
           (blank)
           - Always: Use metric units
```

The regression test also parses and renders the result a second time, since the failure you hit was on the *next* save: the moved line stays put from there on, byte for byte.

Two details from doing it:

- Used `Object.hasOwn(answers, key)` rather than `key in answers`. Same length, and `answers` starts as `{}`, so `in` would answer true for prototype keys. None of the seven field keys collide today, but this way it cannot.
- Side effect I like: a duplicate now counts as `extra`, so `UserContextForm.hasExtra` auto-opens the raw Markdown review in the settings dialog. The user sees the line that moved instead of wondering where it went.

yusupsupriyadi:
@fajarhide Thanks for taking it for a walk from an empty install rather than reading the diff.

The bullet-collision fix is in 1e9cfd09, replied inline. The two smaller notes:

**Em dash** (aecd5503). Gone: `Help the agent understand your preferences. Optional.`

**Timezone.** You are right that it is a second concern, but I kept the control and labeled it instead of moving it out, because dropping it would quietly cost something. The select only persists on change: it shows `Intl.DateTimeFormat().resolvedOptions().timeZone` but never saves it, so an install where nobody touches it stays on `DEFAULT_TIMEZONE`, which is `UTC` (`packages/core/src/user-config.ts:77`, read through `loadUserTimezone` in `agent-service.getUserTimezone`). Step 4 is the only place in setup where that can be corrected, and automations schedule against it. The card now says so:

> Timezone
> Saved on its own, not part of USER.md. Used for scheduled automations and local time awareness.

If you would rather step 4 be USER.md and nothing else, the clean version is to persist the detected zone during setup and drop the control, but that is a behavior change beyond this PR and I would rather do it separately. Say which you prefer and I will follow.

Checks on the new head: `bun test packages/core/src/user-context.test.ts` 14 pass, the db/server/preset tests 13 pass, `tsc --noEmit` in `apps/web` clean, `ultracite check` on the four changed files clean, `knip` clean.

Not checked yet, as you flagged: a seeded install (`NAKAMA_SEED_ADMIN_*`).

yusupsupriyadi:
@fajarhide Ran the two paths you flagged as unchecked, on the current head, so there is evidence for them and not just a test-plan checkbox.

Throwaway instance (its own HOME, server on 4399, Vite on 3099), fresh DB, walked account -> org -> provider -> step 4:

**Draft path.** Picked a role, hit "Set up later". `localStorage` holds

```
nakama:user-context-draft:org_c1a3…  =>  "# About Me\n\n- Name / nickname: Alex Draft\n- What you do: Software engineer\n"
```

and Settings -> Personalisation reopens with the name filled in and the `Software engineer` card `aria-pressed="true"`. Nothing was written to `org_members.user_context` on the skip.

**The bullet-collision fix, through the UI rather than the unit test.** Saved a hand-written file with two `Always` rules via `PUT /v1/user/context`, reopened the dialog (the raw Markdown review auto-opened, holding both lines), edited an unrelated field and hit Save. Read back:

```
# About Me

- Current projects: Billing API split
- Always: Answer in English

- Always: Use metric units
```

Both rules still there. On the old code that save is where the first one disappeared.

Still not checked by anyone: a seeded install (`NAKAMA_SEED_ADMIN_*`).

yusupsupriyadi:
@ahmadrosid This is already in, since 469a3cb9 (2026-09-07), a day before your screenshot.

The card in your screenshot, `Personalisation / USER.md - personalisation for this org` with an `Edit` button next to a Timezone card, is what `main` renders on `/setup` step 4: the settings card dropped into the wizard, so the questions sit behind a dialog. This PR replaces it with the guided flow inline, one question per screen, same as the Settings dialog uses.

Walked it again just now on the current head (`aecd5503`, fresh install, empty DB). Step 4 is four panels with progress dots, `Back` / `Set up later` / `Next`, and `Finish` on the last one:

1. **What do you do?** Name (pre-filled from the account you just created) plus twelve role cards
2. **What are you working on?** Current projects, tech stack, with role-aware stack quick picks
3. **How should replies sound?** Reply language, then length and tone cards with sample sentences
4. **What Nakama will remember** The exact Markdown, editable, before anything is saved

Panel 4 with a role, a project, `Go` and `Concise` picked renders:

```
# About Me

- Name / nickname: Alex
- What you do: Software engineer
- Current projects: Moving billing off the monolith
- Tech stack: Go
- How you like replies (concise, detailed, casual, formal): concise
- Always: Reply in English.
```

`Finish` writes that through `PUT /v1/user/context`; `Set up later` keeps the typed answers as a local draft that Settings -> Personalisation picks up. Nothing is written on a skip.

I have screenshots of all four panels if the text is not enough, say so and I will post them.

Two things from fajarhide's review also landed since your screenshot: the subtitle now reads `Help the agent understand your preferences. Optional.` without the em dash, and the Timezone card says it is saved on its own and is not part of USER.md.

packages/core/src/user-context.ts
fajarhide: Checks out. I ran the reported case plus four more on `aecd5503`, each parsed and rendered twice:

```
two Always rules    -> both kept, second render identical to the first
three repeats       -> `a` stays the answer, `b` and `c` sit in extra
Always and Never x2 -> both repeats kept
empty first bullet  -> the empty one drops, the rule survives
free text heading   -> byte for byte
```

The regression test earns its place too: with `!Object.hasOwn(answers, prefix.key)` reverted it fails with `Expected: "Answer in English"`, `Received: "Use metric units"`. 24 pass across `user-context.test.ts` and `user-context-presets.test.ts` on that head.

`Object.hasOwn` over `in` is the right instinct even though no field key collides today.

fajarhide:
@yusupsupriyadi Timezone: keep it as you have it. Your argument is the one that decides it, an install where nobody touches the control stays on `UTC` and automations schedule against that, so removing the control costs more than the tidiness buys. The labeled card is the honest version. The cleaner shape, persisting the detected zone and dropping the control, is worth its own issue so it does not get lost.

The bullet fix is verified, inline. Nothing else from my side.

@ahmadrosid on the setup step: this is already there, as Yusup says. When I walked a fresh install on `9a8074b8` the guided questions were step 4 itself, not the settings card plus a dialog, and the video in my review above is that walk.

apps/web/src/components/UserContextCard.tsx
```diff
@@ -234,316 +165,75 @@ export function UserContextEditorDialog({
...
+        ? "USER.md is set for this org"
+        : "Answer a few questions so replies fit how you work"}
+    </p>
   );
 }
 
-function useUserContextSettings(autoInit: boolean) {
+/** USER.md row for settings and the setup wizard. Render inside a parent card. */
+export function UserContextSettings({
+  onSaveSuccess,
```
ahmadrosid, line 197: Neither SettingsPage nor SidebarUserMenu supplies onSaveSuccess. You can drop this callback from both components, including UserContextSettingsProps, the forwarding prop, and the optional call after saving. The write mutation already refreshes the user-context query, so the settings status still updates without it.

ahmadrosid (commented):
One small cleanup in the settings/dialog plumbing.

ahmadrosid:
Let's merge this after that 1 simple cleanup. LGTM 👍

apps/web/src/components/UserContextCard.tsx
yusupsupriyadi: @ahmadrosid Done in 00626570. Both props gone: `UserContextSettingsProps` deleted entirely since it only held that one prop, so `UserContextSettings` now takes none, plus the forwarding prop on the dialog, the field on `UserContextEditorDialogProps`, its destructuring, and the call in `handleSave`. Net 2 insertions, 15 deletions in one file.

Checked your reasoning before cutting rather than after: `SettingsPage.tsx:66` renders `<UserContextSettings />` bare, `SidebarUserMenu.tsx:198` passes only `open` and `onOpenChange`, and `useWriteUserContextMutation` invalidates `queryKeys.userContext` in `onSuccess`, so the settings row still flips to "USER.md is set for this org" after a save.

Left `await refetch()` in `handleSave` where it is, since that is a separate question from the callback.

The branch was a few commits behind, so this sits on top of the `main` merge that moved the UI primitives into `packages/ui`. Checks on the new head: `tsc --noEmit` in `apps/web` clean, `bun test` in `apps/web` 463 pass, `ultracite check` clean on the changed file, `knip` clean.

PR #886, opened 2026-09-07
feat(web): show per-reply token usage and cost in chat

### LINKED ISSUE #885: feat(web): show per-reply token usage and estimated cost in chat, behind a settings toggle
### Summary

Chat replies do not show what they cost. Usage is already recorded per LLM call and surfaced as an org-wide running total on the Status page, but nothing reaches the chat where the spending happens. Show tokens and estimated USD under each assistant reply, a session total in the composer toolbar, and a Settings toggle to hide both.

### Why it matters

Today the only readout is the Status page total, which mixes every profile, channel and automation. You cannot tell which reply was expensive, whether a long tool loop just burned 40k input tokens, or whether switching models actually helped, and the answer arrives on a different page from the decision. The numbers already exist: `wrapProviderWithUsageTracking` resolves input/output tokens per call and `estimateUsageCostUsd` already prices them. They are simply never sent to the client.

A second gap blocks the cost half for BYO endpoints. `CustomModelEntry` already carries `inputPerMillionUsd` and `outputPerMillionUsd`, `getCustomModelPricing` already reads them, and `validateCustomModels` already enforces them, but the dashboard passes `showPricing={false}` for OpenAI-compatible providers. So those users have no way to enter rates and would see tokens with no cost, forever.

### Preview

Image

Image

- `packages/core/src/contract.ts` — a `ChatUsage` shape carried on the assistant message, plus a `usage` stream event
- `packages/agent/src/chat.ts` — emit usage once per LLM call in the tool loop, and keep it on the history message
- `apps/server/src/http/shared.ts`, `packages/client/src/stream.ts` — SSE passthrough
- `apps/web/src/components/chat/`, `apps/web/src/pages/SettingsPage.tsx` — reply footer, composer chip, toggle
- `apps/web/src/components/CustomProviderFields.tsx` — unhide the `$/1M` columns that `ModelListEditor` already renders

### Acceptance criteria

- [x] a completed assistant reply shows input tokens, output tokens and estimated USD
- [x] a turn that ran tools reports one total for the whole turn, not one figure per LLM call
- [x] reopening a stored session still shows usage for the replies that recorded it
- [x] cost is omitted, not shown as `$0.00`, when the model has no known pricing
- [x] token counts the provider did not report are marked as approximate
- [x] a Settings toggle hides both the reply footer and the composer total
- [x] OpenAI-compatible custom models accept `$/1M` input and output rates in the dashboard
- [x] setting only one of the two rates is rejected in the browser with the same message the server already returns

yusupsupriyadi:
## Summary

Preview:
Screenshot_21

On settings:
Screenshot_22

Send a chat message → the reply footer says what it cost, and the composer carries the session total.

**Before:** tokens were resolved per LLM call server-side, then surfaced only as one org-wide total on the Status page, and OpenAI-compatible models had no way to enter `$/1M` rates at all.
**After:** a `ChatUsage` rides the assistant message and a `usage` stream event, so a turn reports `1.2k in · 56 out · $0.0042`, and custom providers can set the rates that price it.

Why safe:
1. Reuses the tokens `wrapProviderWithUsageTracking` already resolved and the prices `estimateUsageCostUsd` already computed, so no extra provider calls
2. `usage` is optional end to end, so stored sessions and other clients simply render no footer
3. `record()` returns null for a model with no known pricing, so cost is omitted rather than claimed as `$0.00`

Only revisit if a provider starts reporting usage per stream chunk instead of per call, which would double-count a turn total.

Happy to split the `$/1M` commit into its own PR if you would rather review it alone; it is here because the cost half of the feature is dead for BYO endpoints without it.

## Test plan
- [x] `bun test packages/agent/src packages/client/src apps/web/src/lib` (413 pass)
- [x] `bun test apps/server/src/providers/usage-tracking.test.ts apps/server/src/services/llm-usage-tracker.test.ts` (9 pass)
- [x] `bun run knip` (no findings)
- [x] Send one message that runs a tool, reload the session, then flip Settings → Token usage in chat — one total per turn, and the toggle hides both readouts

Fixes #885


ahmadrosid:
## Lean review

### Medium: fallback pricing can show an unverified cost

The UI treats `getModelPricing(...) !== null` as known pricing, but the pricing layer returns default pricing for models that are not explicitly priced. The chat can therefore show a dollar amount that is only an estimate based on a generic fallback, while the PR claims cost is omitted when pricing is unknown.

Please distinguish explicit catalog pricing from fallback pricing, or omit the dollar amount when the model has no verified rates.

### Low: duplicate DOM id

`chat-message-list.tsx` renders `id="chat-turn-usage"` for every usage badge. This produces duplicate IDs in the document. Use a class or a unique ID only where an ID is required.

Also take care when merging with PR #887 because both modify the same chat files.

**Verdict:** request changes before treating cost output as authoritative.

yusupsupriyadi:
@ahmadrosid Good catch, that gate was doing two jobs. Fixed in 815da5ad, 07140932, 47c15917, ef0836a8, f5f19d3b.

**Fallback pricing shown as cost.** `getModelPricing` returned two different kinds of answer through one value: rates published for the model, and `DEFAULT_PRICING` ($1/$3 per 1M) as a last resort. `hasCatalogPricing` is `!== null` over that, so it could not tell them apart, and `record()` passed the fallback through as `costUsd`.

The two answers are separate now. `getExplicitModelPricing` returns published rates only (the Images table, a catalog entry, or rates the user typed for a custom model) and null where the fallback would apply; `getModelPricing` is that plus the fallback, so its own behavior is unchanged. `LlmUsageTracker.record` returns null unless the rates were published, so the badge shows tokens with no dollar amount, and `chatUsageTitle` already reads "cost unknown (no pricing for this model)" there. Running totals and `StatusPage` keep the fallback, which is what they were built on.

Scope of that fallback, worth recording: all 43 catalog models carry explicit rates, so it only fires for a model id the catalog does not know, such as a hand-typed id or a model newer than the catalog. `gpt-4o` is one of those, which makes it the natural regression case: it goes through the totals and comes back without a cost, while the priced case uses `claude-sonnet-4-6`.

**Duplicate id.** Gone. `AssistantMessageActions` renders once per turn, so `chat-turn-usage` appeared once per reply. Nothing referenced it, so the `id` prop is off `ChatUsageBadge` entirely rather than growing a unique suffix. `chat-session-usage` on the composer chip stays; that one is a single instance.

**#887.** Checked rather than assumed: `git merge-tree --write-tree feat/chat-token-usage feat/edit-user-message` merges clean. The two branches share `chat-message-list.tsx`, `chat-history.ts`, `chat-history.test.ts` and `chat-page-content.tsx` but touch different parts of each, and both merge clean against current `main`. The one collision in my open set is #882 against this branch in `SettingsPage.tsx`, flagged there; I will rebase whichever lands second.

Checks: `bun test` in `apps/web` 359 pass, `apps/server` pricing + usage-tracking + image-generation 32 pass, `cd apps/web && tsc --noEmit` clean, `bun run knip` clean, `bun x ultracite check` clean on the changed files. Four `composio-tool-bridge` failures in my local full-server run are environmental (they read a local `web_public_url`) and reproduce on `main`.

ahmadrosid:
@yusupsupriyadi i think this section make the ui bloated
image
please just put it under this section like in the cursor
image

like when you click this circle context 

image

yusupsupriyadi:
@ahmadrosid Moved in 5d8219ae.

The chip is gone from the composer. The session total now sits under the context line in the ring's tooltip, so the circle answers both questions at once:

image

The toolbar is back to ring + model + effort, so nothing pushes the model select along any more. The per-reply footer under each answer is unchanged, and the Settings toggle still hides both readouts (with it off the tooltip shows only the context line).

Not a full copy of the Cursor panel: theirs breaks the window down by system prompt, tool definitions, rules and skills, and the server here reports one used-tokens figure, so that panel would have a single row in it. If you would rather it open on click than on hover, say so and I will swap the trigger for a popover.

Checks: `bun test apps/web/src` 359 pass, `tsc --noEmit` clean, `bun run knip` clean, `ultracite check` clean. Also driven in a browser against a throwaway server instance: one turn on `gemini-2.5-flash`, reload, the session total comes back from stored usage; toggle off leaves the context line alone.

PR #887, opened 2026-09-07
feat(web): edit a sent user message and regenerate the reply

### LINKED ISSUE #883: feat(web): edit a sent user message and regenerate the reply from it
### Summary

Add an **Edit** action to sent user messages in the web chat. Clicking it turns the bubble into an inline textarea; submitting sends the edited text in place of that turn and the assistant answers it fresh. Everything after the edited message drops out of the visible conversation, the way ChatGPT and Gemini handle "edit message".

The assistant side already has the other half of this. `AssistantMessageActions` (`apps/web/src/components/chat/chat-message-list.tsx:492`) exposes **Try again**, and `handleTryAgainMessage` (`apps/web/src/pages/chat/use-chat-page.ts:959`) branches the session at the checkpoint before the prompt via `POST /v1/sessions/:sessionId/branch`, then re-sends the original prompt into the branch. Edit-and-resend is that same flow with a different prompt string, so it should reuse that path instead of adding a new one.

Measured on `cdb9a97f` (`main`).

### Why it matters

Once a message is sent there is no way to fix it. A typo, a missing constraint, or a wrong file name means either living with a bad answer or copying the text back into the composer by hand, which also keeps the wrong turn in the history the model keeps seeing. **Try again** only helps when the prompt was right and the answer was wrong. When the prompt was wrong, the user bubble (`ChatMessageRow` at line 430, `UserMessageContent` at line 611) has no action at all. Every mainstream chat client ships this, and the branch endpoint that makes it cheap already exists here.

- `apps/web/src/components/chat/chat-message-list.tsx`: `ChatMessageRow` gains a hover/focus **Edit** control styled like `AssistantMessageActions`, an inline edit mode (textarea with Send / Cancel, Esc cancels, Ctrl/Cmd+Enter submits), and a new `onEditMessage?: (message: ChatListItem, text: string) => void` prop threaded through `ChatMessageListProps`.
- `apps/web/src/pages/chat/use-chat-page.ts`: a `handleEditMessage` next to `handleTryAgainMessage` that resolves the checkpoint with `findRetryCheckpoint`, branches (or creates a fresh session when the edited message is the first turn), then calls `sendMessage(editedText, [], { initialMessages, sessionOverride })`. The body of `handleTryAgainMessage` becomes a shared helper that takes the prompt text.
- `apps/web/src/pages/chat/chat-page-content.tsx:163`: wire `onEditMessage` beside `onRetryMessage`.
- `apps/web/src/pages/chat/chat-page.shared.test.ts`: cover any new pure helper.
- No server change expected: `agent.branchSession` (`apps/server/src/services/agent-service.ts:1785`) already copies history up to `messageIndex`.

### Acceptance criteria

- [ ] Hovering or focusing a sent user message in a web session shows an **Edit** action; it is hidden or disabled when `actionsDisabled` is set (`busy` or read-only channel session).
- [ ] Edit opens the message text inline in a textarea pre-filled with the original. **Cancel** or Esc restores the bubble unchanged.
- [ ] Submitting with empty or unchanged text is a no-op (Send disabled).
- [ ] Submitting changed text: the conversation visibly truncates to the messages before the edited one, the edited text appears as the new last user message, and the assistant streams a fresh reply to it.
- [ ] The original session is untouched. The edited conversation continues in a branch session (title suffixed `(Branch)`), consistent with **Try again**, and the URL switches to the branch.
- [ ] Editing the first message of a session works (no checkpoint: new session instead of branch).
- [ ] Edit is disabled while a turn is streaming and while a branch/retry is in flight (`branchingMessageId`).
- [ ] Messages with image or document attachments show Edit disabled with the same text-only hint **Try again** uses, until attachment re-send is supported.
- [ ] Existing **Try again**, **Branch**, and **Copy** actions keep working. `bun test`, `tsc --noEmit`, and `bun run knip` pass.

Non-goals for the first pass: a version switcher across edits (ChatGPT's `< 1/2 >`), in-place truncation of the original session, and editing messages that carry attachments.

---

I am working on this and will open the PR against this issue.

yusupsupriyadi:
## Summary

Fix a typo in a sent prompt and the assistant answers the corrected one, no retyping.

**Before:** a sent user message had no action at all; correcting it meant copying the text back into the composer, and the wrong turn stayed in the history the model keeps reading.
**After:** hover a sent message, edit it inline, send. `branchAndSendPrompt` branches at the checkpoint before it, so the thread truncates there and the reply is regenerated.

Why safe:
1. Reuses the Try again path end to end. No server change: `agent.branchSession` already slices history inclusively.
2. The original session is never mutated; the edit continues in a branch, so a mistaken edit destroys nothing.
3. Edit is offered only on stored, text-only, non-failed user rows (`isEditableUserMessage`), and the shared helper now refuses to branch a read-only session, which also hardens Try again.

Only reconsider if the extra session per edit makes the chat list unwieldy.

Fixes #883

## Test plan
- [x] `bun test apps/web/src` (359 pass)
- [x] `tsc --noEmit` in `apps/web` (clean)
- [x] `bun run knip` (clean)
- [x] Hover a sent message, edit it, send: the thread truncates to that point, the reply is regenerated, and the original chat is still in the sidebar.

### DIFF
--- a/apps/web/src/components/chat/chat-message-list.tsx
+++ b/apps/web/src/components/chat/chat-message-list.tsx
@@ -1,6 +1,7 @@
 import {
   CheckmarkCircle01Icon,
   Copy01Icon,
+  Edit03Icon,
   File01Icon,
   GitBranchIcon,
   MoreHorizontalIcon,
@@ -27,14 +28,20 @@ import { AssistantTurnSegmentView } from "@/components/chat/assistant-tool-group
 import { segmentAssistantTurn } from "@/components/chat/assistant-tool-group.shared";
 import { ImageAttachmentPreview } from "@/components/chat/image-attachment-preview";
 import { TextAttachmentPreview } from "@/components/chat/text-attachment-preview";
+import { Button } from "@/components/ui/button";
 import {
   DropdownMenu,
   DropdownMenuContent,
   DropdownMenuItem,
   DropdownMenuTrigger,
 } from "@/components/ui/dropdown-menu";
+import { Textarea } from "@/components/ui/textarea";
 import { extractTurnArtifacts } from "@/lib/chat-artifacts";
-import { type ChatListItem, formatSessionTimestamp } from "@/lib/chat-history";
+import {
+  type ChatListItem,
+  formatSessionTimestamp,
+  isEditableUserMessage,
+} from "@/lib/chat-history";
 import {
   followOutputBehavior,
   listOverflowsViewport,
@@ -98,6 +105,7 @@ interface ChatMessageListProps {
   messages: ChatListItem[];
   modelLabel?: string | null;
   onBranchMessage?: (message: ChatListItem) => void;
+  onEditMessage?: (message: ChatListItem, text: string) => void;
   onRetryMessage?: (message: ChatListItem) => void;
   profileId?: string | null;
   showThinking?: boolean;
@@ -121,6 +129,7 @@ function ChatMessageListSession({
   streamActive = false,
   turnStartedAt = null,
   onBranchMessage,
+  onEditMessage,
   onRetryMessage,
   emptyMessage,
   className,
@@ -238,7 +247,12 @@ function ChatMessageListSession({
       if (turn.kind === "user") {
         return (
           <div className={itemClassName}>
-            <ChatMessageRow message={turn.message} />
+            <ChatMessageRow
+              busy={branchingMessageId != null || streamActive}
+              disabled={actionsDisabled}
+              message={turn.message}
+              onEditMessage={onEditMessage}
+            />
           </div>
         );
       }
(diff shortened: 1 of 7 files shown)


ahmadrosid:
## Lean review

The implementation correctly reuses the Try again branch path and keeps the original session unchanged.

### Medium: missing behavioral test

Current tests cover `isEditableUserMessage`, but not the full edit flow:

`edit → find checkpoint → branch session → send edited text`

Please add a regression test for that path. The manual test in the PR is still unchecked.

### Low: defense-in-depth guard

`handleEditMessage` checks `images` and `documents`, but not `imageAttachments`. The UI currently prevents this path, but the handler should enforce the same eligibility rules itself.

This PR shares chat files with PR #886, so merge conflicts need care.

**Verdict:** implementation looks sound; add the behavioral regression test.

yusupsupriyadi:
@ahmadrosid Both handled, pushed as bbd99916, dceb72b2, f78c8f54, 7ab7d1c2.

**Behavioral test.** The decisions on that path all lived inside `useChatPage`, and `apps/web` has no harness for hooks or components: no `.test.tsx` anywhere, and no `@testing-library/react`, jsdom or happy-dom in the repo. Rather than add a DOM stack for one test, I moved the decisions into pure functions in `chat-page.shared.ts`, where the sibling helpers of this flow already live and already have a test file:

- `planPromptBranch(messages, prompt)` returns the history index to cut at and the rows the branch carries, or null when nothing precedes the prompt and the flow should open a fresh session instead. `findRetryCheckpoint` is private to it now.
- `editedPromptText(message, text)` returns the trimmed edit, or null when the edit changes nothing.

`branchAndSendPrompt` reads both off the plan instead of recomputing the carried rows and asserting non-null on the checkpoint index twice. The tests walk the path on a two-turn transcript: editing the second prompt cuts the branch at index 1, carries exactly the rows already in history, sends the trimmed edit rather than the original, returns null for an unchanged or whitespace-only edit, and falls back to a fresh session for the opening prompt or a prompt not yet in history. Try again shares the same plan, so it rides along.

That leaves the two awaits as passthroughs: `branchSessionMutation.mutateAsync` and `sendMessage` take `plan.messageIndex`, `plan.initialMessages` and `nextText` straight from the plan. Happy to add the DOM stack and drive the whole hook if you would rather have that.

**Eligibility guard.** `handleEditMessage` routes through `isEditableUserMessage`, the same predicate the list uses to decide whether to offer Edit. One rule in one place, so `imageAttachments` is covered along with the `historyIndex`, failed-turn and questionnaire rules, instead of a subset living in the handler.

**#886.** Checked rather than assumed: `git merge-tree --write-tree feat/chat-token-usage feat/edit-user-message` merges clean. The shared files (`chat-message-list.tsx`, `chat-history.ts`, `chat-history.test.ts`, `chat-page-content.tsx`) are touched in different places, and both branches merge clean against current `main`.

Checks: `bun test` in `apps/web` 365 pass (6 new), `cd apps/web && tsc --noEmit` clean, `bun run knip` clean (it flagged `findRetryCheckpoint` as an unused export once it became internal, hence the unexport), `bun x ultracite check` clean on the changed files. Manual probe in the body is next: I will run it in the UI harness and tick it.

Issue #781, opened 2026-08-30
web: finish window.confirm → Dialog migration and ban native alerts


ahmadrosid:
## Summary

Do two things so `window.confirm` cannot come back. Live calls are already gone.

**Before:** deletes use `Dialog`, but `noAlert` is `"off"` and a stale comment still names `window.confirm`.
**After:** that comment is gone, and `bun x ultracite check` fails if anyone adds `window.confirm`, `alert`, or `prompt`.

About 15 minutes.

## Why it matters

`window.confirm` does not run while a Radix `Dialog` is open (see SkillAssignPicker). It also ignores the app theme. The Dialog pattern already exists — do not add a shared ConfirmDialog.

## Do this

1. Delete the comment at `apps/web/src/pages/profiles/use-profiles-page.ts:802`.
2. In `biome.jsonc`, set `linter.rules.suspicious.noAlert` from `"off"` to `"error"`.
3. If that flags a real `alert` / `confirm` / `prompt`, replace it with the Dialog in `apps/web/src/pages/files/files-delete-dialog.tsx`.
4. Run `bun x ultracite check`.

## Acceptance criteria

- [ ] `git grep -n window.confirm` is empty
- [ ] `biome.jsonc` no longer sets `noAlert` to `"off"`
- [ ] `bun x ultracite check` passes

Thushara3005:
Hi @ahmadrosid ,
I’m a first-time contributor and I’d like to work on this issue. Could you please assign it to me?

ahmadrosid:
Done assigned to you @Thushara3005

AlexMuhammad:
give it to me @ahmadrosid @fajarhide @yusupsupriyadi, i'll fix now!

PR #892, opened 2026-09-07
fix(web): confirm before removing a Telegram allowed user

### LINKED ISSUE #890: web: removing a Telegram allowed user saves immediately with no confirmation
## Summary

Removing a Telegram allowed user saves to the server on one click, with no confirmation. `TelegramAllowedUsersDialog.tsx:226` calls `removeAllowedUserId`, which is `saveAllowedUsers(allowedUsers.filter(...))` at `:137`, and `saveAllowedUsers` runs `saveMutation.mutate` immediately at `:60`.

This is the one site left over from #520. The other six named there now go through a confirmation target: the three profile sections pass `onRemove={setRemoveConfirm}`, automations pass `onDelete={setDeleteTarget}` and `onDeleteRun={setDeleteRunTarget}`, and `DiscordAllowedUsersDialog` and `ModelListEditor` only edit local state, so their remove buttons no longer touch the server at all.

Measured on `60ce1bbf`.

## Why it matters

The Discord dialog beside it stages the change and persists on Save. Telegram persists on the click, so a misclick on the remove icon revokes a user's access with nothing to undo it and no dialog in between. The two dialogs look identical to the user and behave differently.

- `apps/web/src/components/TelegramAllowedUsersDialog.tsx:137` and `:226`
- The Dialog pattern to copy is `apps/web/src/pages/files/files-delete-dialog.tsx`, per #781
- Staging then saving, like `DiscordAllowedUsersDialog.tsx:89`, is the other option and is a smaller diff

## Acceptance criteria

- [ ] Clicking remove does not call `saveMutation` until the user confirms, or does not call it at all until Save
- [ ] Cancelling leaves the allowed-user list unchanged, in the UI and on the server
- [ ] Discord and Telegram dialogs behave the same way as each other
- [ ] A test covers the cancel path

fajarhide:
## Summary

Removing a Telegram allowed user now asks first.

**Before:** the remove icon called `removeAllowedUserId`, which called `saveAllowedUsers`, which ran `saveMutation.mutate` on the click. One misclick revoked a user's access with no dialog and nothing to undo it.

**After:** the click sets a `removeTarget` and a confirm dialog decides. Cancel leaves the list untouched, in the UI and on the server.

Why safe:
1. The confirm is shaped like the import dialog already in this file, so no shared ConfirmDialog is added. #781 asked for exactly that restraint.
2. Add and import still save on the action. They are additive and undoable by removing the entry; this is about the destructive one.
3. `DiscordAllowedUsersDialog` stages and saves on Save, so the two dialogs now agree that a remove is not a one-click server write.

Not done: the cancel-path test the issue asked for. `apps/web` has no component test runner, no `.test.tsx` anywhere and no testing-library dependency, and this file's own `TelegramAllowedUsersDialog.test.ts` tests the parser rather than the component. Adding a rendering stack for one dialog is out of proportion; say the word if you want it and I will open that separately.

## Test plan
- [x] `bun x ultracite check` on the changed file: clean
- [ ] Manual: Settings, Telegram, remove a user, cancel, confirm

Closes #890

### DIFF
--- a/apps/web/src/components/TelegramAllowedUsersDialog.tsx
+++ b/apps/web/src/components/TelegramAllowedUsersDialog.tsx
@@ -49,6 +49,9 @@ export function TelegramAllowedUsersDialog({
   const [importDraft, setImportDraft] = useState("");
   const [importError, setImportError] = useState<string | null>(null);
   const [formError, setFormError] = useState<string | null>(null);
+  const [removeTarget, setRemoveTarget] = useState<AllowedTelegramUser | null>(
+    null
+  );
 
   function saveAllowedUsers(
     nextUsers: AllowedTelegramUser[],
@@ -134,8 +137,16 @@ export function TelegramAllowedUsersDialog({
     });
   }
 
-  function removeAllowedUserId(id: string) {
-    saveAllowedUsers(allowedUsers.filter((entry) => entry.id !== id));
+  function confirmRemoveAllowedUser() {
+    if (!removeTarget) {
+      return;
+    }
+
+    const id = removeTarget.id;
+    saveAllowedUsers(
+      allowedUsers.filter((entry) => entry.id !== id),
+      () => setRemoveTarget(null)
+    );
   }
 
   return (
@@ -223,7 +234,7 @@ export function TelegramAllowedUsersDialog({
                     <Button
                       aria-label={`Remove Telegram user ID ${user.id}`}
                       disabled={saveMutation.isPending}
-                      onClick={() => removeAllowedUserId(user.id)}
+                      onClick={() => setRemoveTarget(user)}
                       size="icon-sm"
                       type="button"
                       variant="ghost"
@@ -307,6 +318,41 @@ export function TelegramAllowedUsersDialog({
               type="button"
             >
               Add user
+            </Button>
+          </DialogFooter>
+        </DialogContent>
+      </Dialog>
+
+      <Dialog
+        onOpenChange={(next) => !next && setRemoveTarget(null)}
+        open={removeTarget !== null}
+      >
+        <DialogContent className="gap-5 p-6 sm:max-w-lg">
+          <DialogHeader className="gap-2">
+            <DialogTitle>Remove Telegram user</DialogTitle>
+            <DialogDescription>
+              {removeTarget?.username
+                ? `@${removeTarget.username} (${removeTarget.id})`
+                : removeTarget?.id}{" "}
+              loses access to this profile as soon as this is saved.
+            </DialogDescription>
+          </DialogHeader>
+
+          <DialogFooter className="gap-3 border-t-0 bg-transparent p-0 sm:justify-end">
+            <Button
+              onClick={() => setRemoveTarget(null)}
+              type="button"
+              variant="outline"
+            >
+              Cancel
+            </Button>
+            <Button
+              disabled={saveMutation.isPending}
+              onClick={confirmRemoveAllowedUser}
+              type="button"
+              variant="destructive"
+            >
+              Remove
             </Button>
           </DialogFooter>
         </DialogContent>
--- a/apps/web/src/components/TelegramAllowedUsersDialog.tsx
+++ b/apps/web/src/components/TelegramAllowedUsersDialog.tsx
@@ -57,6 +57,10 @@ export function TelegramAllowedUsersDialog({
     nextUsers: AllowedTelegramUser[],
     afterSuccess?: () => void
   ) {
+    // Optimistic, so a failed save has to put the old list back. Otherwise the
+    // dialog shows a user as removed while the bot still answers them, which is
+    // the wrong direction to be wrong in for an access list.
+    const previousUsers = allowedUsers;
     onAllowedUsersChange(nextUsers);
     setFormError(null);
 
@@ -67,6 +71,7 @@ export function TelegramAllowedUsersDialog({
       },
       {
         onError: (err) => {
+          onAllowedUsersChange(previousUsers);
           const message = formatError(err);
           setFormError(message);
           onError?.(message);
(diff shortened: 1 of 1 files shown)


ahmadrosid:
Thanks for the focused fix. The main behavior looks correct: clicking remove now opens a confirmation dialog, and cancel, Escape, or overlay dismissal do not call the save mutation.

I found two points to address before closing #890:

- The issue acceptance criteria ask for a cancel-path test. The PR explains that the web app does not currently have a component-test setup, but this still leaves the most important regression path covered only manually. Please either add a test at the available layer or document the manual verification clearly in the issue.
- Telegram and Discord do not use exactly the same save model. Telegram calls `saveAllowedUsers` after confirmation, while Discord updates local state and waits for the explicit Save action. Both avoid the original one-click removal bug, but the issue says the dialogs should behave the same. Please confirm whether this difference is intentional.

One smaller existing behavior is worth noting: `saveAllowedUsers` updates the local list before the mutation resolves. If the save fails, the UI can show the user as removed while the server still contains them. This is not introduced by the confirmation dialog, but the new path makes it worth considering whether the list should be rolled back on mutation failure.

The implementation itself is lean and the confirmation flow is wired correctly. After the cancel, confirm, Escape, overlay, and failed-save cases are verified, this should be in good shape.

fajarhide:
All three addressed.

**Cancel-path test.** Verified against a live instance and written up on #890 with the numbers. Cancel, Escape and overlay dismissal each leave three users and fire zero `PUT /v1/settings/telegram`; confirm leaves two and fires one. Counting the request is what makes "did not save" an observation rather than a claim. Still no component test, because `apps/web` has no runner for one and adding a rendering stack for this dialog is a bigger decision than the fix.

**Telegram and Discord save models.** The difference is intentional. Telegram saves on add and on import too, so making only remove stage-and-wait would leave one action in that dialog behaving unlike its two neighbours. What the issue asked for is that neither dialog removes a user on a single click, and both now clear that. Happy to convert Telegram to Discord's model as its own change if you want them identical.

**Optimistic list on a failed save.** Fixed rather than noted. The error handler restores the previous list, and the confirm now closes before the save resolves, since the error renders in the dialog underneath and nobody can read that through an open confirm. Control run with the restore deleted is on #890: the user disappears from the UI while the server still has them.

fajarhide:
Putting the verification numbers here rather than only linking #890, since this is where they are being asked for.

Isolated instance (`NAKAMA_CONFIG_DIR=/tmp/nakama-tg-demo`, server 4399, Vite 3399), driven with Playwright, three allowed users seeded. The column that matters is the `PUT /v1/settings/telegram` count, because cancel, Escape and overlay all look identical in a screenshot.

| Case | Users after | PUT count |
|---|---|---|
| Confirm opens, Cancel | 111111, 222222, 333333 | 0 |
| Confirm opens, Escape | 111111, 222222, 333333 | 0 |
| Confirm opens, overlay click | 111111, 222222, 333333 | 0 |
| Confirm opens, Remove | 111111, 333333 | 1 |
| Server killed, then Remove | 111111, 333333 | 2 attempted |

Last row is the rollback: the save failed, the list still shows the user, the error renders in the dialog underneath.

Control with the `onAllowedUsersChange(previousUsers)` line deleted, so the last row is not measuring nothing:

```
before:            [ 111111, 222222, 333333 ]
after failed save: [ 111111, 333333 ]
```

The user vanishes from the UI while the server still has them.

Note for anyone repeating this: the bot token is validated against `api.telegram.org/getMe`, hardcoded at `agent-service.ts:1104`, so the card cannot be configured on a throwaway instance. The run above pointed that fetch at a local stub through a temporary uncommitted edit, reverted before committing.

ahmadrosid:
Thanks for the follow-up. The rollback on save failure is a good improvement: if the mutation fails, the previous allowed-user list is restored, so the UI no longer stays optimistically out of sync with the server.

The remaining points are mostly about verification and behavior consistency:

- The cancel-path acceptance criterion from #890 is still not covered by an automated test. Please document the manual verification in the issue or add a test at the highest available layer.
- Telegram saves immediately after confirmation, while Discord changes local state and waits for the explicit Save action. Both prevent the original one-click removal, but the dialogs do not use exactly the same save model. Please confirm that this difference is intentional.

Before closing this, please manually verify that cancel, Escape, and overlay dismissal leave the user unchanged; confirm removes the user; and a failed save restores the previous list. The implementation itself looks correct after the rollback change.

**Verdict:** needs manual verification.

fajarhide:
Demo of the five cases, in order: Cancel, Escape, click outside, Confirm, then a save that fails with the server stopped. The banner carries the running `PUT /v1/settings/telegram` count, because cancel, Escape and overlay dismissal look identical without it.

Telegram allowed-user removal confirm

Three users seeded. Cancel, Escape and overlay each end with three users and a PUT count of 0. Confirm ends with two users and a count of 1. The last case removes `333333` with the server down: the count goes to 2, the save fails, and `333333` is back in the list with the error rendered underneath.

Control, so the last frame is not measuring nothing. Same case with the `onAllowedUsersChange(previousUsers)` line deleted:

```
before:            [ 111111, 222222, 333333 ]
after failed save: [ 111111, 333333 ]
```

The user vanishes from the UI while the server still has them.

Your two points crossed with my previous comment by about forty seconds, so to answer them directly: the manual verification is the run above and is also written up on #890, and the Telegram-versus-Discord difference is intentional, because Telegram saves on add and on import too and staging only the remove would leave one action in that dialog behaving unlike its two neighbours. Happy to convert Telegram to Discord's model as its own change if you want them identical.

ahmadrosid:
I reviewed the latest update and the manual verification added to the PR and linked issue.

The previous findings are addressed:

- The save-failure rollback is present, restoring the previous allowed-user list in `onError`.
- The Telegram/Discord save difference is intentional and documented. Telegram persists after confirmation, while Discord stages the change until Save. Both prevent the original one-click server removal.
- The cancel flow is now documented with PUT request counts and a GIF covering cancel, Escape, overlay dismissal, confirm, and failed-save behavior. There is still no automated component test because the web app has no component test runner, but the manual evidence is sufficient for this scoped change.

I found no new code issues. The confirm dialog closes before the mutation resolves so an error can be shown in the list, and the rollback keeps the UI consistent if the request fails.

**Verdict: ready to merge.**

PR #893, opened 2026-09-07
feat(error-tracking): report a tool that exhausts its retry budget

### LINKED ISSUE #891: feat(error-tracking): send an exhausted tool failure to the tracker, tagged by tool
## Summary

Route a custom tool failure that has exhausted its retries into the Sentry-compatible sink, tagged with the tool name. This is phase 4 of #441, the only phase of that issue with no code behind it.

Phases 1 to 3 are on `main`: `packages/core/src/error-tracking.ts`, `error-tracking-queue.ts`, `error-tracking-sentry.ts` and `error-tracking-config.ts` exist, and all four entrypoints call `installErrorHandlers` (`apps/server/src/index.ts:14`, plus the telegram, whatsapp and discord workers). Nothing calls `reportError` for a tool failure: a grep for callers outside `packages/core/src/error-tracking*` finds only the four handler installs and the Integrations test-event path in `agent-service.ts:1232`.

The prerequisite this waited on, #449, is closed.

Measured on `60ce1bbf`.

## Why it matters

The strategy document makes error rate a P1 concern and asks for every tool failure to be recorded for error-rate monitoring. Today an exhausted tool failure is returned to the model as `{ error: message }` by `executeToolCall` and reaches no sink, so an operator running a tracker sees crashes but not the failure mode that actually degrades a turn.

- `packages/agent/src/tool-loop.ts`, `executeToolCall` is the single place every tool result passes through
- The retry wrapper from #449
- `reportError` in `packages/core/src/error-tracking.ts:184`

## Acceptance criteria

- [ ] A tool failure past its retry budget produces one event tagged with the tool name
- [ ] A failure that a retry then recovers produces no event
- [ ] No DSN means no event and no queue write, matching the rest of the sink
- [ ] The event carries no tool arguments, since those are user content and the scrubber does not see them
- [ ] The reported error still reaches the model unchanged

fajarhide:
## Summary

A custom tool that fails every attempt in its retry budget now reaches the operator's error tracker, tagged by tool. This is phase 4 of #441, the last one with no code behind it.

**Before:** the sink shipped in #444 and nothing outside the four crash handlers called it. A grep for `reportError` callers found the four `installErrorHandlers` sites and the Integrations test-event path, nothing else. An exhausted tool returned `{ error: message }` to the model and vanished.

**After:** one event per exhausted tool, `kind: "tool"` and `source: "tool:<name>"`, so it filters apart from a crash and still lands at Sentry level `error`.

Why safe:
1. `withToolRetries` is the only place that can tell an exhausted tool from one failed attempt, and it wraps both the javascript and python loaders through one seam.
2. Not awaited. `reportError` queues synchronously and never throws, and the model should not wait out the tracker's HTTP timeout on a turn that has already failed. The error is re-thrown unchanged.
3. An aborted signal throws before the budget check, so a cancelled turn reports nothing.
4. `toolName` is a required parameter, so a call site that forgets it fails the typecheck rather than sending `tool:undefined`.

The event carries only what `buildErrorReport` reads from the error, so no tool arguments. "No DSN means no event and no queue write" is `reportError`'s own contract, already covered in `error-tracking.test.ts`, so it is not re-tested here.

## Test plan
- [x] `bun test apps/server/src/services/custom-tool-handlers.test.ts`: 11 pass, 0 fail
- [x] Mutation 1, delete the `reportError` call: 2 tests fail
- [x] Mutation 2, pass `record.id` instead of `definition.name` at the seam: the wiring test fails with `Expected: "tool:doomed_tool"  Received: "tool:tool_flaky"`

Closes #891

### DIFF
--- a/apps/server/src/services/custom-tool-handlers.test.ts
+++ b/apps/server/src/services/custom-tool-handlers.test.ts
@@ -39,9 +39,15 @@ function makeRecord(
 
 let reportedErrors: ErrorReport[] = [];
 let firstReport: Promise<ErrorReport>;
+let queueDir = "";
 const realConsoleError = console.error;
 
 beforeEach(async () => {
+  // reportError queues before it sends, and that queue is a read-modify-write on
+  // one file under NAKAMA_CONFIG_DIR. Without a temp dir here the suite trims
+  // the real queue on whatever machine runs it.
+  queueDir = await mkdtemp(path.join(os.tmpdir(), "nakama-tool-report-"));
+  process.env.NAKAMA_CONFIG_DIR = queueDir;
   process.env.NAKAMA_ERROR_TRACKING_DSN = "https://key@errors.example.test/7";
   await refreshErrorTrackingEnabled();
   reportedErrors = [];
@@ -60,9 +66,15 @@ beforeEach(async () => {
 
 afterEach(async () => {
   delete process.env.NAKAMA_ERROR_TRACKING_DSN;
+  if (originalConfigDir === undefined) {
+    delete process.env.NAKAMA_CONFIG_DIR;
+  } else {
+    process.env.NAKAMA_CONFIG_DIR = originalConfigDir;
+  }
   await refreshErrorTrackingEnabled();
   setErrorSink(null);
   console.error = realConsoleError;
+  await rm(queueDir, { force: true, recursive: true });
 });
 
 describe("withToolRetries", () => {
@@ -133,6 +145,7 @@ describe("withToolRetries", () => {
       withToolRetries(run, "flaky")({}, ctx(controller.signal))
     ).rejects.toBe(cancellation);
     expect(attempts).toBe(1);
+    expect(reportedErrors).toEqual([]);
   });
 
   test("an already-aborted signal never starts the run", async () => {
@@ -149,6 +162,7 @@ describe("withToolRetries", () => {
       withToolRetries(run, "flaky")({}, ctx(controller.signal))
     ).rejects.toBe(cancellation);
     expect(attempts).toBe(0);
+    expect(reportedErrors).toEqual([]);
   });
 
   test("aborting during the backoff cancels the retry", async () => {
@@ -169,6 +183,7 @@ describe("withToolRetries", () => {
     await expect(pending).rejects.toBe(cancellation);
     // Never reached the second attempt.
     expect(attempts).toBe(1);
+    expect(reportedErrors).toEqual([]);
   });
 });
 
--- a/apps/server/src/services/custom-tool-handlers.test.ts
+++ b/apps/server/src/services/custom-tool-handlers.test.ts
@@ -1,8 +1,13 @@
-import { describe, expect, test } from "bun:test";
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
 import { mkdir, mkdtemp, rm, writeFile } from "node:fs/promises";
 import os from "node:os";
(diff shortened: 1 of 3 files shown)


ahmadrosid:
The reporting hook is in the right place: it reports only after the custom tool exhausts its retry budget, keeps the original tool error behavior, and does not report when the tool succeeds after a retry or when the call is aborted.

There is one important test-isolation problem to fix before merging. The new tests set `NAKAMA_ERROR_TRACKING_DSN` and the sink, but they do not set a temporary `NAKAMA_CONFIG_DIR`. The exhausted path writes to the pending error-report queue through `appendPendingErrorReport`. On a developer machine or CI worker with an existing full queue, the test can trim the real queue and remove data that belongs to another run.

Please create an isolated temporary config directory in test setup, set `NAKAMA_CONFIG_DIR` for the test, and clean it up in teardown. The existing error-tracking tests already show the safer pattern.

Please also add explicit assertions that aborted calls produce no report. The recovered-success case checks this, but the abort cases currently only verify the thrown result. That is a small test gap for the no-report contract.

The remaining design looks good: both JavaScript and Python custom tools go through the same retry wrapper, the report does not include tool arguments, and reporting is non-blocking. After isolating the queue, this should be ready for another pass.

fajarhide:
Both fixed, and the queue point was a real leak rather than a theoretical one.

**Queue isolation.** Measured on this machine: with no `NAKAMA_CONFIG_DIR` override, `getPendingErrorReportsPath()` resolves to `~/.nakama/error-tracking/pending.json`, and the earlier runs of this suite created it. `beforeEach` now points `NAKAMA_CONFIG_DIR` at a fresh `mkdtemp` and `afterEach` restores the original and removes it, matching what the seam tests already do per test.

**Abort assertions.** All three abort arms now assert `reportedErrors` is empty, so the no-report contract is covered on every arm rather than only on the recovered-retry one. That is 11 tests, 34 assertions, 0 fail.

`bun run test` is clean across every workspace.

ahmadrosid:
The follow-up fixes the issues from the previous review.

The tests now use an isolated temporary config directory through `mkdtemp` and `NAKAMA_CONFIG_DIR`, so they cannot modify a real pending error-report queue on a developer machine or CI worker. The abort-path tests also explicitly assert that no report was emitted.

The reporting seam still looks correct:

- Reports only after the custom tool exhausts its retry budget.
- A tool that succeeds after retry does not report.
- Aborted calls do not report.
- Tool arguments are not included in the report.
- Reporting remains non-blocking and does not change the original tool error behavior.

The fingerprint does not include the tool name, but the tool name is available as a tag. That is acceptable for this scope and is not a blocker.

**Verdict: ready to merge.**

ahmadrosid:
I reviewed the latest update. The previous test-isolation issue is fixed.

The tests now create an isolated temporary config directory with `mkdtemp` and set `NAKAMA_CONFIG_DIR` during setup, then clean it up afterward. This prevents the test from touching a real pending error-report queue.

The abort cases now explicitly assert that no report was emitted. The behavior remains correct across the main paths:

- A report is emitted only after the custom tool exhausts its retry budget.
- A tool that succeeds after a retry does not emit a report.
- Aborted calls do not emit a report.
- Tool arguments are not included in the report.
- Reporting is non-blocking and does not change the original tool error behavior.

The fingerprint still does not include the tool name, but the tool name is available in the source tag. That is acceptable for this scope and is not a blocker.

**Verdict: ready to merge.**

PR #894, opened 2026-09-07
fix(db): apply bootstrap schema once

### LINKED ISSUE #607: [db] db.exec(sql) loads whole schema.sql on every boot (no version check)
## Bug
Every statement is idempotent (`IF NOT EXISTS`) — but if the schema file ever ships a non-idempotent statement, `migrateDatabase` will run it on every boot and break. No checksum/version comparison.

## Files
- `packages/db/src/migrate.ts:6-10`

## Fix
Add a `schema_version` table and refuse to re-apply past that.

## Severity
Medium — robustness.

isnaenihidayat:
## Summary

Open a SQLite database repeatedly → the bootstrap schema runs once, while compatibility migrations still run on every open.

**Before:** `migrateDatabase` reads and executes the complete `schema.sql` on every database open.
**After:** `schema_version` records bootstrap version 1 transactionally and skips subsequent baseline replays.

Why safe:
1. Foreign-key enforcement remains connection-scoped and runs before the version check
2. The regression test separates a bootstrap-only table from a compatibility-migrated table

Only revisit if an existing database needs a future schema change without a targeted compatibility migration.

Full-suite note: `bun test` reached 2,921 pass with one unrelated local shell PID failure in `apps/server/src/tools/bash.test.ts`, reproducible in isolation.

Closes #607

## Test plan
- [x] `bun test packages/db/src` (101 pass)
- [x] `bun run check && bun run knip && bun run build`
- [ ] Restart the same SQLite-backed server twice and confirm normal startup

### DIFF
--- a/packages/db/src/migrate.test.ts
+++ b/packages/db/src/migrate.test.ts
@@ -493,6 +493,43 @@ describe("schema path resolution", () => {
   });
 });
 
+describe("schema bootstrap version", () => {
+  test("applies schema.sql once while compatibility migrations keep running", () => {
+    const db = new Database(":memory:");
+
+    try {
+      migrateDatabase(db);
+      db.exec("DROP TABLE artifact_shares; DROP TABLE attachments;");
+
+      migrateDatabase(db);
+
+      const artifactSharesTable = db
+        .prepare(
+          "SELECT name FROM sqlite_master WHERE type = 'table' AND name = 'artifact_shares'"
+        )
+        .get();
+      const attachmentsTable = db
+        .prepare(
+          "SELECT name FROM sqlite_master WHERE type = 'table' AND name = 'attachments'"
+        )
+        .get();
+      expect(artifactSharesTable).toBeNull();
+      expect(attachmentsTable).toEqual({ name: "attachments" });
+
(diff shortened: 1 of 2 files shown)


ahmadrosid:
This looks good and addresses #607 with a small, clear change.

The bootstrap schema is now applied once and recorded with a version, while compatibility migrations continue to run on every startup. Keeping `PRAGMA foreign_keys = ON` before the version check is correct because that setting is connection-scoped. The bootstrap and version insert also happen in one transaction, which avoids leaving a partially initialized database marked as complete.

The tests are useful here because they distinguish a bootstrap-only table from a compatibility migration. That verifies the important behavior: the baseline is not replayed on every open, but compatibility work still runs.

One operational caveat is worth documenting. Once version `1` is recorded, a new table added only to `schema.sql` will not appear in existing databases. Future schema changes need a compatibility migration or a version bump with an intentional bootstrap strategy. This is not a blocker for this PR, but it should be clear for future contributors.

The optional remaining check is a manual double-restart against a real SQLite data directory: start with an existing database, restart twice, and confirm the schema and version record remain correct.

**Verdict: ready to merge.**

PR #898, opened 2026-09-07
fix(agent): preserve compacted history in session archives

### LINKED ISSUE #588: [agent] Compaction summary can drop tool calls and results, history loss bounded only by tailTurns
## Bug
`selectCompactionRange` collapses everything older than the last 2 turns into a single LLM summary message. Tool calls/results from the summarized region disappear from `history` and are not recoverable for the rest of the session.

## Files
- `packages/agent/src/history-compaction.ts:225-295`

## Fix
Persist raw tool calls/results to a sidecar JSONL per session, and prepend a short summary + pointer.

## Severity
Medium — data loss in compaction.

mamenesia:
## Summary

Recover original tool calls and results after chat compaction. Fixes #588.

**Before:** compaction replaced the only persisted transcript, permanently discarding older messages.
**After:** original messages are archived to server-private, per-session JSONL before working history is replaced; a session-bound read-only tool retrieves bounded chunks.

Why safe:
1. Failed archival leaves working history intact; no-op compaction writes nothing.
2. Recovery cannot select another session or path. Branches copy archives; clear, purge, and profile deletion clean them up. Failed profile cleanup keeps session IDs available for retry.
3. In-flight writes are coordinated with deletion; invalidated session objects cannot recreate cleared archives. The model still receives compacted history, not the full archive.

Residual: snapshots consume disk until clear, purge, or profile deletion; automatic retention is not introduced here. Assignment requested in https://github.com/ahmadrosid/nakama/issues/588#issuecomment-5570558006. CI, ponytail review, and fresh standards/correctness review pass on the latest revision; this remains draft pending publication approval.

## Test plan
- [x] `bun test apps/server/src/services/session-persistence.test.ts apps/server/src/services/profile-service.test.ts packages/agent/src` — 137 pass, including FK-enabled profile cleanup/retry, concurrent session creation, and stale-session rejection after clear.
- [x] Targeted Ultracite and `bun run knip` pass; latest CI test, knip, and react-doctor checks pass. Full local TypeScript checking remains blocked by existing repository diagnostics.

### DIFF
--- a/apps/server/src/services/session-persistence.test.ts
+++ b/apps/server/src/services/session-persistence.test.ts
@@ -82,7 +82,7 @@ async function seedSession(
   });
 }
 
-describe("wrapPersistedSession", () => {
+describe("session persistence", () => {
   setupTestConfigDir("nakama-history-archive-");
 
   test("compaction replaces working history but preserves raw messages for scoped recovery", async () => {
@@ -254,7 +254,12 @@ describe("wrapPersistedSession", () => {
         "utf8"
       )
     ).toBe(source);
+    await replaceSessionHistory(db, branch!.sessionId, history);
+    service.setAutomationTools([]);
+    const detached = await service.resolveSession(branch!.sessionId, "org_1");
+    service.setAutomationTools([]);
     expect(await service.clearSession(branch!.sessionId, "org_1")).toBe(true);
+    await expect(detached!.compact({ force: true })).rejects.toThrow();
     await expect(
       readFile(sessionHistoryArchivePath("org_1", branch!.sessionId))
     ).rejects.toThrow();
--- a/apps/server/src/services/session-persistence.test.ts
+++ b/apps/server/src/services/session-persistence.test.ts
@@ -331,43 +331,8 @@ describe("session persistence", () => {
       expect(await db.getSession("deleted")).not.toBeNull();
       await rm(path, { recursive: true });
 
-      let queuedWrites: Promise<PromiseSettledResult<unknown>[]> | undefined;
-      const concurrentDb = new Proxy(db, {
-        get(target, property, receiver) {
-          if (property === "listSessions") {
-            return async () => {
-              const snapshot = await db.listSessions();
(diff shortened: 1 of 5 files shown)


ahmadrosid (commented):
Archive-then-replace plus a session-bound read tool is the right shape. Two spots that can get shorter without dropping the safety story.

packages/agent/src/chat.ts
```diff
@@ -267,16 +268,50 @@ export function createAgentChatSession(
...
       systemPrompt,
       tools: llmTools,
     });
 
     if (result.action !== "none") {
+      assertUnchanged();
+      const recovery = await options.archiveHistory?.(original);
+      assertUnchanged();
+      if (recovery) {
+        for (let index = 0; index < compacted.length; index += 1) {
```
ahmadrosid: This loop stamps the same recovery pointer onto every truncated tool message. One append on `compacted[0]` (or the first pruned tool) is enough — `read_session_history` is already session-scoped, so the extra copies don't buy anything.

apps/server/src/services/session-persistence.test.ts
```diff
@@ -1,9 +1,398 @@
...
+      await rm(path);
+      await mkdir(path);
+      const service = new ProfileService(db);
+      await expect(service.deleteProfile("org_1", "deleted")).rejects.toThrow();
+      expect(await db.getProfile("deleted")).not.toBeNull();
+      expect(await db.getSession("deleted")).not.toBeNull();
+      await rm(path, { recursive: true });
+
+      let queuedWrites: Promise<PromiseSettledResult<unknown>[]> | undefined;
+      const concurrentDb = new Proxy(db, {
```
ahmadrosid: Clear-during-write already proves the org lock admits before the first await and releases in `finally`. The `listSessions` Proxy, late session, and overlapping `writing` walk the same lock again. Keep the failed-cleanup retry plus a post-delete reject; you can drop this concurrent block and still have the same confidence.

PR #899, opened 2026-09-07
fix(agent): bound per-turn tool loop output

### LINKED ISSUE #590: [agent] MAX_TOOL_ITERATIONS = 100 permits unbounded token cost
## Bug
100 tool iterations × 1k–32k token results can blow up context. Combined with `pruneToolOutputs` mutating in place, can also produce history that is large yet post-prune misleading.

## Files
- `packages/agent/src/chat.ts:53`

## Fix
Cap by total tokens produced (e.g. 200k) instead of iteration count.

## Severity
Medium — cost / memory.

mamenesia:
## Summary

Stop runaway tool loops after a turn reaches its output budget. Fixes #590.

**Before:** a turn could accumulate more than 200k output tokens and keep calling the provider until the 100-iteration limit.
**After:** assistant output and tool results consume a 200k-token per-turn budget; reaching it ends the turn with an explicit message rather than making another provider call.

Why safe:
1. Completed tool results and already-streamed reply text stay in history; unexecuted tool calls are not retained.
2. Send and streaming share the guard, and serial/parallel batches use the same accounting. Provider-reported output is counted with a conservative text-estimate fallback.
3. The budget resets on the next user turn; the 100-iteration backstop remains for small-output loops.

Residual: a completed tool batch can overshoot the threshold. This is a continuation guard, not a hard memory or billing cap; repeated input tokens are not included.

Assignment requested in https://github.com/ahmadrosid/nakama/issues/590#issuecomment-5572619760. CI, ponytail review, and fresh standards/correctness review pass on the latest revision. This remains draft pending publication approval.

## Test plan
- [x] `bun test packages/agent/src` — 98 pass. Send/stream regressions failed before the fix; coverage includes budget reset, serial/parallel results, and reported/estimated assistant output. Streamed-text preservation regressions also failed before the correction and now pass.
- [x] Targeted Ultracite, `bun run knip`, and `git diff --check` pass; latest CI test, knip, and react-doctor checks pass.

### DIFF
--- a/packages/agent/src/chat-tool-loop.test.ts
+++ b/packages/agent/src/chat-tool-loop.test.ts
@@ -69,6 +69,148 @@ const sampleTool: ToolDefinition = {
 };
 
 describe("agent chat tool loop", () => {
+  test.each(["send", "stream"] as const)(
+    "%s stops accumulating large tool results before another provider call and resets the budget next turn",
+    async (mode) => {
+      let providerCalls = 0;
+      let toolRuns = 0;
+      const tool: ToolDefinition = {
+        ...sampleTool,
+        async run() {
+          toolRuns += 1;
+          return "x".repeat(128_000);
+        },
+      };
+      const provider: ProviderClient = {
+        ...createMockProvider([]),
+        async generateChat() {
+          providerCalls += 1;
+          const toolCalls =
+            providerCalls <= 8
+              ? [
+                  {
+                    arguments: {},
+                    id: `call_${providerCalls}`,
+                    name: tool.name,
+                  },
+                ]
+              : [];
+          return {
+            assistantMessage: { content: "", role: "assistant", toolCalls },
+            content: "",
+            toolCalls,
+          };
+        },
+        streamChat(input) {
+          return this.generateChat(input);
+        },
+      };
+      const session = createAgentChatSession({ provider }, { tools: [tool] });
+      let streamed = "";
+      const reply =
+        mode === "send"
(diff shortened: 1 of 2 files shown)


ahmadrosid (commented):
Loop accounting is the right size. Two test matrices can go — same function, same asserts.

packages/agent/src/chat-tool-loop.test.ts
```diff
@@ -69,6 +69,156 @@ const sampleTool: ToolDefinition = {
 };
 
 describe("agent chat tool loop", () => {
+  test.each(["send", "stream"] as const)(
```
ahmadrosid: send and stream both hit `runConversation`. Keep `sendStream` so you still check the notice goes through `onChunk`, and drop the `send` half of this `test.each`.

packages/agent/src/chat-tool-loop.test.ts
```diff
@@ -69,6 +69,156 @@ const sampleTool: ToolDefinition = {
...
+      if (mode === "stream") {
+        expect(streamed).toBe(reply);
+      }
+      await session.send("Continue");
+      expect(providerCalls).toBe(9);
+      expect(toolRuns).toBe(8);
+    }
+  );
+
+  test.each([false, true])(
```
ahmadrosid: Serial and parallel both finish the whole batch, then count `history.slice(toolHistoryStart)`. One `parallelSafe` value is enough — the other doesn't change the accounting.

PR #900, opened 2026-09-07
fix(server): make the install deadline cancel the install

vianbas:
## Summary

The install deadline now stops the install, instead of only ending the response.

**Before:** the timeout closed the stream and the work carried on. `installAgentBrowser` went on to start the browser download, and the per-command kill reached only the process it spawned.
**After:** `streamInstallEvents` aborts a signal the executor watches, and `runTimedInstallCommand` signals the whole process group.

Why safe:
1. The `signal` option is additive, so the six existing `runTimedInstallCommand` tests pass untouched, and the other caller in `coding-agent-harness-service` needs no change
2. The group signal falls back to the direct child when the group is already gone, which is also the path on Windows where a negative pid is unsupported
3. Mutation-checked: removing `detached`, downgrading the group kill to a child kill, dropping the abort listener, not aborting on the deadline, and not passing the signal each fail a test

Only follow up if an install needs to survive a server stop: the detached child no longer receives the terminal signal that used to reach it through the shared process group.

## Test plan
- [x] `bun test apps/server/src` (1138 pass, 1135 on base)
- [x] `bun test apps/server/src/services/cli-package-install.test.ts` (8 pass)
- [x] `bun test apps/server/src/http/coding-harness-install-stream.test.ts` (7 pass)
- [ ] Start an agent-browser install, close the tab before it finishes, and confirm no install process is left behind

Refs #439

### DIFF
--- a/apps/server/src/services/cli-package-install.ts
+++ b/apps/server/src/services/cli-package-install.ts
@@ -146,6 +146,7 @@ export async function runTimedInstallCommand(
   options: {
     settleTimeoutMs?: number;
     sigtermGraceMs?: number;
+    signal?: AbortSignal;
     timeoutMs?: number;
   } = {}
 ): Promise<{
@@ -158,9 +159,13 @@ export async function runTimedInstallCommand(
   const timeoutMs = options.timeoutMs ?? CLI_INSTALL_TIMEOUT_MS;
   const sigtermGraceMs = options.sigtermGraceMs ?? CLI_SIGTERM_GRACE_MS;
   const settleTimeoutMs = options.settleTimeoutMs ?? CLI_SETTLE_TIMEOUT_MS;
+  const signal = options.signal;
 
   return new Promise((resolve) => {
     const child = spawn(plan.command, plan.args, {
+      // Its own process group, so a deadline can signal the whole install
+      // rather than only the command we spawned. Installers shell out, and
+      // those grandchildren outlive a kill aimed at the direct child.
       env: getToolExecutionEnv(),
       stdio: ["ignore", "pipe", "pipe"],
     });
@@ -177,6 +182,7 @@ export async function runTimedInstallCommand(
       clearTimeout(timeoutId);
       clearTimeout(killTimeoutId);
       clearTimeout(settleTimeoutId);
+      signal?.removeEventListener("abort", onAbort);
     };
 
     /**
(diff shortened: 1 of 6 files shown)


ahmadrosid (commented):
The cancellation wiring is focused. One small simplification in the shared command runner.

apps/server/src/services/cli-package-install.ts
```diff
@@ -195,21 +202,59 @@ export async function runTimedInstallCommand(
...
       }
 
-      child.kill("SIGTERM");
-      killTimeoutId = setTimeout(() => child.kill("SIGKILL"), sigtermGraceMs);
+      killTree("SIGTERM");
+      killTimeoutId = setTimeout(() => killTree("SIGKILL"), sigtermGraceMs);
       settleTimeoutId = setTimeout(settleAsTimedOut, settleTimeoutMs);
+    };
+
+    const onAbort = () => {
```
ahmadrosid: Both callbacks set `timedOut = true` and call `terminate()`. Move that assignment to the start of `terminate` and use it directly for the timeout, abort listener, and already-aborted path (including listener cleanup). That drops `onAbort` and the duplicate timeout wrapper without changing behavior.

vianbas: Done, all four parts: the assignment is the first line of `terminate`, and the timeout, the abort listener, the already-aborted path and the listener cleanup all pass `terminate` itself. `onAbort` and the wrapper are gone, six lines net.

I checked the "without changing behavior" part rather than taking it on trust. `terminate` had exactly two callers, at the timeout and inside `onAbort`, and each set the flag on the line before, so no path reached it with the flag unset.

The line that moved is covered: dropping `timedOut = true` from the top of `terminate` now fails 6 tests, and the other five mutations still fail exactly one each (removing `detached`, downgrading the group kill to a child kill, dropping the abort listener, not aborting the stream on the deadline, not passing the signal to the executor).

Gate on `8eaf3af4`: `bun test apps/server/src` 1138 pass 0 fail, unchanged from before the refactor; `tsc --noEmit` in `apps/server` no new errors; `ultracite` clean.

One note on how I verified, since the earlier red build here was mine: the two touched test files run 15 pass 0 fail on macOS and, in a container on the Bun version CI uses, 15 pass 0 fail on Linux too. That earlier failure was a missing `detached: true` in the commit rather than anything platform-specific, and it is in this branch now.

PR #907, opened 2026-09-08
feat(providers): direct Perplexity Sonar integration

### LINKED ISSUE #390: feat(providers): direct Perplexity Sonar integration
## Summary

Add **Perplexity** as a direct first-party provider (search-grounded Sonar models).

|---|---|---|---|---|
| Global | `https://api.perplexity.ai` | `perplexity` | `PERPLEXITY_API_KEY` | Perplexity Sonar |

OpenAI-compatible chat completions → DeepSeek template. No coverage today.

**Value note:** Sonar models return citations for search-grounded answers — worth checking how citation blocks surface through Nakama's message mapping rather than dropping them.

| Capability | Status |
|---|---|
| Chat vision (image input) | ⚠️ some sonar-pro variants accept images — set per catalog entry |
| Audio transcription (STT) | ❌ |
| Image generation | ❌ |

Standard six: `ProviderName` union · `apiKeyEnvVarForProvider` · `create.ts` case · `models.ts` curated entries · `pricing.ts` · tests.

## Acceptance criteria

- [ ] `perplexity` card with own key field and default base URL
- [ ] Search citations render sensibly in the transcript (or are documented as truncated)

⚠️ Re-verify model lineup (sonar / sonar-pro / reasoning variants) against docs during implementation.

isnaenihidayat:
## Summary

Add Perplexity Sonar in Settings → LLM providers → search-grounded chat keeps its citation links.

**Before:** Sonar required a custom compatible endpoint, and top-level citation URLs were dropped.
**After:** first-party `perplexity` uses `PERPLEXITY_API_KEY`, four curated Sonar models, and numbered source links for send + stream.

Why safe:
1. Reuses the existing OpenAI-compatible transport at Perplexity's documented base URL
2. A recorded live cassette and focused tests pin routing, request shape, and citation delivery

Only revisit the transport if Perplexity changes its OpenAI-compatible response or citation fields.

## Test plan

- [x] `bun test apps/server/src/providers/perplexity.llm.test.ts packages/core/src/provider-resolution.test.ts packages/core/src/provider-setup-prompt.test.ts apps/server/src/providers/create.test.ts apps/server/src/providers/models.test.ts apps/server/src/providers/openai/stream.test.ts apps/web/src/lib/models.test.ts` (106 pass)
- [x] `bun run check`, `bun run knip`, `bun run build`
- [x] Add Perplexity in Settings with a live key, record the cassette, send one streamed Sonar reply, and confirm numbered source links render

## Browser proof

Settings → Add provider → Perplexity Sonar → select Sonar in chat → streamed answer with numbered source URLs.

https://github.com/user-attachments/assets/2f294f93-d464-45e4-ae5d-e4493c79d4a7

Closes #390


apps/server/src/providers/openai/index.ts
```diff
@@ -357,6 +361,9 @@ async function buildChatCompletionRequestBody(options: {
     ...(provider === "deepseek"
       ? buildDeepSeekThinkingBody(options.thinking)
       : {}),
+    ...(provider === "perplexity"
+      ? buildPerplexityThinkingBody(options.thinking)
```
ahmadrosid: Could inline this helper: check `provider === "perplexity" && options.thinking?.enabled` here and spread `{ reasoning_effort: normalizeThinkingEffort(options.thinking.effort) }`. That keeps the same behavior and removes a function with just one caller.

ahmadrosid (commented):
Reusing the existing transport keeps this small. Just one minor simplification below.

apps/server/src/providers/openai/index.ts
isnaenihidayat: Updated in 6c3c204f. The reasoning option is now inlined at the call site and the single-use helper is removed. The focused provider tests, format check, and unused-code check pass.

ahmadrosid:
Could you share a short screen recording showing this working in action? I don’t have a `PERPLEXITY_API_KEY`, so I can’t test it locally. A simple walkthrough of the flow and the result would help me review it. Thanks!

isnaenihidayat:
Live proof is ready.

The recording shows Settings → Add provider → Perplexity Sonar → select Sonar in chat → receive a streamed answer with numbered source URLs. The live response is also captured as an offline replay cassette in `439e558b`.

https://github.com/user-attachments/assets/2f294f93-d464-45e4-ae5d-e4493c79d4a7

ahmadrosid:
Very nice 👍

PR #909, opened 2026-09-08
fix(server): preserve event order when replacing replay snapshots

Yudis-bit:
## Summary

Reconnect to an active chat and receive buffered tool events in their original order.

**Before:** Replacing a todo snapshot could replay `tool_end` before `tool_start`, so the web client dropped the result and showed a completed tool as running.
**After:** `removeEventAt` removes the obsolete snapshot in place and adjusts subsequent snapshot indexes, preserving the order of retained events.

Why safe:
1. Live delivery and the policy of retaining the latest snapshot are unchanged.
2. Regression tests cover replay versus live ordering and repeated updates to interleaved tool-input snapshots.

## Test plan
- [x] `bun test apps/server/src/services/session-turn-registry.test.ts apps/server/src/http/routes/sessions.stream.test.ts` (17 pass; the new ordering regression fails on unmodified `main`).
- [x] `bun run check` (1,295 files, LF validation checkout), `bun run knip`, and `bun run --filter @nakama/server build`.
- [ ] Reload an active chat after two todo updates; the completed tool should retain its result and completed status.

The full server suite on Windows had 1,102 passes and 46 failures: 45 reproduced on unmodified `main`; the remaining theme-hash failure disappears with LF checkout. The other workspace suites also encountered Windows file-permission/locking failures in core and db. Linux CI is still pending.

### DIFF
--- a/apps/server/src/services/session-turn-registry.test.ts
+++ b/apps/server/src/services/session-turn-registry.test.ts
@@ -1,4 +1,5 @@
 import { describe, expect, test } from "bun:test";
+import type { AgentTodo, StreamEvent } from "@nakama/core";
 import { SessionTurnRegistry } from "./session-turn-registry";
 
 describe("SessionTurnRegistry", () => {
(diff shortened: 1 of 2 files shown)


ahmadrosid (commented):
LGTM

PR #914, opened 2026-09-08
feat(providers): direct Moonshot Kimi integration (intl + CN) (#384)

### LINKED ISSUE #384: feat(providers): direct Moonshot Kimi integration (intl + CN)

masrayfa:
Add Moonshot as a direct first-party provider. Like MiniMax (#380), Moonshot runs split platforms with separate keys and catalogs:

- moonshot     -> https://api.moonshot.ai/v1  (MOONSHOT_API_KEY)
- moonshot_cn  -> https://api.moonshot.cn/v1  (MOONSHOT_CN_API_KEY)

Both endpoints are OpenAI-compatible, so each follows the DeepSeek template (createOpenAIProvider with a region base URL), and both join DISCOVERY_MODEL_PROVIDERS rather than shipping a hardcoded catalog: Browse hits {baseUrl}/models and the fetched entries are stored as instance customModels. Resolution, defaults, vision flags and pricing all derive from that list, so new K-series releases appear without code changes — and the intl/CN catalogs stay independent even where the two platforms expose overlapping model ids.

That matters for capabilities here: intl K2.x is text-only while the CN platform exposes *vision-preview variants, so supportsVision is opt-in from discovered metadata instead of being inferred from the id. No Transcription entry and document-content sets empty, like deepseek.

Two fixes the registry surfaced, both of which Moonshot would otherwise have inherited:

- config.ini round-trip: the writer persists customModels for every provider type, but the reader only parsed models_json for a hardcoded whitelist that excluded discovery providers. A discovered catalog was written and then silently dropped on the next load, leaving the instance with no models after a restart. The reader now routes the whole registry through parseCustomModelsJson.
- pricing: discovery providers have no catalog to price against, so getModelPricing fell through to DEFAULT_PRICING ($1/$3) and reported a confident-but-wrong cost estimate. They now resolve saved per-model pricing like openai_compatible/cerebras/fireworks, and report unknown when the user has not entered rates.

Endpoints and the /v1/models capability were re-verified against the Moonshot platform docs, as the issue asks.

Tests: env-key mapping per region, parseProviderName acceptance, discovery-registry membership, discovered-model resolution with fallback to the instance default, region-independent catalogs, vision opt-in, the models_json round-trip regression (fails without the reader fix), and pricing with and without saved rates.

Closes #384

### DIFF
--- a/apps/server/src/providers/create.ts
+++ b/apps/server/src/providers/create.ts
@@ -96,6 +96,8 @@ function createProvider(options: CreateProviderOptions): ProviderClient {
       });
     case "minimax":
     case "minimax_cn":
+    case "moonshot":
+    case "moonshot_cn":
     case "zhipu":
     case "zhipu_cn":
       return createOpenAIProvider({
--- a/apps/server/src/providers/models.test.ts
+++ b/apps/server/src/providers/models.test.ts
@@ -206,6 +206,37 @@ describe("resolveModel", () => {
       "glm-5.2"
     );
   });
+
+  test("resolves Moonshot models from discovered custom models", () => {
+    const customModels = [
+      { default: true, id: "kimi-k2.5-turbo-preview" },
+      { id: "kimi-k2.5" },
+      { id: "moonshot-v1-128k" },
+    ];
+
+    expect(resolveModel("moonshot", "kimi-k2.5", customModels)).toBe(
+      "kimi-k2.5"
+    );
+    expect(getDefaultModel("moonshot", customModels)).toBe(
+      "kimi-k2.5-turbo-preview"
+    );
+    expect(getDefaultModel("moonshot_cn", customModels)).toBe(
+      "kimi-k2.5-turbo-preview"
+    );
+  });
+
+  test("keeps Moonshot region catalogs independent", () => {
+    // Both platforms expose overlapping ids, so an id discovered on one
+    // instance still resolves against that instance's own list.
+    const cnModels = [{ default: true, id: "moonshot-v1-8k-vision-preview" }];
+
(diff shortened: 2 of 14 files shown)


ahmadrosid (commented):
LGTM

masrayfa:
Thanks, mas! I'm so glad I was able to contribute 😄

PR #919, opened 2026-09-08
chore: replace PATH and timezone packages with platform APIs

ahmadrosid:
## Summary

Server and web lean on platform PATH, tzdata, and the shared model browse list instead of extra packages and alias wrappers.

**Before:** `fix-path` + `@countrystatecity/timezones` plus `maskSecret` / `emailConfigToMailboxConfig` aliases and a one-off OpenCode Go list.
**After:** login-shell PATH on macOS, `Intl` ∪ IANA `zone.tab` (bundled fallback if tzdata is missing), `maskTrailingSecret` / `toMailboxConfig`, `CatalogModelsBrowseList`.

Why safe:
1. `NAKAMA_DISABLE_FIX_PATH` still skips PATH mutation; tool env still prepends Bun / `~/.local` / `/usr/local`
2. Country groups come from tzdata, then the bundled zone.tab; Indiana/Kentucky IDs stay selectable
3. Email, CLI ini keys, and browse search/id/price rows stay the same

Only follow up if a GUI launch still has an empty PATH after the login-shell copy.

## Test plan
- [x] `bun test apps/server/src/services/timezone-catalog-service.test.ts` (3 pass)
- [x] `bun test packages/core/src/email-config.test.ts apps/cli/src/cli-config.test.ts packages/core/src/tools/email.test.ts packages/core/src/tools/extract-document-text.test.ts apps/server/src/services/profile-service.test.ts apps/server/src/services/agent-browser-service.test.ts` (66 pass)
- [ ] Settings → Timezone: search `Indianapolis` → United States / `Indiana/Indianapolis`

## Browser proof

Settings → Timezone: open list, scroll country groups, search New York, pick it, save.

https://github.com/user-attachments/assets/d5a26e66-e1bc-48d6-acb1-659e911bc028


ahmadrosid (commented):
Found two issues. CI is green, but I would not merge this yet.

1. **P1: The catalog drops valid IANA aliases.**

`Intl.supportedValuesOf("timeZone")` returns canonical zones only. It excludes valid IDs such as `America/Indiana/Indianapolis`, `America/Indiana/Knox`, and `America/Kentucky/Louisville`, which were included by `@countrystatecity/timezones`. Existing saved values still work, but users can no longer select these zones from Settings.

Affected code: `apps/server/src/services/timezone-catalog-service.ts:100`.

Please retain the complete IANA zone list, including backward-link aliases, or keep equivalent bundled data.

2. **P1: Country grouping silently breaks without Unix tzdata files.**

Country metadata is loaded only from `/usr/share/zoneinfo/zone1970.tab` and `/usr/share/zoneinfo/zone.tab`. If neither file exists, every zone gets `countryCode: "ZZ"` and the UI groups zones by prefixes such as `America` instead of countries. This affects Windows and runtimes or slim images without those files.

Affected code: `apps/server/src/services/timezone-catalog-service.ts:11-14` and `:45-68`.

Please ship a fallback country map, require/install tzdata explicitly, or fail loudly instead of returning an incorrect catalog.

The added test only covers `America/New_York`, so it does not catch either regression.

ahmadrosid:
> The catalog drops valid IANA aliases. `Intl.supportedValuesOf("timeZone")` returns canonical zones only.

> Country grouping silently breaks without Unix tzdata files.

Fixed in 7355ad78:

1. Catalog is `Intl.supportedValuesOf("timeZone")` ∪ IANA `zone.tab` ids, so `America/Indiana/Indianapolis` and `America/Kentucky/Louisville` show up under United States.
2. If `/usr/share/zoneinfo/zone*.tab` is missing, country codes come from the bundled `zone.tab` fallback instead of dumping everything into `ZZ`.

Tests cover the omitted IANA ids and a missing-tzdata `readFileSync` failure.

PR #922, opened 2026-09-08
feat(providers): add Doubao Volcengine Ark (CN)

### LINKED ISSUE #386: feat(providers): direct ByteDance Doubao integration (Volcengine)
## Summary

Add **ByteDance Doubao (Volcengine Ark)** as a direct first-party provider.

|---|---|---|---|---|
| China | `https://ark.cn-beijing.volces.com/api/v3` | `doubao` | `DOUBAO_API_KEY` | Doubao (Volcengine) |
| International | ⚠️ BytePlus variant exists — verify endpoint at implementation | `doubao_intl`? | `DOUBAO_INTL_API_KEY` | Doubao (Intl) |

OpenAI-compatible → DeepSeek template. No coverage today.

**Naming note:** Volcengine's primary platform is CN; the international offering runs under BytePlus branding with a separate console. Confirm whether to ship CN-only first or both.

| Capability | Status |
|---|---|
| Chat vision (image input) | ✅ Doubao vision variants |
| Audio transcription (STT) | ✅ Seed-ASR / speech models |
| Image generation | ✅ Seedream via Ark image API |

Standard six: `ProviderName` union · `apiKeyEnvVarForProvider` · `create.ts` case(s) · `models.ts` curated entries per region · `pricing.ts` · tests.

## Acceptance criteria

- [ ] Region scope decided and recorded here (CN-only vs CN+intl)
- [ ] Card(s) with own key fields and default base URLs
- [ ] STT + image generation opted in where endpoints confirmed

⚠️ Re-verify endpoints and current Doubao 2.x lineup against docs during implementation.

akmalsyrf:
## Summary

Settings → LLM providers gains **Doubao (Volcengine)** — Ark Beijing chat via `DOUBAO_API_KEY`, with `thinking.type` always set (default-on Seed models opt out unless UI enables) and shortlists that survive restart.

**Before:** Doubao only via custom OpenAI-compatible endpoint; thinking unset left Ark default-on; shortlists dropped on reload.
**After:** first-party `doubao` at `https://ark.cn-beijing.volces.com/api/v3` with Seed 2.1 Pro/Turbo + Seed 1.8; `thinking.type` only (no DeepSeek `reasoning_effort`); Turbo marked thinking-capable.

Why safe:
1. Reuses `createOpenAIProvider` — same path as DeepSeek/Mistral/Together
2. CN-only scope; BytePlus intl left out until the endpoint is confirmed
3. Routing + catalog + Ark thinking on/off covered by unit tests

Only re-add BytePlus / STT / image-gen, or refresh USD list prices, if operators hit wrong region or cost drift.

Closes #386.

## Test plan
- [x] `bun test packages/core/src/provider-resolution.test.ts packages/core/src/user-config.test.ts apps/server/src/providers/create.test.ts apps/server/src/providers/models.test.ts apps/web/src/lib/models.test.ts` (120 pass)
- [ ] Add Doubao with a real Ark key → chat with `doubao-seed-2-1-pro-260628`; confirm thinking off sends `thinking.type=disabled`


apps/server/src/providers/models.ts
```diff
@@ -785,21 +818,23 @@ export function getDefaultModel(
...
-                  : provider === "cerebras"
-                    ? "gpt-oss-120b"
-                    : provider === "fireworks"
-                      ? "accounts/fireworks/models/kimi-k2p6"
-                      : provider === "opencode_go"
-                        ? "opencode-go/kimi-k2.7-code"
-                        : provider === "cloudflare"
-                          ? "@cf/meta/llama-3.3-70b-instruct-fp8-fast"
-                          : "gpt-5.4";
+            : provider === "doubao"
```
fajarhide: This branch never runs. `getModelsForProvider("doubao")` returns your three entries, the pro one carries `default: true`, and the last line of the function reads `models.find((model) => model.default)` before it ever reaches `fallback`. I deleted these two lines on your branch and ran your own tests:

```
models.test.ts + create.test.ts -> 44 pass, 0 fail
getDefaultModel("doubao") still returns doubao-seed-2-1-pro-260628
```

Dropping it also undoes the 15-line reindent of the chain below, so the diff gets smaller in two places.

apps/server/src/providers/openai/index.ts
```diff
@@ -364,7 +368,9 @@ async function buildChatCompletionRequestBody(options: {
     ),
     ...(provider === "deepseek"
       ? buildDeepSeekThinkingBody(options.thinking)
-      : {}),
+      : provider === "doubao"
+        ? buildDoubaoThinkingBody(options.thinking)
```
fajarhide: Nesting inside the deepseek ternary means the next provider nests one deeper. Perplexity, twelve lines down, spreads its own conditional instead. Same line count, flatter, and #923 lands right here too.

fajarhide (commented):
Reusing `createOpenAIProvider` with a base-URL constant, and letting the catalog carry the three models, is the shape this repo wants; the Ark thinking body is the only real logic added and its test bites when I disconnect it. One branch in `getDefaultModel` never runs.

Both notes below are small, nothing blocks. What I did not check: the Ark model ids and the USD prices, I have no key and did not reach Volcengine, so those rest on your reading of their AI Hub. Worth knowing that #923 edits the same 19 files, so whichever lands first leaves the other with conflicts in `models.ts`, `create.ts` and `provider-resolution.ts`.

akmalsyrf:
Thanks — dropped the dead doubao default fallback and flattened the thinking body like Perplexity. Still green on models + create.

fajarhide (approved):
Both taken, and the flattened spread reads better than what I suggested: doubao sits beside perplexity now instead of hanging off deepseek, so the next provider adds a line rather than a level.

Ran it on 9546949: 44 pass across models and create, `getDefaultModel("doubao")` still returns `doubao-seed-2-1-pro-260628` with the fallback branch gone, which was the point. Pulled the thinking spread back out and your routing test goes red, so it still bites. Full `apps/server` 1217 pass, `apps/web` 390 pass.

Two things I still cannot vouch for, same as before: the Ark model ids and the USD prices, no key here. And CI has not reported on this head yet, it says no checks on the branch, so that is worth waiting for rather than reading my local run as the same thing.

Code reads ready to me. #923 still edits these same files, so whichever of the two goes first leaves the other a rebase.

PR #923, opened 2026-09-08
feat(providers): add Qwen DashScope intl and CN

### LINKED ISSUE #385: feat(providers): direct Qwen DashScope integration (intl + CN)

akmalsyrf:
## Summary

Settings → LLM providers gains **Qwen (DashScope)** and **Qwen (DashScope CN)** as separate cards — intl/CN keys, vision on `qwen3.7-plus`, and `enable_thinking` that actually opts out of DashScope’s default-on hybrid thinking.

**Before:** Qwen only via OpenCode Go or a hand-rolled compatible endpoint; thinking off was a no-op on hybrid models; shortlists dropped on reload.
**After:** `qwen` / `qwen_cn` with curated text + VL; always send `enable_thinking`; catalog marks `qwen3.7-plus` vision; shortlists reload from `models_json`.

Why safe:
1. Same OpenAI-compatible factory as DeepSeek/Mistral — only base URL + provider name change
2. Region variants stay independent (two types, two env keys, duplicated catalog rows)
3. Mock routing tests pin `/compatible-mode/v1/chat/completions` and both thinking on/off via `test.each`

Only re-record a live cassette or tweak list prices if DashScope renames the Qwen3.7 lineup.

Closes #385.

## Test plan
- [x] `bun test packages/core/src/provider-resolution.test.ts packages/core/src/user-config.test.ts apps/server/src/providers/create.test.ts apps/server/src/providers/models.test.ts apps/web/src/lib/models.test.ts` (122 pass)
- [x] Merged `upstream/main` (Perplexity) — conflicts resolved; both providers coexist
- [ ] Add Qwen intl (or CN) with a real DashScope key → chat text + one VL turn on `qwen3.7-plus`; confirm thinking off skips reasoning tokens


apps/server/src/providers/create.test.ts
```diff
@@ -142,4 +142,127 @@ describe("createProviderForInstance routing", () => {
       mock.stop(true);
     }
   });
+
+  test("routes qwen instances to the DashScope compatible-mode path with auth", async () => {
```
ahmadrosid: These two regional tests repeat the mock server, response, instance setup, and cleanup. A test.each over qwen and qwen_cn can share that body while keeping the path, auth, model, provider name, and thinking assertions for both regions. That removes the second server fixture without losing either routing check.

apps/server/src/services/provider-instance-helpers.ts
```diff
@@ -156,15 +157,22 @@ export function modelExistsOnInstance(
...
+    instance.type === "qwen_cn"
   ) {
     if (instance.customModels?.length) {
       return findCustomModel(instance.customModels, trimmed) !== undefined;
     }
-    return Boolean(getModelById(trimmed)?.provider === instance.type);
+    // Region variants (qwen / qwen_cn) share model ids — scope by provider.
+    return getModelsForProvider(instance.type).some(
+      (model) => model.id === trimmed
+    );
```
ahmadrosid: This catalog lookup is identical to the final return below. Drop this inner return and let providers without custom models fall through to the shared lookup. The custom-model early return still preserves shortlist behavior, and the final lookup already scopes by provider.

ahmadrosid (commented):
Two small simplifications: share the regional routing test and keep one catalog fallback.

apps/server/src/providers/create.test.ts
akmalsyrf: Done — folded qwen / qwen_cn into one test.each over the shared DashScope mock.

apps/server/src/services/provider-instance-helpers.ts
akmalsyrf: Done — custom-model shortlist stays early-return; catalog falls through to the shared getModelsForProvider lookup.

apps/server/src/providers/openai/index.ts
```diff
@@ -457,6 +467,17 @@ function buildDeepSeekThinkingBody(
   };
 }
```
ahmadrosid, line 470: This helper only turns a boolean into a one-field object and has one caller. Inline `{ enable_thinking: Boolean(options.thinking?.enabled) }` in the Qwen branch and drop the helper. It still sends an explicit false when thinking is absent or disabled.

ahmadrosid (commented):
One small simplification in the thinking payload; the rest follows the existing provider setup.

apps/server/src/providers/openai/index.ts
akmalsyrf: Done — inlined `{ enable_thinking: Boolean(options.thinking?.enabled) }` and dropped the helper. Routing test still asserts true when on and false when unset.

PR #924, opened 2026-09-08
fix(web): poll skill-review overlay via total AgentChannel map

### LINKED ISSUE #910: [web] The skill review overlay polls only channel 'web', so suggestions from four channels are never read
## Summary

The post-turn skill review runs for `cli` sessions and writes `skill_suggestions` rows, but nothing polls for them. The overlay is the only reader of those rows, and it polls only when the session channel is exactly `web`.

- `apps/server/src/services/skill-post-turn-review-service.ts:25` has `cli: true`.
- `apps/web/src/hooks/use-post-turn-skill-review-overlay.tsx:45` requires `sessionChannel === "web"`.

This is #213 one layer up. There the server gate excluded channels the product wanted; here the server includes `cli` and the client gate excludes it.

One limit on the scope, stated up front. This is the suggestion branch only. `agent-service.ts:573` stages a proposal instead when the org requires write approval, and the proposal readers at `SkillProposalsPanel.tsx:265` and `use-notifications.ts:43` pass no `sessionId`, so proposals from any channel are reachable. Only the `createSuggestion` branch at `agent-service.ts:588` produces rows with nowhere to appear.

## Measured on

`ab71ef9b`

## Reproduction

Both gates read from source, then the reader chain followed hop by hop.

```
grep -rn 'usePostTurnSkillReviewOverlay' apps packages --exclude-dir=node_modules
grep -rn 'useSkillSuggestions' apps packages --exclude-dir=node_modules
grep -rn 'listSkillSuggestions' apps packages --exclude-dir=node_modules
```

One caller at each hop, no extension filter applied so the search can fail: `apps/web/src/pages/chat/chat-page-content.tsx:59`, then `use-post-turn-skill-review-overlay.tsx:147`, then `use-skill-suggestions.ts:19`. Every other hit on the last command is the plumbing behind that one call: the SDK method in `packages/client`, the route in `apps/server`, the service, and the SQLite adapter.

The overlay passes `sessionId` into both queries at `use-post-turn-skill-review-overlay.tsx:143`, so the query is per session. That is what makes the rows unreachable rather than merely displaced: were the query org wide, a `cli` session's suggestions would surface in any web session, the way proposals already do.

`apps/cli` does not display them either. Every `suggestion` in `apps/cli/src` is a `PromptSuggestion` or `ComposerSuggestion`, the slash-command autocomplete built in `commands.ts:173` and drawn in `terminal-renderer.ts:84`. Nothing there reads the API.

The server gate measured at runtime rather than by reading the table, by calling `runPostTurnSkillReview` once per channel against a stub database:

```
web        REVIEW RUNS  (profile_missing)
cli        REVIEW RUNS  (profile_missing)
telegram   REVIEW RUNS  (profile_missing)
whatsapp   REVIEW RUNS  (profile_missing)
discord    REVIEW RUNS  (profile_missing)
automation gated        (channel_not_interactive)
task       gated        (channel_not_interactive)
subagent   gated        (channel_not_interactive)
```

`profile_missing` is the next check after the channel gate, so it is the marker for a channel that passed it.

## Expected vs actual

Cross-tabulating that against the two web tables in `chat-history.ts`:

```
channel    review  listed  readOnly  polled  rows nobody reads
automation false   false   false     false   no
cli        true    false   false     false   YES
discord    true    true    true      false   YES
subagent   false   false   false     false   no
task       false   false   false     false   no
telegram   true    true    true      false   YES
web        true    true    false     true    no
whatsapp   true    true    true      false   YES
```

Expected: a channel the server generates suggestions for is a channel some surface can show them on.

Actual: four of the eight write suggestion rows that no surface reads. `cli` is the one where the channel comparison is the only thing stopping it. The other three are read only in the web UI, so `!readOnlySession` in the same predicate excludes them a second time, and that gate is a deliberate decision with a comment on it.

## Root cause or proposed fix

The decision lives in two places that can disagree, and #798 already fixed that class everywhere else. It converted seven channel decisions into total maps over `AgentChannel`. This comparison was not one of them. Adding a ninth member to `AGENT_CHANNELS` and changing nothing else:

```
cd apps/web && bun x tsc@5.9.3 --noEmit
before 0 errors   after 4 errors   new 4
```

All four land in `src/lib/chat-history.ts`, the file #798 converted. None land in `use-post-turn-skill-review-overlay.tsx`. Two runs per side, identical, and reverting the ninth member returns it to 0. So a ninth channel would repeat this silently at exactly this line, which is what #474 was about.

Three ways to close it. I lean to the first, but it is a product decision more than a code one, so I would rather ask than guess.

1. Make the overlay's channel decision a total map over `AgentChannel`, matching #798. Smallest change, and it makes the next channel ask instead of going quiet. It does not by itself surface `cli`, because `HISTORY_SESSION_CHANNEL` in `chat-history.ts` has `cli: false`, so a `cli` session has no web page for a banner to appear on.
2. List `cli` sessions in the web history as well, and let the overlay poll them. This is the only option that actually shows the work the review already does.
3. Treat the current behaviour as intended and stop running the review for `cli` on the server, so it stops writing rows nothing reads.

Whichever way it goes, the decision covers four channels rather than one, so it is worth settling `telegram`, `whatsapp` and `discord` in the same change.

I found this while measuring the `sessionChannel` consumers for #522.

akmalsyrf:
## Summary

Post-turn skill-review overlay poll gate is a total `AgentChannel` map, and it polls `web` + `cli` so CLI suggestion rows are readable when that session is opened in the dashboard.

**Before:** `sessionChannel === "web"` — a ninth channel would not typecheck-fail here; `cli` suggestions were written and never polled.
**After:** `POST_TURN_OVERLAY_POLL_CHANNEL[sessionChannel]` with `web` and `cli` true; telegram/whatsapp/discord stay false (read-only in web).

Why safe:
1. `cli` sessions already open via the chat route / `resumeSession` — history listing is unrelated (#910 correction)
2. Matches the #798 total-map pattern; new channels fail the typecheck here
3. Read-only channel sessions still skip poll via `!readOnlySession`

Only change telegram/whatsapp/discord (or stop server review for them) if product wants suggestion apply from those histories.

Closes #910

## Test plan
- [x] Diff review: map covers every `AgentChannel` via `satisfies Record<AgentChannel, boolean>`; `cli` + `web` true
- [x] CI green on prior head; re-run after `cli: true`
- [ ] Open a web chat with post-turn review on — banner still polls after a complex turn
- [ ] Open a `cli` session URL in the dashboard after a CLI turn with review on — overlay polls / banner can show

### DIFF
--- a/apps/web/src/hooks/use-post-turn-skill-review-overlay.tsx
+++ b/apps/web/src/hooks/use-post-turn-skill-review-overlay.tsx
@@ -17,6 +17,26 @@ import { canAccessSystemPage } from "@/lib/navigation";
 const POST_TURN_POLL_WINDOW_MS = 45_000;
 const POST_TURN_POLL_INTERVAL_MS = 3000;
 
+/**
+ * Which channels the post-turn skill-review overlay polls. Total over
+ * `AgentChannel` so a new channel fails the typecheck here rather than
+ * silently skipping (#474 / #798). Keep aligned with
+ * `HISTORY_SESSION_CHANNEL` ∩ !`READ_ONLY_SESSION_CHANNEL` in chat-history:
+ * only a writable listed session has a web surface that can apply
+ * suggestions. `cli` is false until history lists it (issue option 2);
+ * telegram/whatsapp/discord stay false because they are read-only in web.
+ */
+const POST_TURN_OVERLAY_POLL_CHANNEL = {
+  automation: false,
+  cli: false,
+  discord: false,
+  subagent: false,
+  task: false,
+  telegram: false,
+  web: true,
+  whatsapp: false,
+} as const satisfies Record<AgentChannel, boolean>;
+
 interface UsePostTurnSkillReviewOverlayArgs {
   lastSuccessfulTurnAt: number | null;
   profile: ProfileSummary | undefined;
@@ -42,7 +62,7 @@ function canPollPostTurnReview({
     reviewEnabled &&
     Boolean(activeOrgId) &&
     Boolean(sessionId) &&
-    sessionChannel === "web" &&
+    POST_TURN_OVERLAY_POLL_CHANNEL[sessionChannel] &&
     !readOnlySession
   );
 }
--- a/apps/web/src/hooks/use-post-turn-skill-review-overlay.tsx
+++ b/apps/web/src/hooks/use-post-turn-skill-review-overlay.tsx
@@ -20,15 +20,16 @@ const POST_TURN_POLL_INTERVAL_MS = 3000;
 /**
  * Which channels the post-turn skill-review overlay polls. Total over
  * `AgentChannel` so a new channel fails the typecheck here rather than
- * silently skipping (#474 / #798). Keep aligned with
- * `HISTORY_SESSION_CHANNEL` ∩ !`READ_ONLY_SESSION_CHANNEL` in chat-history:
- * only a writable listed session has a web surface that can apply
- * suggestions. `cli` is false until history lists it (issue option 2);
- * telegram/whatsapp/discord stay false because they are read-only in web.
+ * silently skipping (#474 / #798).
+ *
+ * `web` and `cli`: poll — a cli session opens in the web chat route with
+ * `sessionChannel === "cli"` (history listing is separate; see #910).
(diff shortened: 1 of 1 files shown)


ahmadrosid (commented):
LGTM

PR #925, opened 2026-09-08
fix(cli): allowlist @path image reads to cwd and config dir

### LINKED ISSUE #545: [security] parseImageLine reads arbitrary user-supplied file paths
## Bug
The `@<path>` command reads any file the user names (or that an attacker pastes into stdin) up to `MAX_IMAGE_BYTES`. There is no allowlist of directories. `existsSync` + `Bun.file` happily read `/etc/passwd`-sized content (capped by 5 MB) and base64-encodes it for transmission. A clipboard-stuffing prompt injection can exfiltrate any file the CLI user can read.

## Files
- `apps/cli/src/image-input.ts:15-53`

## Fix
Require `@./` or an explicit base directory; or warn and confirm when the path is outside `~/.nakama` or `/Users/vincent`.

## Severity
High — local file exfil via prompt injection.

akmalsyrf:
## Summary

CLI `@path` image attachment only reads files under `process.cwd()` or the Nakama config dir (`~/.nakama` / `NAKAMA_CONFIG_DIR`).

**Before:** any readable path up to 5 MB was base64-encoded into the prompt.
**After:** `resolveAllowedImagePath` realpath-checks against those two roots (true directory containment, not string-prefix) and rejects everything else (e.g. `/etc/passwd`, sibling `/tmp/foo` when cwd is `/tmp/fo`).

Why safe:
1. Relative images under the operator cwd still work (`@./shot.png`)
2. Paths under the config dir still work for stored uploads
3. Symlink escape blocked via `realpathSync` on both the file and the roots

Only widen the allowlist if operators need a deliberate third root (document it then).

Closes #545

## Test plan
- [x] `bun test apps/cli/src/image-input.test.ts` (9 pass)
- [ ] From the CLI, `@./some.png` under cwd still attaches; `@/etc/hosts` errors with outside allowed directories

### DIFF
--- a/apps/cli/src/image-input.test.ts
+++ b/apps/cli/src/image-input.test.ts
@@ -1,30 +1,95 @@
-import { describe, expect, test } from "bun:test";
-import { mkdtemp, writeFile } from "node:fs/promises";
+import { afterEach, describe, expect, test } from "bun:test";
+import { realpathSync } from "node:fs";
+import { mkdir, mkdtemp, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
-import { mergeSendInput, parseImageLine } from "./image-input";
+import {
+  mergeSendInput,
+  parseImageLine,
+  resolveAllowedImagePath,
+} from "./image-input";
 
 const tinyPng = Buffer.from(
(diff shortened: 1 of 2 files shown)


ahmadrosid (commented):
LGTM

PR #926, opened 2026-09-08
fix(server): allowlist NAKAMA_PYTHON_BIN for python tools

### LINKED ISSUE #593: [server] python-tool-loader does not enforce a path whitelist for PYTHON_BIN
## Bug
`PYTHON_BIN` from env is configurable; an operator who sets `NAKAMA_PYTHON_BIN=/path/to/some-binary` runs that binary with full host privileges every time the tool is called.

## Files
- `apps/server/src/services/python-tool-loader.ts:78-108`

## Fix
Reject `NAKAMA_PYTHON_BIN` values not on a server allow-list, or document the security implication prominently in the env var name.

## Severity
Medium — operator-driven privilege escalation.

akmalsyrf:
## Summary

`NAKAMA_PYTHON_BIN` must be a bare `python`/`python3…` name or an absolute interpreter under known system prefixes before a custom Python tool spawns.

**Before:** any env value was passed straight to `spawnJsonTool` as `bin`.
**After:** `resolvePythonBin()` allowlists bare names and paths under `/usr/bin`, `/usr/local/bin`, Homebrew bin/Cellar python trees (and linuxbrew equivalents), with a path boundary so `…/Cellar/pythonfoo` cannot sneak through.

Why safe:
1. Default `python3` unchanged — existing installs keep working
2. Reject happens before spawn, so a bad env fails closed with a clear error
3. Homebrew Cellar realpaths are covered so `/opt/homebrew/bin/python3` symlinks still resolve

Only extend the prefix list if a supported distro ships Python elsewhere (e.g. pyenv) and operators need it.

Closes #593

## Test plan
- [x] `bun test apps/server/src/services/python-tool-loader.test.ts` (21 pass)
- [ ] With default env, playground/custom Python tool still runs; with `NAKAMA_PYTHON_BIN=/tmp/python3`, tool run fails with allowlist error

### DIFF
--- a/apps/server/src/services/python-tool-loader.test.ts
+++ b/apps/server/src/services/python-tool-loader.test.ts
@@ -51,6 +51,18 @@ describe("resolvePythonBin", () => {
   test("rejects absolute paths outside the allowlist", () => {
     expect(() => resolvePythonBin("/tmp/python3")).toThrow(/allowlist/i);
     expect(() => resolvePythonBin("/bin/bash")).toThrow(/basename/i);
+    expect(() =>
+      resolvePythonBin("/opt/homebrew/Cellar/pythonfoo/bin/python3")
+    ).toThrow(/allowlist/i);
+  });
+
+  test("accepts Homebrew Cellar python and python@ paths by prefix", () => {
+    expect(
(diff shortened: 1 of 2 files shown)


ahmadrosid (commented):
LGTM

Issue #407, opened 2026-08-21
Add admin ability to disable a user account


vincentri:
> From the incident-response review (audit 2026-08-21). Related: #367 (session governance), #355 (password change keeps sessions alive).

# Summary

A compromised or offboarded account cannot be neutralized. `users` has no disabled state (`schema.sql:240-250`), no delete/disable method or route exists, and changing the password doesn't revoke live sessions (#355). Today the only containment lever is password rotation — which an attacker with an active session can simply re-change.

# Proposed

- `users.disabled_at` timestamp, enforced in auth middleware (login + every authenticated request) and channel pairing checks.
- Platform-admin toggle in member management (API + UI).
- Disabling also revokes the user's browser sessions (builds on #367).

# Acceptance criteria

- [ ] Disabled user fails login and every authenticated request with a clear error.
- [ ] Disabled user's channel pairings stop working.
- [ ] Toggle emits an audit event (#369).

xietsunzao:
Claiming this, scoping to the security-critical piece: `disabled_at` on `users`, enforced in `authenticateRequest` (covers bearer/api-key/browser-session, every authenticated request) and the login route, plus a platform-admin-only disable/enable toggle in `org-members.ts` that also revokes the user's live sessions.

Leaving two acceptance boxes out of this PR as follow-ups:
- Channel pairing enforcement — the discord/telegram/whatsapp auth-stores check a locally-cached config file keyed by platform user id, no `databaseAdapter` lookup at all, so wiring this in means resolving platform-id to nakama user id across three platforms. Real design work, not this PR.
- Audit event — the issue frames this as building on #369, but no audit-log infrastructure exists in the repo yet (`git grep -i audit` is empty). Nothing to hook into.

Web UI toggle also left out, one concern per PR, the API is independently testable.

Draft PR incoming.

fajarhide:
@xietsunzao assigned, thanks!

Issue #369, opened 2026-08-21
Enterprise compliance: append-only audit log of admin & security events


vincentri:
> **Enterprise compliance gap** — identified in the nakama security & compliance audit (2026-08-21). Absence verified by reading all 33 tables in `schema.sql`. Required for SOC 2 / ISO 27001-aligned enterprise self-hosting.

# Summary

There is no audit trail. Role changes, provider-key changes, profile deletions, exports, and token rotations leave no durable record of who did what.

# Why this blocks enterprise compliance

- SOC 2 CC7.2: system monitoring incl. security-event logging is a hard requirement; auditors ask "show me who changed X" — currently unanswerable.
- ISO 27001 A.8.15: event logging.
- GDPR Art. 30 record-keeping support; incident forensics generally.

# Current state (verified)

- `packages/db/sql/schema.sql` (491 lines, 33 tables): no audit/event/activity table of any kind.
- Closest constructs record actors only within their own workflow: `org_memory_proposals.proposed_by_user_id/reviewer_user_id`, `skill_proposals`, `artifact_shares.created_by_user_id`.
- Server logs are raw console.log stdout — not structured, not persisted, contain no reliable actor attribution.

# Proposed approach

1. New append-only `audit_events` table: id, timestamp, actor_user_id, org_id, action, resource_type, resource_id, metadata JSON, request_id. No UPDATE/DELETE code paths.
2. Emit from the service layer for security-relevant mutations: auth events (login failure/success, logout, password change, token rotation), membership/role changes, invite lifecycle, provider/settings changes, profile/org CRUD, data export/import, session revocation.
3. Read API (platform-admin) + retention hook (ties into the retention-policies issue).
4. Never write secrets/tokens into metadata.

# Acceptance criteria

- [ ] Every event class above produces a queryable record with actor attribution.
- [ ] No code path mutates or deletes audit rows.
- [ ] Coverage checklist lives next to the emitter helpers so new routes don't silently skip auditing.

vincentri:
Follow-up to consider once the audit stream exists: optional alerting on security events (login failures spike, token rotation, mass revocations) via the existing notification-webhook delivery machinery.

PR #928, opened 2026-09-08
feat(server): let platform admins disable a user account

xietsunzao:
Ref #407

## Summary

A platform admin can now disable a user's account, cutting off login and every authenticated request instantly, without waiting for their session to expire.

**Before:** the only containment lever was a password reset, which does nothing against an attacker who already has a live session.
**After:** `users.disabled_at`, checked in `authenticateRequest` (bearer, api-key, and browser-session paths) and at login. Disabling also revokes the account's live sessions immediately.

Why safe:
1. Root-cause fix in `authenticateRequest`, the single function every authenticated request already funnels through, not per-route patches
2. Reused `revokeBrowserSessionsForUser`, the same session-kill call `removeMember` and role-downgrade already use
3. Last-admin protection is disabled-status-aware: reusing the existing `assertCanChangeAdminMembership` guard as-is would not have worked here, since disabling a user does not touch the `org_members` row, so it would let every admin be disabled one by one. Added a dedicated check that only counts admins who are not already disabled.

Two acceptance boxes from #407 are explicitly out of this PR:
- **Channel pairing enforcement.** The discord/telegram/whatsapp auth-stores check a locally-cached config file keyed by platform user id, with no database lookup at all. Wiring `disabled_at` in means resolving platform-id to nakama user id across three platforms, real design work on its own.
- **Audit event.** The issue frames this as building on #369, but no audit-log infrastructure exists in the repo yet (`git grep -i audit` is empty). Nothing to hook into.

Web UI toggle is also left out (API-only here), one concern per PR and the API is independently testable.

Route lives under `/v1/platform/orgs/{orgId}/members/{userId}/disable` (and `/enable`), not `/v1/orgs/...`, because the global org-context middleware requires the caller to be a member of the target org, which a platform admin managing another org's account often isn't. Same reason the existing org CRUD platform-admin routes live under `/v1/platform/orgs/...` instead of `/v1/orgs/...`.

## Test plan
- [x] `bun test apps/server/src/http/platform-orgs.test.ts` (9 pass)
- [x] `bun test apps/server/src/http/org-members.test.ts apps/server/src/services/org-service.test.ts` (36 pass)
- [x] `bun test apps/server/src/http` full directory (275 pass, 1 pre-existing unrelated failure confirmed against clean baseline)
- [x] `bun test packages/db` (confirmed the one failure there is a pre-existing Windows file-lock flake, unrelated)
- [x] `bunx tsc --noEmit` and `bun run knip` clean against this change
- [ ] Manually disable a user from a fresh install, confirm their browser tab gets logged out on the next request without a page refresh


apps/server/src/services/org-service.ts
```diff
@@ -584,6 +584,57 @@ export class OrgService {
...
+    if (!member) {
+      throw new NakamaApiError("Not found", 404);
+    }
+
+    // assertCanChangeAdminMembership only counts org_members rows, which a
+    // disable never touches: a disabled admin still counts as "admin" there,
+    // so that guard alone would let every admin be disabled one by one. Count
+    // admins who are actually usable (not already disabled) instead.
+    if (member.role === "admin") {
+      const members = await this.databaseAdapter.listOrgMembers(orgId);
```
fajarhide: `listOrgMembers(orgId)` only sees the org in the URL, but `disabled_at` sits on `users`, so one org's route decides access everywhere. Two orgs, a user who is org A's only admin and a plain member of org B, disabled through org B:

```
disable via org B -> 204   (expected 409, org A now has no usable admin)
```

The same shape reaches the install's only platform admin through any org it merely joined, also 204, and `/enable` needs a platform admin, so that one is unrecoverable without touching the database.

Counting usable admins across every org the user belongs to closes both. Happy to hand over the probe file.

fajarhide (changes requested):
Putting the check in `authenticateRequest` is the right call, and I went through all three of its return paths plus login looking for a fourth way in and did not find one. The last-admin guard is the problem: it counts one org, and `disabled_at` is install-wide.

I checked out the branch and ran a probe against it, output in the inline comment. Nothing in the diff struck me as over-built; the OpenAPI blocks and the `!orgService` 500 are what the rest of that file already does. I did not run a live instance, and I did not look at the channel paths you scoped out.

apps/server/src/services/org-service.ts
xietsunzao: Good catch, fixed in c3dcf29.

`disableMember` now pulls every org the target user administers (`listUserOrganizations`, filtered to `role === "admin"`), and checks usable-admin coverage on each one, not just the org in the URL. The old `member.role === "admin"` gate on the URL's org is gone entirely, since it's exactly why the plain-member-of-org-B path skipped the check.

Added a regression test (`disabling a user via one org is blocked if it would leave another org with no usable admin`) that reproduces your scenario: org A's admin joins org B as a plain member, local-client gets disabled in org A first (leaving the human as org A's sole usable admin), then disabling the human through org B is asserted 409. Confirmed it fails against the old guard (204) and passes against the fix.

apps/server/src/services/org-service.ts
```diff
@@ -584,6 +584,65 @@ export class OrgService {
...
+
+    // disabled_at is install-wide, not per-org, so this has to check every org
+    // the user administers, not just the org named in the URL: a user can be
+    // a plain member of orgId while being the sole admin of a different org,
+    // and disabling never touches org_members rows, so assertCanChangeAdminMembership
+    // (which only looks at one org) would miss that entirely.
+    const memberships =
+      await this.databaseAdapter.listUserOrganizations(userId);
+    const adminMemberships = memberships.filter(
+      (membership) => membership.role === "admin"
```
fajarhide, line 604: This loop only sees orgs where the target is an **admin**, so a platform admin who is merely a member everywhere passes through it untouched. Measured on c3dcf29, with the platform admin added to one org as a plain member:

```
memberships of the target : org_8abacd58:member      (no admin row, so the loop above is empty)
user_local_client         : isPlatformAdmin false
target after the call     : isPlatformAdmin true, disabledAt 2026-09-09T12:08:28Z
POST .../disable          -> 204
```

That is the install's only platform admin disabled, and `/enable` is platform-admin only, so recovery means editing the database. Same shape as the org case you just fixed: `disabled_at` is install-wide, and the guard is scoped narrower than the effect.

The symmetric fix is small: if the target `isPlatformAdmin`, require another platform admin that is not disabled, the same "usable" test you already wrote for org admins. There is no adapter method that lists platform admins today, so that is the one new piece.

fajarhide (changes requested):
The org half is closed, and your test is a better construction than my probe was: disabling the local client first, so the human really is the sole admin, is the state that matters. I rebuilt mine the same way on c3dcf29 and it now answers 409 where it used to answer 204. `apps/server` 1192 pass, CI green.

The second half of that finding is still open, and it is the unrecoverable one. Details inline.

apps/server/src/services/org-service.ts
xietsunzao: Fixed in 0172c79, same shape as the org fix.

Added `listPlatformAdminUsers()` to `DatabaseAdapter` (backed by `SELECT * FROM users WHERE is_platform_admin = 1`), and `disableMember` now checks it after the org-admin loop: if the target `isPlatformAdmin`, requires another platform admin whose `disabledAt` is unset before allowing the disable.

Added a regression test (`disabling the install's only platform admin is blocked even when they are just a plain org member`) reproducing your probe: platform admin is a plain member of an org (no admin row anywhere), disabling through that org's route is asserted 409. Confirmed it fails against the pre-fix code (204) and passes with the fix.

apps/server/src/services/org-service.ts
```diff
@@ -584,6 +584,84 @@ export class OrgService {
...
+          409
+        );
+      }
+    }
+
+    // Same shape as the org-admin check: is_platform_admin is also install-wide
+    // and unrelated to org membership, so a platform admin who is a plain
+    // member (or not a member) everywhere would otherwise slip past the loop
+    // above and could be the install's only platform admin.
+    const targetUser = await this.databaseAdapter.getUserById(userId);
```
fajarhide, line 633: Third variant of the same class, and it sits in `assertCanChangeAdminMembership` (`org-service.ts:923`) rather than here. That one counts admin rows, so it cannot see `disabled_at`, and `updateMember` and `removeMember` both route through it. Once one admin is disabled, the remaining admin can step down and leave the org with nobody usable, the end state the route above refuses.

Org with the human admin plus local-client, which every new org seats. Disable local-client first (allowed, the human is still active), then have the human act on themselves. Measured on 0172c79:

```
POST  /v1/platform/orgs/{org}/members/{human}/disable  -> 409   (correctly refused)
PATCH /v1/orgs/{org}/members/{human} {"role":"member"} -> 200
DELETE /v1/orgs/{org}/members/{human}                  -> 204

  member user_local_client  role=admin  disabledAt=2026-09-10T04:15:38Z
  member user_f4bac686...   role=member disabledAt=null
usable admins left in org: 0
```

Same filter you already wrote closes it: count admins whose `disabledAt` is unset, so `adminCount > 1` means more than one usable admin. A platform admin can `/enable` their way out of this one, so here or as a follow-up is your call.

fajarhide (commented):
@xietsunzao  Both fixes hold up. I rebuilt my probe on 0172c79 and the platform-admin path answers 409 where it answered 204, and reusing the same usable-admin test instead of inventing a second rule for it is the right shape. `listPlatformAdminUsers` is the only new adapter surface, which is the least this could have cost.

One variant of the same bug is still reachable, through the role path rather than the disable path. Details inline. That one is recoverable by a platform admin, so I am leaving this as a comment and treating my earlier changes-requested as answered.

I ran `apps/server/src/http/platform-orgs.test.ts`, `org-members.test.ts` and `org-service.test.ts` on the branch (47 pass). I did not run a live instance, and I did not look at the channel paths you scoped out.

apps/server/src/services/org-service.ts
xietsunzao: Fixed in 5064a04, taking it now rather than deferring.

`assertCanChangeAdminMembership` now resolves each admin row to its user record and counts only those with `disabledAt` unset, instead of the raw `org_members` admin count. `updateMember` and `removeMember` both route through it, so demoting or removing an admin is blocked the moment it would leave zero usable admins, same as the disable route already does.

Added a regression test (`protects the last org admin from removal or demotion when the other admin is disabled`) reproducing your probe: disable local-client first, then try to remove or demote the human admin, both asserted 409. Confirmed it fails against the pre-fix code (resolves instead of rejecting) and passes with the fix.

ahmadrosid (commented):
LGTM

apps/server/src/services/org-service.ts
```diff
@@ -843,8 +921,16 @@ export class OrgService {
...
+    const admins = members.filter((entry) => entry.role === "admin");
+    const adminUsers = await Promise.all(
+      admins.map((entry) => this.databaseAdapter.getUserById(entry.userId))
+    );
+    // disabled_at is install-wide and does not touch org_members rows, so a raw
+    // admin-row count still sees a disabled admin as usable coverage.
+    const usableAdminCount = adminUsers.filter(
+      (user) => user && !user.disabledAt
+    ).length;
+    if (usableAdminCount > 1) {
```
fajarhide, line 933: This counts the target as well, so removing or demoting an admin who is already disabled now gets 409 while a usable admin is still left. That is the likely next step after disabling someone who left. Probe on 5064a04, with the human admin usable and the local client disabled:

```
remove disabled admin -> 409 Cannot remove the last org admin.
demote disabled admin -> 409 Cannot change role of the last org admin.
```

Both resolve on 0172c79. The shape `disableMember` uses fixes it: add `entry.userId !== userId` to the admin filter and return when at least one of the others is usable. With that change both resolve, and all 34 in `org-service.test.ts` still pass, including yours. The two checks are then the same rule, so one helper could serve both if you want it smaller.

fajarhide (commented):
The role path is closed, and your new case fails on 0172c79 and passes on 5064a04 when I run it. The new count overshoots in one direction, details inline. Not blocking: @ahmadrosid, fine to merge on your LGTM and take this as a small follow-up.

fajarhide (approved):

PR #931, opened 2026-09-09
fix(server): enforce request boundary limits

isnaenihidayat:
## Summary

Reject oversized API bodies and invalid import/filter input before services run.

**Before:** every route accepted unbounded bodies, import schemas were documentation-only, and unknown `status` values reached services.
**After:** normal bodies cap at 10 MiB, known imports cap at 140 MiB, six import handlers validate their schemas, and three status filters reject unknown values.

Why safe:
1. Hono rejects oversized bodies before JSON parsing, while Bun enforces the same hard ceiling at the server boundary.
2. The import allowance preserves the documented 100 MiB archive flow after base64 expansion and reuses existing request schemas.

Only extend schema-aware parsing to remaining mutating routes once each route has a real request schema.

## Test plan
- [x] `bun test apps/server/src/http` (284 pass)
- [x] `bun run check`, `bun run knip`, `bun run build`
- [ ] Send an 11 MiB JSON body to a normal route and confirm HTTP 413

Progresses #357

### DIFF
--- a/apps/server/src/http/app.test.ts
+++ b/apps/server/src/http/app.test.ts
@@ -11,7 +11,11 @@ import {
 import { AuthService } from "../services/auth-service";
 import { OrgService } from "../services/org-service";
 import { setupTestConfigDir } from "../test-config-dir";
-import { createHonoApp } from "./app";
+import {
+  createHonoApp,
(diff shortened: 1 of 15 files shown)


ahmadrosid (commented):
LGTM

Issue #780, opened 2026-08-30
feat(integrations): per-org channel credentials + one bot per worker


ahmadrosid:
## Outcome

Each organization configures and owns its own WhatsApp, Telegram, Discord, and Instagram integration. Org A’s bot / phone / Meta app never serves Org B.

**Hard rule:** one channel identity (WhatsApp session, Telegram bot token, Discord bot token, Instagram account) runs on **exactly one worker** at a time. Two workers claiming the same bot/session will break the system (split polling, session kicks, duplicate replies, auth races).

## Before → After

**Before:** One install-wide channel config (`~/.nakama/{whatsapp,telegram,discord}/config.ini`). Every org shares the same bot/token/session; `channel-org` only picks which org a paired user chats in. One worker process per channel per install.

**After:** Each org stores its own channel credentials. Integrations UI reads/writes the active org’s config. Workers own at most one live claim per channel identity; inbound traffic routes only through that owner to the matching org.

## Why

Multi-tenant installs need isolation for channel identity, billing, and secrets — same gap called out in #307 for provider + channel config. Instagram is not in the product yet; ship it org-scoped from day one.

Per-org bots also means multiple live channel identities on one install. Without exclusive worker claims, the same token/session can be started twice (two orgs pasted the same token, restart overlap, multi-replica deploy) and the platform fails hard.

### In
- **WhatsApp** — per-org session / pairing / phone linkage
- **Telegram** — per-org bot token, pairing
- **Discord** — per-org bot token, pairing, invite
- **Instagram** — new channel, org-scoped credentials (Meta app / page / webhook)
- Integrations UI scoped to active org (`X-Org-Id`)
- Org admin can manage that org’s channels; no cross-org overwrite
- Worker routing: inbound message → owning worker → correct org → that org’s profiles/sessions
- **Exclusive claim:** reject or refuse start when another worker already holds the same WhatsApp session / Telegram token / Discord token / Instagram account (across orgs and replicas)

### Out (unless needed to unblock)
- Per-org LLM provider keys (covered by #307 option 2/3)
- Changing tools/MCP platform-global model

## Current anchors

| Piece | Today |
|---|---|
| Config paths | `packages/core/src/{whatsapp,telegram,discord}-config.ts` → one `config.ini` under `getUserConfigDir()` |
| Settings API | `/v1/settings/{whatsapp,telegram,discord}` — no org id on the record |
| UI | `IntegrationsPage` + channel settings cards |
[...]

PR #933, opened 2026-09-09
feat(telegram): per-org bot credentials and one bot per identity

fajarhide:
## Summary

Each org configures its own Telegram bot, and the bridge runs one bot per identity. First channel of #780; Discord, WhatsApp and Instagram are separate PRs on the same layout.

**Before:** one install-wide `~/.nakama/telegram/config.ini`. `setTelegramSettings` took no org id, so any org admin could overwrite another org's bot token, and one bot served every org.
**After:** config resolves per scope (`~/.nakama/orgs/<orgId>/telegram/config.ini`, or the old path for installs that predate this). The settings API reads and writes only the caller's active org, a save is refused when another scope already runs that token, and the bridge starts one bot per configured identity, each with its own heartbeat file. A bot owned by an org binds every chat to that org, so `/org` cannot move a conversation to another tenant.

Why safe:
1. The four new tests were watched failing: with the scope dropped from `getTelegramConfigDir` and the duplicate-token guard removed, all four go red, and green again with them back.
2. Existing installs keep working. A single-org install has its config moved into that org at startup (log line asks for a bridge restart); a multi-org install keeps the shared bot until an admin saves their own, `/org` included.
3. `TELEGRAM_BOT_TOKEN` stays install-wide only, so the env override cannot hand two scopes the same token.

Design pick 3 from the issue, reasoning in https://github.com/ahmadrosid/nakama/issues/780#issuecomment-5596012385.

## Test plan
- [x] `bun run test`, every workspace green (core 741, server 1189, telegram 103)
- [x] new tests confirmed failing against the unscoped code before the fix
- [ ] Not run live. The per-identity claim is covered by a heartbeat test, not by a bridge started twice against two real bot tokens.

### DIFF
--- a/apps/platform/telegram/src/index.ts
+++ b/apps/platform/telegram/src/index.ts
@@ -6,10 +6,7 @@ import {
   log,
 } from "@nakama/core";
 import { hasActiveStreams } from "@nakama/core/channel-active-stream";
-import {
-  ChannelOrgStore,
-  getChannelOrgSelectionPath,
-} from "@nakama/core/channel-org";
+import { ChannelOrgStore } from "@nakama/core/channel-org";
 import { ChannelSessionStore } from "@nakama/core/channel-session-store";
 import {
   ensureServerRunning,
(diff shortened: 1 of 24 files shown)


apps/platform/telegram/src/index.ts
```diff
@@ -95,15 +73,91 @@ try {
...
+    }
+  }, 15_000);
+
+  await Promise.all(running);
+} catch (error) {
+  const message = error instanceof Error ? error.message : String(error);
+  console.error(message);
+  stopAll();
+  // Await before exit: void + process.exit can leave a stale heartbeat file.
+  await Promise.all(started.map((identity) => identity.clearHeartbeat()));
```
ahmadrosid: `stopAll()` already calls `clearHeartbeat()` for every started identity, so this clears each heartbeat a second time just to await completion. Return the cleanup promise from `stopAll()` and use `await stopAll()` here; the signal callback can still call it with `void`. That keeps the cleanup in one place and removes this second pass.

ahmadrosid (commented):
One small simplification in the shutdown path.

apps/platform/telegram/src/index.ts
fajarhide: Done in 309fabb. `stopAll` returns the cleanup promise now, the catch awaits it, and the signal handler still calls it with `void`. Telegram 103 and server 1189 still green.

fajarhide:
@ahmadrosid Pushed `bc42702` in the same spirit while I was in there: `resolveTelegramScopeForOrg` takes the scope now, so the two callers that wrapped it in their own null ternary just call it, and the org listing no longer sorts for nobody. Full suite green.

Issue #358, opened 2026-08-21
Add rate limiting middleware (global + stricter on auth/public routes)


vincentri:
# Summary

No rate limiting exists anywhere in the API. Every endpoint accepts unlimited request volume from any client.

# Evidence (validated against source)

- Middleware chain `apps/server/src/http/app.ts:42-144` contains only security headers (`:61`), auth (`:103`), and org context (`:108`) — no limiter.
- Repo-wide grep for `rate.?limit|throttle|limiter` in `apps/server/src` matches only test fixtures asserting upstream provider 429s.
- `express-rate-limit@8.5.2` exists only as a transitive dep of `@modelcontextprotocol/sdk` (`bun.lock:520,1366`) — never imported by nakama code.
- `POST /v1/auth/login` (`apps/server/src/http/routes/auth.ts:440`) is public and unthrottled; no lockout or failed-attempt tracking exists.
- LLM/provider-spend routes are unthrottled: `POST /v1/sessions/:sessionId/messages` (`sessions.ts:490`), `POST /v1/tools/:toolId/run` (`tools.ts:338`), `/v1/models/discover` (`models.ts:1124`), `/v1/audio/transcribe` (`models.ts:1251`), `/v1/images/generate` (`models.ts:1292`). The usage tracker records stats but enforces no quota.
- Public unauthenticated set (`apps/server/src/http/public-routes.ts:1-32`): setup, setup-import preview/restore, accept-invite, notify webhook, artifact shares — all unthrottled.
- No reverse-proxy `limit_req` in `Dockerfile` / `.docker/`.

# Impact

- Unlimited online password brute-force against login and accept-invite (bcrypt cost 10 ≈ 50–100 ms/attempt is the only friction).
- Unlimited provider spend per compromised (or ordinary) credential on every LLM/image/transcription route.
- Unauthenticated DoS surface during the pre-first-admin bootstrap window.

The Docker image binds port 4310 publicly, so this applies to normal deployments, not just localhost.

# Suggested fix

- Global per-IP limiter as middleware in `app.ts`.
- Stricter buckets: login + accept-invite (e.g. 5–10/min/IP with backoff or temporary lockout); LLM-spend routes per-user/per-org quota; modest caps on public routes.
- While here: flatten the login timing oracle by running a dummy bcrypt compare for unknown emails (response time currently distinguishes existing accounts, `auth.ts:446-457`).

## Verification notes

[...]

PR #934, opened 2026-09-09
feat(server): rate limit the boundary and flatten the login timing oracle

vianbas:
## Summary

An unauthenticated flood now gets 429 with `Retry-After` before it reaches bcrypt or the database, and login stops answering "does this email have an account here".

**Before:** no limiter anywhere in the chain, and an unknown email returned 401 without ever running a password comparison.
**After:** a per-address fixed-window limiter ahead of auth in `createHonoApp`, with a separate smaller budget for the credential routes, and a dummy comparison on the unknown-email path.

Why safe:
1. The key is the real peer address. Bun hands `Server` to `fetch`, and `apps/server/src/index.ts` was dropping it, so it is now passed through as the Hono env. `X-Forwarded-For` is read only when `NAKAMA_TRUST_PROXY` is set, since trusting it with no proxy in front makes the limiter a formality.
2. A request whose address Bun cannot report is not counted at all, so one such caller cannot lock out everyone else. That is also why the existing suite, which calls `app.fetch` without a `Server`, is untouched: 1189 pass before, 1198 after, and the 9 are the new ones.
3. Every new test was scored by mutation rather than by being green. Six changes were applied one at a time: never enforcing the limit, giving the auth routes the general budget, always trusting the forwarded header, never refreshing the window, dropping the address from the key, and removing the dummy comparison. All six fail at least one test, twice.

Only revisit the default budget if a shared address behind NAT starts seeing 429 on ordinary dashboard use; `NAKAMA_RATE_LIMIT_MAX` and `NAKAMA_RATE_LIMIT_AUTH_MAX` tune it without a deploy.

Measured through the real app with a peer address: 60 wrong-password logins answer 10 times 401 then 50 times 429, three runs identical. Without a peer address the same 60 stay 401. The timing gap on login went from roughly three orders of magnitude to none: median 0.02 ms against 47 to 92 ms before, 55.7 against 56.7, 58.7 against 62.5, and 59.0 against 57.3 after, on macOS arm64 with Bun 1.3.14.

Scope: this is the throttling half of #358. The per-user and per-org spend quota is a different concern, needing identity and accounting rather than middleware, so it is not here.

## Test plan
- [x] `bun test apps/server` (1198 pass, 0 fail; `62dc9ed2` is 1189)
- [x] `bun test apps/server/src/http/rate-limit-middleware.test.ts apps/server/src/http/login-timing.test.ts` (9 pass)
- [x] `bun run knip` and `bun x ultracite check` (knip findings identical to main, no lint findings)
- [ ] Get the password wrong 11 times in a row against a running server: the 11th answers 429 with `Retry-After`, and a correct login works again once the window passes

### DIFF
--- a/apps/server/src/http/app.ts
+++ b/apps/server/src/http/app.ts
@@ -6,6 +6,7 @@ import { createAuthMiddleware } from "./auth-middleware";
 import type { ServerOptions } from "./context";
 import { serializeHttpOpenApiSpec } from "./openapi";
 import { createOrgContextMiddleware } from "./org-middleware";
+import { createRateLimitMiddleware } from "./rate-limit-middleware";
 import { registerArtifactShareRoutes } from "./routes/artifact-shares";
 import { registerAuthRoutes } from "./routes/auth";
 import { registerAutomationWorkerSettingsRoutes } from "./routes/automation-worker-settings";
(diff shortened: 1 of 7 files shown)


ahmadrosid (commented):
LGTM

PR #936, opened 2026-09-09
fix(server): preserve split UTF-8 installer output

### LINKED ISSUE #935: fix(server): preserve split UTF-8 characters in installer output
> *This was generated by AI during triage.*

## Bug

`runTimedInstallCommand` decodes each stdout/stderr Buffer separately with `chunk.toString()`. A pipe read may end inside a valid UTF-8 character; decoding each read independently replaces the fragments with U+FFFD.

This affects agent-browser installation progress and failure diagnostics: `installAgentBrowser` forwards the decoded lines to its SSE response and uses the captured output in installation errors. This is diagnostic corruption, not evidence that the installation itself fails.

## Reproduction

Confirmed on current main [62dc9ed2](https://github.com/ahmadrosid/nakama/commit/62dc9ed23e5d6b720b8205c5c177e3b2d21d62db), with Windows/Bun 1.4.0 and Linux/Bun 1.3.14. Run from the repository root:

```ts
import { runTimedInstallCommand } from './apps/server/src/services/cli-package-install';

const result = await runTimedInstallCommand({
  command: process.execPath,
  args: ['-e', `
    process.stdout.write(Buffer.from([0xf0, 0x9f]));
    process.stderr.write(Buffer.from([0xe2]));
    await Bun.sleep(100);
    process.stdout.write(Buffer.from([0x9a, 0x80, 0x0a]));
    process.stderr.write(Buffer.from([0x82, 0xac]));
  `],
  displayCommand: 'split UTF-8 output',
}, console.log);
console.log(result);
```

Expected: stdout is U+1F680 (rocket), stderr is U+20AC (euro), and progress preserves both. Actual: each becomes three replacement characters, despite exit code 0. The delayed writes exercise separate pipe reads; the regression failed twice on Windows and once on Linux before the fix.

## Proposed fix and acceptance

Use the native `setEncoding('utf8')` on both installer pipes so incomplete characters are buffered between reads. Preserve line splitting, final unterminated-line delivery, and timeout/cancellation behavior. One real-child regression should assert both captured output and progress.

Local proof is ready: the regression passes after the fix on both platforms, all nine installer tests pass on Linux, and a smoke check preserves Unicode through progress and failure SSE events. No custom decoder or dependency is needed.

mamenesia:
## Summary

Agent-browser installation diagnostics and CLI version probes preserve Unicode across pipe reads.

**Before:** decoding individual Buffer chunks turns split UTF-8 characters into replacement characters.
**After:** native `setEncoding('utf8')` buffers incomplete characters independently on stdout and stderr.

Why safe:
1. No custom decoder or dependency; line buffering, final-line delivery, and install lifecycle remain unchanged.
2. One real-child regression checks both pipes and captured/progress output, including a final line without a newline; it failed before the fix on Windows and Linux.

Scope includes installer diagnostics and CLI version probes; installation behavior and the bash tool are unchanged. Fixes #935; assignment requested in the issue before opening the draft.

## Test plan
- [x] `bun test apps/server/src/services/cli-package-install.test.ts` on Linux/Bun 1.3.14 (9 pass); new regression also passes on Windows/Bun 1.4.0.
- [x] Targeted Ultracite check and `bun run knip`; smoke checks preserve Unicode through progress/error SSE and retain replacement behavior for incomplete UTF-8 at EOF.
- [x] CI tests, Knip, and React Doctor pass on [the latest head](https://github.com/ahmadrosid/nakama/commit/443666c6ddc76a04a2adc738cec39436a93fe8fd). Refreshed simplification self-review and independent static correctness review found no blockers. A real-executable version-probe smoke check failed before the follow-up and passed afterward for both pipes; 5 existing status/timeout checks pass.

Broader Linux installer/service/SSE run: 21 pass, 1 fail (`install stream emits progress events`: fixture cannot find `bun`); the same failure reproduces with unchanged upstream production code. Windows shell-fixture checks also fail. A later LF-normalized test copy did not resolve the Linux failure. No live package installation or browser UI check was performed.

### DIFF
--- a/apps/server/src/services/cli-package-install.test.ts
+++ b/apps/server/src/services/cli-package-install.test.ts
@@ -43,6 +43,35 @@ const PIPE_HOLDING_PLAN = {
 };
 
 describe("runTimedInstallCommand", () => {
+  test("preserves UTF-8 characters split across stdout and stderr chunks", async () => {
+    const progress: string[] = [];
+    const result = await runTimedInstallCommand(
+      {
+        args: [
+          "-e",
+          `
+          process.stdout.write(Buffer.from([0xf0, 0x9f]));
+          process.stderr.write(Buffer.from([0xe2]));
+          await Bun.sleep(100);
+          process.stdout.write(Buffer.from([0x9a, 0x80, 0x0a]));
+          process.stderr.write(Buffer.from([0x82, 0xac]));
+          `,
+        ],
+        command: process.execPath,
+        displayCommand: "split UTF-8 output",
+      },
+      (message) => progress.push(message)
+    );
+
+    expect(result).toEqual({
+      exitCode: 0,
+      stderr: "\u20ac",
+      stdout: "\ud83d\ude80",
+      timedOut: false,
+    });
+    expect(progress).toEqual(["stdout: \ud83d\ude80", "stderr: \u20ac"]);
+  });
+
   test("gives up on an installer that outlives the timeout", async () => {
     const result = await runTimedInstallCommand(STALLING_PLAN, undefined, {
       timeoutMs: 50,
--- a/apps/server/src/services/cli-package-install.ts
+++ b/apps/server/src/services/cli-package-install.ts
@@ -117,11 +117,11 @@ export async function probeCliVersion(command: string): Promise<{
       resolve({ installed: false, missing: false, version: null });
     }, timeoutMs);
 
-    child.stdout?.on("data", (chunk) => {
-      stdout += chunk.toString();
+    child.stdout?.setEncoding("utf8").on("data", (text: string) => {
+      stdout += text;
     });
-    child.stderr?.on("data", (chunk) => {
-      stderr += chunk.toString();
+    child.stderr?.setEncoding("utf8").on("data", (text: string) => {
+      stderr += text;
     });
     child.once("error", (error) => {
       clearTimeout(timeoutId);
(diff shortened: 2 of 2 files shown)


fajarhide (commented):
@mamenesia  `setEncoding("utf8")` is the right fix, no decoder to maintain and the stream buffers the partial character itself. I checked out the branch and ran your regression both ways: reverted to the previous production file it fails, with your change 9 pass. Nothing in the diff I would cut.

One question. The same pattern sits about 160 lines above in this file, in the version probe:

```
apps/server/src/services/cli-package-install.ts:120
  child.stdout?.on("data", (chunk) => { stdout += chunk.toString(); });
```

You scoped that out on purpose and a version string is ASCII in practice, so the odds of it biting are low. It is also the same one-line change in the same file, which is cheaper to do now than to leave as a known-identical spot for the next reader. Your call, I do not think it blocks.

Separately, and definitely not for this PR: `apps/server/src/tools/bash.ts:274` does `String(chunk)` per chunk with no `setEncoding`, and that output goes to the model rather than to a diagnostic. That one looks worth its own issue.

I did not run a real package install or the browser UI, so the SSE path is only covered by your test and mine.

mamenesia:
@fajarhide Applied the same native decoding to both version-probe pipes in [443666c6](https://github.com/ahmadrosid/nakama/commit/443666c6ddc76a04a2adc738cec39436a93fe8fd).

A real executable smoke check reproduced split UTF-8 corruption separately on stdout and stderr before the change; both return the intact version afterward on Linux/Bun 1.3.14. The 9 installer tests and 5 status/timeout checks pass. The existing install-route fixture still fails to find `bun`, as it did against the earlier unchanged upstream code; targeted lint passes. CI is rerunning on this head.

No version extraction or lifecycle changes. The bash-tool path remains untouched and outside this PR.

Issue #368, opened 2026-08-21
Enterprise compliance: complete data deletion & erasure (org delete, user erasure, disk cleanup)


vincentri:
> **Enterprise compliance gap** — identified in the nakama security & compliance audit (2026-08-21). Absences verified by exhaustive grep and delete-path tracing. Required for GDPR Art. 17 / Art. 5(1)(e) and basic tenancy hygiene.

# Summary

Deletion is structurally incomplete: org deletion does not exist at all, user erasure does not exist, and profile deletion orphans every personal file on disk forever.

# Why this blocks enterprise compliance

- GDPR Art. 17 (right to erasure) and Art. 5(1)(e) (storage limitation) cannot be fulfilled — there is no mechanism to erase a person or an organization.
- Enterprises require tenant offboarding ("delete everything for ACME Corp") as a supported operation.

# Current state (verified)

- **Org deletion: MISSING.** No service method, no route (`platform-orgs.ts:107-267` has GET/POST/PATCH only), no UI. Grep `deleteOrganization|DELETE FROM organizations` → zero matches. Schema's `ON DELETE CASCADE` from `organizations` is unreachable; even manual DB deletion would orphan the entire `~/.nakama/orgs/{orgId}/` tree.
- **User erasure: MISSING.** `removeMember` (`org-service.ts:477-484`) deletes only the membership row. Grep `deleteUser|removeUser` → zero matches. Users + browser_sessions persist indefinitely.
- **Profile deletion: half-done.** `deleteProfile` (`profile-service.ts:345-359`) relies on correct FK cascades (`schema.sql:49,59,77-79`) but performs NO filesystem cleanup — SOUL.md/MEMORY.md, artifacts/, knowledge-base/, skills/, attachments/, avatar remain on disk under `{soulDir}`.
- **Attachment bytes: dead code.** `deleteAttachmentBytes()` (`packages/core/src/attachments/store.ts:48-57`) has zero production callers; `attachments.session_id` FK is SET NULL (`schema.sql:79`), so purged sessions leave attachment files behind.
- Individual skill/avatar/KB deletes DO clean up correctly (`skills-service.ts:542-553`, `profile-service.ts:595-608,643-660`).

# Proposed approach

1. Platform-admin `DELETE /v1/platform/orgs/:orgId`: DB cascade + recursive removal of `~/.nakama/orgs/{orgId}/` (validate path segments against `assertConfigPathSegment`).
[...]

PR #937, opened 2026-09-09
fix(server): remove profile files on delete

isnaenihidayat:
## Summary

Delete a profile → remove its soul, artifacts, attachments, skills, knowledge-base files, and public share snapshots from disk.

**Before:** profile deletion removed database rows and session archives but left the profile workspace and external share snapshots behind.
**After:** `deleteProfileWithHistoryArchives` enumerates and deletes validated share snapshots, then removes the validated profile directory before the database cascade.

Why safe:
1. Both cleanup paths validate their organization/profile boundaries before touching disk.
2. Files are removed before database deletion, so a filesystem error keeps profile, session, and share IDs available for retry.

Organization and user erasure remain separate follow-ups in #368.

## Test plan
- [x] `bun test apps/server/src/services` (551 pass); `bun test packages/db` (101 pass)
- [x] Targeted Ultracite check and `bun run build`
- [ ] Share an artifact, delete its profile, and confirm both workspace and share snapshot are gone

Progresses #368

### DIFF
--- a/apps/server/src/services/session-persistence.ts
+++ b/apps/server/src/services/session-persistence.ts
@@ -2,7 +2,12 @@ import { copyFile, mkdir, open, rm } from "node:fs/promises";
 import { dirname, join } from "node:path";
 import type { AgentChatSession } from "@nakama/agent";
 import type { ChatMessage, ToolDefinition } from "@nakama/core";
-import { createId, getUserConfigDir, jsonSchemaFromZod } from "@nakama/core";
+import {
+  createId,
+  getProfileSoulDir,
+  getUserConfigDir,
+  jsonSchemaFromZod,
+} from "@nakama/core";
 import type { DatabaseAdapter } from "@nakama/db";
 import { z } from "zod";
 
@@ -93,7 +98,12 @@ export function deleteProfileWithHistoryArchives(
           });
         }
       }
-      // Keep session IDs available for retry if filesystem cleanup fails.
+      // Keep the profile and session IDs available for retry when filesystem
+      // cleanup fails instead of leaving unreachable files after the cascade.
+      await rm(getProfileSoulDir(orgId, profileId), {
+        force: true,
+        recursive: true,
+      });
       return db.deleteProfile(profileId);
     }
   );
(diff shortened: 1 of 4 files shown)


apps/server/src/services/session-persistence.ts
```diff
@@ -93,7 +98,12 @@ export function deleteProfileWithHistoryArchives(
           });
         }
       }
-      // Keep session IDs available for retry if filesystem cleanup fails.
+      // Keep the profile and session IDs available for retry when filesystem
+      // cleanup fails instead of leaving unreachable files after the cascade.
+      await rm(getProfileSoulDir(orgId, profileId), {
+        force: true,
+        recursive: true,
+      });
```
fajarhide, line 115: Share snapshots sit outside this directory, in `orgs/{orgId}/artifact-shares/{shareId}/`, and `artifact_shares` cascades on `profile_id`, so the rows go and the files stay. That is exactly the "unreachable files after the cascade" your comment describes. `deleteArtifactShareSnapshot` already exists for it. Worth pulling in here, or a follow-up on #368?

fajarhide (commented):
Cleanup lands in the right place and the new test earns its keep. I checked it out and ran it both ways: reverted to main it fails, with your change it passes. One thing the cascade still leaves on disk.

fajarhide (approved):
The share cleanup is right, and `listArtifactSharesForProfile` taking the org id rather than filtering later is the shape I would have asked for.

I ran it again on 9088566: removed the new loop and your test goes red (`removes artifact share snapshots before their rows cascade`), put it back and it is 41 pass. Full `apps/server` 1215 pass, `packages/db` 101 pass. I also checked the case that worried me, a share whose file is already gone, and `deleteArtifactShareSnapshot` swallows ENOENT, so a revoked share does not break the delete.

One thing I noticed and am deliberately not asking you to fix: unlinking the snapshot leaves the empty `artifact-shares/<shareId>/` directory behind. Revoke has always done that too, so it is not this PR's to carry.

Reads ready to me. @ahmadrosid over to you.

apps/server/src/services/session-persistence.ts
isnaenihidayat: Good catch — pulled this into the PR in 9088566e. Profile deletion now lists every share for the profile and calls `deleteArtifactShareSnapshot` before the database cascade. The regression was red before the cleanup and is green now; it also keeps filesystem failures retryable while the share rows still exist.

ahmadrosid:
LGTM

PR #942, opened 2026-09-09
feat(providers): add Vercel AI Gateway integration

### LINKED ISSUE #391: feat(providers): direct Vercel AI Gateway integration (optional)
## Summary

Add **Vercel AI Gateway** as a direct provider option — one Vercel key → many labs' models, billed through Vercel.

|---|---|---|---|---|
| Global | `https://ai-gateway.vercel.sh/v1` ⚠️ verify | `vercel_ai_gateway` | `VERCEL_AI_GATEWAY_API_KEY` | Vercel AI Gateway |

OpenAI-compatible → DeepSeek template.

**Scope caveat:** this is structurally an aggregator/gateway (same category as OpenRouter, which is already integrated). Include only if the "I already pay Vercel, one invoice" story is wanted. Capabilities are passthrough — they depend on the underlying model routed to, so catalog entries should mark conservatively and defer to each model's real capabilities.

## Capability matrix

| Capability | Status |
|---|---|
| Chat vision | ⚠️ passthrough — depends on routed model |
| Audio transcription | ⚠️ passthrough — depends on routed model |
| Image generation | ⚠️ passthrough — depends on routed model |

Standard six: `ProviderName` union · `apiKeyEnvVarForProvider` · `create.ts` case · `models.ts` curated entries · `pricing.ts` · tests.

## Acceptance criteria

- [ ] Endpoint confirmed against current Vercel docs and recorded here
- [ ] `vercel_ai_gateway` card with own key field and default base URL
- [ ] Curated entries show correct context window/pricing for gateway-routed models

akmalsyrf:
## Summary

Settings → LLM providers gains **Vercel AI Gateway** — one Vercel key, many labs' `creator/model` ids via `https://ai-gateway.vercel.sh/v1`, with Gateway `reasoning` wired and shortlists that survive restart.

**Before:** gateway only as a hand-rolled Custom OpenAI-compatible endpoint; thinking toggle was a no-op on the wire; shortlists dropped on reload.
**After:** first-party `vercel_ai_gateway` card with `VERCEL_AI_GATEWAY_API_KEY`, curated shortlist, Chat Completions `reasoning` object + `message.reasoning` parse, and `models_json` reload.

Why safe:
1. Same Together/DeepSeek OpenAI-compat factory — Bearer auth matches Vercel docs
2. Capability flags stay opt-in per catalog row; reasoning body only when UI enables thinking
3. Unit coverage for parse/env, routing, reasoning round-trip, catalog, and UI opt-in

Only re-add a live cassette if gateway response quirks show up in production that mocks miss.

Closes #391

## Test plan
- [x] `bun test packages/core/src/provider-resolution.test.ts packages/core/src/user-config.test.ts apps/server/src/providers/create.test.ts apps/server/src/providers/models.test.ts apps/web/src/lib/models.test.ts` (120 pass)
- [ ] Add Vercel AI Gateway in Settings with a real key → chat with `openai/gpt-4o-mini`; enable thinking on `openai/gpt-5` and confirm reasoning text appears

---

HARNESS

### DIFF
--- a/apps/server/src/providers/openai/index.ts
+++ b/apps/server/src/providers/openai/index.ts
@@ -369,6 +369,9 @@ async function buildChatCompletionRequestBody(options: {
     ...(provider === "deepseek"
       ? buildDeepSeekThinkingBody(options.thinking)
       : {}),
+    ...(provider === "vercel_ai_gateway"
+      ? buildVercelReasoningBody(options.thinking)
+      : {}),
     ...(provider === "perplexity" && options.thinking?.enabled
       ? {
           reasoning_effort: normalizeThinkingEffort(options.thinking.effort),
@@ -444,6 +447,22 @@ function buildDeepSeekThinkingBody(
   };
 }
 
+/** AI Gateway Chat Completions extension: top-level `reasoning` object. */
+function buildVercelReasoningBody(
+  thinking: ProviderChatOptions["thinking"] | undefined
+) {
+  if (!thinking?.enabled) {
+    return {};
+  }
+
+  return {
+    reasoning: {
+      effort: normalizeThinkingEffort(thinking.effort),
+      enabled: true,
+    },
+  };
+}
+
 function readReasoningContent(
   value: unknown,
   options?: { preserveWhitespace?: boolean }
@@ -456,7 +475,9 @@ function readReasoningContent(
   const direct =
     typeof record.reasoning_content === "string"
       ? record.reasoning_content
-      : undefined;
+      : typeof record.reasoning === "string"
+        ? record.reasoning
+        : undefined;
 
   if (direct === undefined) {
     return;
--- a/apps/server/src/providers/openai/index.ts
+++ b/apps/server/src/providers/openai/index.ts
@@ -152,6 +152,10 @@ function providerLabel(providerName: ProviderName): string {
     return "Together AI";
   }
 
+  if (providerName === "vercel_ai_gateway") {
+    return "Vercel AI Gateway";
+  }
+
   if (providerName === "mistral") {
     return "Mistral";
   }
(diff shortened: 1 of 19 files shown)


ahmadrosid (commented):
The integration stays close to the existing provider flow. Two small simplifications below.

apps/server/src/providers/openai/index.ts
```diff
@@ -440,6 +447,22 @@ function buildDeepSeekThinkingBody(
   };
 }
 
+/** AI Gateway Chat Completions extension: top-level `reasoning` object. */
+function buildVercelReasoningBody(
```
ahmadrosid: This helper has one caller and only wraps an enabled check around an object. Inline the reasoning object in the existing spread, guarded by `provider === "vercel_ai_gateway" && options.thinking?.enabled`, like the Perplexity branch beside it. Both streaming and non-streaming requests already share that builder.

packages/core/src/user-config.ts
```diff
@@ -638,15 +639,11 @@ function loadProvidersFromSections(
...
-      type === "openrouter" ||
-      type === "cerebras" ||
-      type === "fireworks" ||
-      type === "ollama" ||
-      type === "opencode_go"
-        ? parseCustomModelsJson(values.models_json)
-        : undefined;
+    // Writer persists models_json for every type that has a shortlist/catalog
+    // override; load any present JSON so restarts keep shortlists.
+    const customModels = values.models_json?.trim()
```
ahmadrosid: `parseCustomModelsJson` already returns undefined for missing or blank input. This can just be `const customModels = parseCustomModelsJson(values.models_json);` — the extra trim check and ternary duplicate the parser's guard.

apps/server/src/providers/openai/index.ts
akmalsyrf: Done — inlined beside the Perplexity spread; streaming/non-streaming still share that builder.

packages/core/src/user-config.ts
akmalsyrf: Done — now just `parseCustomModelsJson(values.models_json)`.

ahmadrosid (approved):
LGTM

PR #943, opened 2026-09-09
feat(providers): add Xiaomi MiMo integration

### LINKED ISSUE #382: feat(providers): direct Xiaomi MiMo integration
## Summary

Add **Xiaomi MiMo** as a direct first-party provider.

|---|---|---|---|---|
| TBD | ⚠️ verify Xiaomi's developer platform docs at implementation | `xiaomi` | `XIAOMI_API_KEY` | Xiaomi MiMo |

OpenAI-compatible (verify exact path/shape) → follows the existing DeepSeek template. Today MiMo is reachable only via the opencode-go aggregator (`mimo-v2.5`, `mimo-v2.5-pro`).

**Endpoint is unconfirmed** — this ticket's first task is locating Xiaomi's official OpenAI-compatible base URL and documenting it here before wiring anything.

## Capability matrix (verify at implementation)

| Capability | Status |
|---|---|
| Chat vision (image input) | ❌ expected text-only (verify) |
| Audio transcription (STT) | ❌ expected none |
| Image generation | ❌ expected none |

If any capability turns out to exist, update this matrix and opt into the corresponding settings; otherwise keep the provider chat-only.

1. `ProviderName` union (`@nakama/core`) — add `"xiaomi"`
2. `apiKeyEnvVarForProvider` (`@nakama/core`) — `XIAOMI_API_KEY`
3. `create.ts` case → `createOpenAIProvider({ baseUrl, providerName })`
4. `apps/server/src/providers/models.ts` — curated MiMo entries with `contextWindow`, `maxOutputTokens`, USD pricing (CNY→USD conversion noted in PR), `supportsVision: false` unless verified otherwise
5. `pricing.ts`
6. Tests: extend `models.test.ts`; live-call tests named `*.llm.test.ts` with committed MSW cassettes

## Acceptance criteria

- [ ] Official base URL confirmed and recorded in this issue
- [ ] `xiaomi` appears as its own card with API-key field and default base URL
- [ ] Curated entries show correct context window, pricing, and vision flags
- [ ] Sending an image to a MiMo entry produces the clean non-vision fallback, not an error

akmalsyrf:
## Summary

Settings → LLM providers gains **Xiaomi MiMo** — pay-as-you-go chat via `https://api.xiaomimimo.com/v1`, with thinking control and shortlists that survive restart.

**Before:** MiMo only via OpenCode Go aggregator (`opencode-go/mimo-v2.5*`); thinking/shortlists incomplete or dropped on reload; Pro list price was the OpenCode Go 4× markup.
**After:** first-party `xiaomi` card with `XIAOMI_API_KEY`, curated `mimo-v2.5-pro` / `mimo-v2.5` at official pay-as-you-go rates, `thinking.type` on every request, and `models_json` reload for **all** catalog shortlist providers (not only Xiaomi).

Why safe:
1. Same Together/DeepSeek OpenAI-compat factory — Bearer auth matches Xiaomi’s OpenAI SDK examples
2. Vision opt-in on Pro; both models `supportsThinking`; Pro price + mimo-v2.5 max output match Xiaomi docs
3. Unit coverage for parse/env, routing, thinking body, shortlist round-trip, and UI opt-in

Only revisit auth headers if a live key rejects Bearer and requires Xiaomi’s alternate `api-key` header.

Closes #382

## Test plan
- [x] `bun test packages/core/src/provider-resolution.test.ts packages/core/src/user-config.test.ts apps/server/src/providers/create.test.ts apps/server/src/providers/models.test.ts apps/web/src/lib/models.test.ts` (121 pass)
- [ ] Add Xiaomi MiMo in Settings with a real `sk-` key → chat with `mimo-v2.5-pro`; toggle thinking on/off; send an image on Pro and confirm non-vision fallback

---

HARNESS


apps/server/src/providers/models.ts
```diff
@@ -350,6 +350,29 @@ const BASE_MODELS: ProviderModelOption[] = withVisionDefaults([
     supportsThinking: true,
     supportsVision: true,
   },
+  {
+    contextWindow: 1_048_576,
+    default: true,
+    id: "mimo-v2.5-pro",
+    inputPerMillionUsd: 1.74,
```
fajarhide: Official pay-as-you-go for `mimo-v2.5-pro` is $0.435 in and $0.87 out ([pricing page](https://mimo.mi.com/docs/en-US/price/pay-as-you-go), effective 2026-08-06), so this is 4x. `1.74 / 3.48` is what `opencode-go/mimo-v2.5-pro` charges at :522, which looks like the source. It reaches the usage cost through `llm-usage-tracker.ts:91`:

```
mimo-v2.5-pro  1M in + 1M out: branch $5.220, official $1.305, ratio 4.00x
mimo-v2.5      1M in + 1M out: branch $0.420, official $0.420, ratio 1.00x
```

apps/server/src/providers/models.ts
```diff
@@ -350,6 +350,29 @@ const BASE_MODELS: ProviderModelOption[] = withVisionDefaults([
...
+    outputPerMillionUsd: 3.48,
+    provider: "xiaomi",
+    supportsThinking: true,
+    supportsVision: false,
+  },
+  {
+    contextWindow: 1_048_576,
+    id: "mimo-v2.5",
+    inputPerMillionUsd: 0.14,
+    maxOutputTokens: 32_768,
```
fajarhide: The [model page](https://mimo.mi.com/models/en-US/mimo-v2.5) lists 128K max output for `mimo-v2.5` too, so `131_072` like Pro. Only compaction reads this (`agent-service.ts:4012`), so nothing gets truncated, it just reserves less than it could.

apps/server/src/providers/models.test.ts
```diff
@@ -115,6 +115,28 @@ describe("resolveModel", () => {
...
+  test("resolves catalog models for Xiaomi MiMo", () => {
+    expect(resolveModel("xiaomi", "mimo-v2.5-pro")).toBe("mimo-v2.5-pro");
+    expect(resolveModel("xiaomi", "mimo-v2.5")).toBe("mimo-v2.5");
+    expect(getDefaultModel("xiaomi")).toBe("mimo-v2.5-pro");
+    expect(getModelById("mimo-v2.5-pro")?.supportsThinking).toBe(true);
+    expect(getModelById("mimo-v2.5")?.supportsThinking).toBe(true);
+    expect(getModelById("mimo-v2.5-pro")?.supportsVision).toBe(false);
+    expect(getModelById("mimo-v2.5")?.supportsVision).toBe(true);
+    expect(getModelById("mimo-v2.5-pro")?.contextWindow).toBe(1_048_576);
+    expect(getModelById("mimo-v2.5")?.maxOutputTokens).toBe(32_768);
```
fajarhide: This pinned the 32_768 above, so fixing the catalog means editing the test too. The vision half also repeats the `modelSupportsVision` case at :365. You could keep the `resolveModel` / `getDefaultModel` lines and drop the field asserts with the same confidence.

packages/core/src/user-config.ts
```diff
@@ -638,15 +638,12 @@ function loadProvidersFromSections(
...
-      type === "cerebras" ||
-      type === "fireworks" ||
-      type === "ollama" ||
-      type === "opencode_go"
-        ? parseCustomModelsJson(values.models_json)
-        : undefined;
+    // Writer persists models_json for every type that has a shortlist/catalog
+    // override; load any present JSON so restarts keep OpenAI/DeepSeek/Xiaomi/…
+    // shortlists (not only discovery / OpenRouter / Cerebras / …).
+    const customModels = values.models_json?.trim()
```
fajarhide, line 644: Good catch, and bigger than Xiaomi: on current main every catalog type loses its shortlist on reload. Same save-then-load probe against both:

```
            main 3f1e3af0   this branch
deepseek    []              ["shortlisted-model"]
together    []              ["shortlisted-model"]
openai      []              ["shortlisted-model"]
anthropic   []              ["shortlisted-model"]
mistral     []              ["shortlisted-model"]
```

Worth a line in the PR body, or its own PR, since it changes restart behavior for every provider with a shortlist rather than only the new one.

fajarhide (commented):
Reusing `createOpenAIProvider` keeps this thin, and the wire shape matches Xiaomi's own [Chat Completions doc](https://mimo.mi.com/docs/en-US/api/chat/openai-api): Bearer is accepted, `thinking: {type}` is the real parameter with both models defaulting to enabled, and the adapter already sends `reasoning_content` back on tool-call turns, which the doc recommends in thinking mode. The catalog numbers are where it drifts from the official pages, inline. The Pro price is the one I'd fix before merge.

I checked the registration map (every list that names `together` also names `xiaomi`), ran your five test files (121 pass), and reverted the production side to confirm the new tests fail without it (10 fail). I have no MiMo key either, so the live chat and cassette from #382's acceptance criteria are still open.

apps/server/src/providers/models.ts
akmalsyrf: Fixed — Pro is now $0.435 / $0.87 (official pay-as-you-go). The 1.74 / 3.48 figures were copied from the OpenCode Go catalog row.

apps/server/src/providers/models.ts
akmalsyrf: Fixed — `mimo-v2.5` maxOutputTokens is 131_072 to match the model page.

apps/server/src/providers/models.test.ts
akmalsyrf: Done — kept resolveModel / getDefaultModel; dropped the field asserts (vision stays in the modelSupportsVision case).

packages/core/src/user-config.ts
akmalsyrf: Agreed — called out in the PR body. The reload fix applies to every catalog shortlist provider (openai/deepseek/together/…), not only Xiaomi.

PR #944, opened 2026-09-09
fix(auth): validate JSON request bodies

### LINKED ISSUE #357: Add runtime request-body validation + body size limits

puspoaditya:
## Summary

Auth JSON endpoints now reject invalid request bodies with HTTP 400 before auth services run.

Fixes #357.

**Before:** TypeScript-only request casts let malformed shapes reach handlers and could throw a 500.
**After:** Each mutating auth handler passes its Zod schema into `readJson` for runtime validation.

Why safe:
1. Schemas mirror the existing auth request contracts and handler fields
2. `readJson` keeps malformed JSON mapped to the same 400 response
3. HTTP regressions cover wrong-typed, missing required, and malformed setup payloads

Only rework this if an existing valid auth client payload is rejected by the new schemas.

## Test plan
- [x] `bun test apps/server/src/http/app.test.ts` (39 pass)
- [x] `bun test apps/server/src` (1,216 pass)
- [x] `bun run --filter @nakama/server build`
- [ ] Send an invalid body to one authenticated auth endpoint and confirm it returns 400 without side effects.

### DIFF
--- a/apps/server/src/http/app.test.ts
+++ b/apps/server/src/http/app.test.ts
@@ -1165,6 +1165,60 @@ describe("createHonoApp", () => {
     });
   });
 
+  describe("auth request bodies", () => {
+    test("rejects wrong-typed setup fields before invoking auth services", async () => {
+      const app = createHonoApp(createServerOptions());
+      const response = await app.fetch(
+        new Request("http://localhost:4310/v1/auth/setup", {
+          body: JSON.stringify({
+            admin: {
+              email: "admin@example.com",
+              name: true,
+              password: "password123",
+            },
+            organization: { name: "Acme", slug: "acme" },
+          }),
+          headers: { "Content-Type": "application/json" },
+          method: "POST",
+        })
+      );
+
+      expect(response.status).toBe(400);
+    });
+
+    test("rejects missing required setup fields", async () => {
+      const app = createHonoApp(createServerOptions());
+      const response = await app.fetch(
+        new Request("http://localhost:4310/v1/auth/setup", {
+          body: JSON.stringify({
+            admin: {
+              email: "admin@example.com",
+              name: "Admin",
+            },
+            organization: { name: "Acme", slug: "acme" },
+          }),
+          headers: { "Content-Type": "application/json" },
+          method: "POST",
+        })
+      );
+
+      expect(response.status).toBe(400);
+    });
+
+    test("keeps malformed auth JSON as a bad request", async () => {
+      const app = createHonoApp(createServerOptions());
+      const response = await app.fetch(
+        new Request("http://localhost:4310/v1/auth/setup", {
+          body: "{",
+          headers: { "Content-Type": "application/json" },
+          method: "POST",
+        })
+      );
+
+      expect(response.status).toBe(400);
+    });
+  });
+
   describe("platform admin routes", () => {
     test("allows profile list for org members but blocks profile management", async () => {
       const options = createServerOptions();
--- a/apps/server/src/http/routes/auth.ts
+++ b/apps/server/src/http/routes/auth.ts
@@ -76,6 +76,33 @@ export function registerAuthRoutes(app: HonoApp, options: ServerOptions): void {
     .object({ error: z.string() })
     .openapi("ApiErrorResponse");
 
+  const setupAuthSchema = z
+    .object({
+      admin: z.object({
+        email: z.string(),
+        name: z.string(),
+        password: z.string(),
+        phone: z.string().optional(),
+      }),
+      organization: z.object({
+        name: z.string(),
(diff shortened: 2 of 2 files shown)


ahmadrosid (commented):
LGTM

PR #946, opened 2026-09-10
fix(server): skip platform tools when profile has no tools

isnaenihidayat:
## Summary

A profile with zero tools answers with an empty tool list — no platform helpers injected.

**Before:** stripping every tool from a profile still left 16 platform tools (automation, workflow, todo, question, skill_manage, org-memory, session-history, discord artifacts) in every chat turn, so small models hallucinated fake tool calls.
**After:** `resolveProfileTools` returns `[]` when builtin + MCP + Composio are all empty; every platform group is gated on `hasOwnTools`.

Why safe:
1. Profiles with ≥1 own tool get exactly the same groups as before — only the zero-tool case changes
2. Skill_manage injection test updated to assign one own tool first, keeping the manage-skills gate covered
3. New test pins the no-tools contract both ways (empty list → no groups; one tool → groups return)

Only re-add platform groups for zero-tool profiles if a downstream caller turns out to depend on them (none do today: automation/workflow/sub-agent execution paths already omit most groups).

## Test plan
- [x] `bun test apps/server/src/services/agent-service.test.ts` (29 pass)
- [ ] Manual: strip all tools from a profile, send a chat turn, confirm the model tool list is empty

### DIFF
--- a/apps/server/src/services/agent-service.test.ts
+++ b/apps/server/src/services/agent-service.test.ts
@@ -737,6 +737,18 @@ describe("AgentService skill_manage injection", () => {
   test("injects skill_manage for web/cli only when manage-skills is assigned", async () => {
     const db = createInMemoryDatabaseAdapter();
     await db.upsertProfile(createDefaultProfile());
+    // Give the profile at least one own tool so the platform groups (incl.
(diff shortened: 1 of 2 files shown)


ahmadrosid (commented):
LGTM

Issue #478, opened 2026-08-26
[security] /v1/system/web-public-url lets any org admin redirect OAuth callbacks


vincentri:
## Bug

`PUT /v1/system/web-public-url` only requires `requireOrgAdminFromContext` and persists `body.webPublicUrl` to user config. `isValidBaseUrl` only checks `http`/`https`. After saving, both `resolveComposioCallbackBaseUrl` and `resolveArtifactShareBaseUrl` prefer this value when the request-derived URL is loopback.

A malicious or compromised org admin can pin a host they control, and every subsequent OAuth flow initiated from a channel worker (which typically has no `Origin`/`Referer`) will redirect the code to that host.

## Files

- `apps/server/src/http/routes/system.ts:199-215`
- `persistWebPublicUrl`

## Impact

- Same OAuth-completion hijack as the header-injection bug, but via a persistent config knob
- Stealthier: persists across restarts

## Fix

Restrict setting the web public URL to platform admins (it's deployment-wide), or scope it per-org but require explicit confirmation and only allow it once during setup.

## Severity

Critical — security, privilege escalation.

fajarhide:
Two things here, and only one of them is a bug.

The header path is real and #712 fixes it. `resolveComposioCallbackBaseUrl` preferred `clientOrigin`, `Origin`, `Referer` and `X-Forwarded-Host` over the configured URL, and validated none of them. After that change a configured `NAKAMA_WEB_PUBLIC_URL` wins, so the knob no longer competes with a header.

The guard is a product decision rather than a defect. `PUT /v1/system/web-public-url` requires org admin, which is the level PR #305 settled for every workspace-global setting after the same question came up for the provider `baseUrl`: `SettingsPage` and `IntegrationsPage` render those cards for `activeOrg.role === "admin"`, so anything stricter takes away access people have today. Seventeen mutating routes sit at that level. Raising this one alone would leave the setting reachable through its neighbours and would be inconsistent for no gain.

So the open question is whether workspace-global config should require platform admin across all of those routes. That is worth a decision, and it is yours rather than mine. Say the word and I will move the whole set in one PR, along with the UI that renders the cards. If the answer is no, this issue is covered by #712 and can close with it.

fajarhide:
PR #772 takes the half of this that needs no decision from you.

[...]

PR #951, opened 2026-09-10
feat(mcp): browser OAuth authorization for remote MCP servers

### LINKED ISSUE #477: [security] OAuth callback URL injection via X-Forwarded-Host / Origin
## Bug

OAuth callback URL is derived from attacker-controlled request headers (`X-Forwarded-Host`, `Origin`, `Referer`) when no browser header is present. Channel workers (Telegram/WhatsApp/Discord/automation) typically don't send `Origin`/`Referer`, so the caller's `X-Forwarded-Host` wins and the OAuth provider redirects the authorization code to the attacker's host.

The same path affects `resolveArtifactShareBaseUrl`, so public share URLs can be minted under an attacker-controlled host too.

## Files

- `apps/server/src/services/composio-callback-url.ts:59-89`
- `apps/server/src/services/composio-service.ts:460-470`
- `apps/server/src/services/artifact-share-service.ts:130-138`

## Impact

- Composio: attacker attaches their own Composio account to the victim's connection
- Artifact shares: attacker can mint "public" share URLs under a phishing domain

## Fix

Pin the callback base to `process.env` (`NAKAMA_WEB_PUBLIC_URL` / `NAKAMA_PUBLIC_URL`) or `~/.nakama/web-public-url`. If header-derived URL must remain a fallback, reject when `X-Forwarded-Host` differs from a configured allow-list and require proto agreement with the listener URL.

## Severity

Critical — security, auth bypass.

### LINKED ISSUE #950: feat(mcp): browser OAuth authorization for remote MCP servers
## Problem

Remote MCP servers increasingly authenticate with OAuth 2.1 (the MCP authorization spec) instead of a static API key: Notion, Linear, Sentry, Atlassian, Cloudflare, GitHub's remote MCP. Every other MCP client handles this by opening the authorization URL in a browser and finishing the flow on a callback URL (Claude Code, Hermes, OpenClaw all do exactly that, and OpenClaw hands the callback link back to the user to complete the connection).

Nakama has no such mechanism. `apps/server/src/services/mcp-client-manager.ts` builds `StreamableHTTPClientTransport` with `requestInit.headers` only and never passes an `authProvider`, and there is no OAuth callback route for MCP. So registering one of those servers is impossible: `POST /v1/mcp/servers` fails with `422 Could not connect MCP server "notion": HTTP 401`, and the operator has nowhere to sign in.

Static headers only cover servers that accept a long-lived token. They cannot cover a per-tenant OAuth grant that needs consent and refresh.

## Current behavior

1. **Agent -> System -> MCP -> Add server**, HTTP, `https://mcp.notion.com/mcp`
2. Test connection / Save
3. `Could not connect MCP server: HTTP 401`, nothing is persisted, no way forward

## Proposed

The MCP TypeScript SDK already ships the client half of the flow (`OAuthClientProvider`, `auth()`, `StreamableHTTPClientTransport({ authProvider })`), so this is wiring plus persistence, not a new protocol implementation.

1. **`OAuthClientProvider` backed by the server record.** Client registration, tokens, PKCE verifier and the CSRF state live in the stored `config.oauth` of the `mcp_servers` row. No migration: `config` is already an opaque JSON blob, and `redactMcpConfig()` returns only `url` and `headers`, so the grant never reaches the API response.
2. **Public callback route** `GET /v1/mcp/oauth/callback/:serverId?code&state`, added to `isPublicRouteRequest()` next to the Composio callback. `redirect_uri` is built from the same `resolveComposioCallbackBaseUrl()` helper the Composio flow already uses, so a configured `webPublicUrl` stays the single source of truth for what a redirect URI may point at. State is a single-use nonce compared against the stored one.
3. **Hand the link to the operator.** `POST /v1/mcp/servers` and `.../connect` answer `200 { server, authorizationUrl }` instead of `422` when the server demands authorization; the server row is persisted as `disconnected` with a `lastError` that says authorization is pending. The dashboard opens `authorizationUrl` in a new tab and polls until the row flips to `connected`.
4. **Callback completes the connection**: exchange the code with `auth()`, persist tokens, connect, cache tools, render a "connected, you can close this tab" page like the Composio callback does.

Token refresh comes for free: the transport keeps the provider for the life of the connection, and startup reconnect rebuilds it from the stored grant.

## Scope

- HTTP transport only; stdio is unchanged
- Dynamic client registration only. Servers that require a manually issued `client_id` / `client_secret` are a follow-up
- The grant is stored per server record, so it is shared by every profile the server is assigned to, exactly like the `Authorization` header is today. Per-user grants are a separate (bigger) question
- Docs: an OAuth section in `docs/website/content/docs/mcp.mdx`

## Test plan

- Fake OAuth authorization server plus a real SDK MCP server behind a Bearer check, served from `Bun.serve`: connect returns an `authorizationUrl`, the callback exchanges the code, tokens land in the record and tools are cached
- A callback with a mismatched `state` is refused
- `isPublicRouteRequest` covers the new callback and nothing else under `/v1/mcp`

## Open question

@ahmadrosid dynamic client registration only for the first pass, and one grant per server record (shared across assigned profiles) rather than per user. Both are the smaller change and match how MCP headers already behave. Say if you want per-user grants instead and I will scope that separately.

I am taking this one.

yusupsupriyadi:
## Summary

image

Register a hosted MCP server that signs in with OAuth (Notion, Linear, Sentry), approve it in a browser, and Nakama connects and caches its tools.

**Before:** `POST /v1/mcp/servers` on such a server answered `422 Could not connect: HTTP 401` and persisted nothing, because the HTTP transport only ever sent static headers.
**After:** the row is kept as the new `needs_auth` status with an `authorizationUrl`; the operator clicks the link, and the public `GET /v1/mcp/oauth/callback/:serverId` exchanges the code (`McpService.completeOAuth`) and connects.

Why safe:
1. No OAuth is attempted unless a `callbackBaseUrl` is passed and the transport is `http`, so stdio and header-authenticated servers take the exact path they take today
2. The grant lives in the existing `config` JSON (no migration) and `redactMcpConfig` already narrows that to `url` and `headers`, so tokens cannot leave through the API
3. The callback is GET-only, is guarded by a single-use `state`, and derives its `redirect_uri` from the configured public URL or the host the request landed on, never from the provider's `Origin` or `Referer` (the hole closed in #477 / #478)

Only follow up if a provider needs a pre-issued `client_id` / `client_secret`: this pass registers dynamically and has no field to paste one into.

Fixes #950

Related, not fixed: #477 and #478 (both closed) set the rule this callback follows for redirect URIs; #329 tracks the wider SaaS-integration UX question that Composio answers today.

### What the operator sees

- **Add server** asks for the server type up front: **HTTP** (URL, optional key header), **Sign-in** (URL, no header field, the provider issues the credential) or **Command**. Finding the OAuth flow no longer means failing a connection first, and picking a type clears the other type's fields, because the saved request was built from whichever field still had text rather than from the tab on screen
- **Test connection** on such a server answers "This server signs in with a browser" in amber, not a red `HTTP 401`, and only when the URL actually advertises OAuth (protected resource metadata, then authorization server metadata, the same pair `auth()` asks for)
- **Add and sign in** opens a dialog with the provider link as an anchor. A `window.open` fired after the request that produced the URL is what popup blockers stop, and a blocked popup is indistinguishable from nothing happening
- The row shows **Sign-in required** with a visible **Sign in** button rather than red error text and a way out hidden in the overflow menu; the dialog closes itself once the callback has connected the server

## Test plan
- [x] `bun test apps/server/src/services/mcp-oauth.test.ts` (3 pass, fake authorization server in front of a real SDK MCP server: 401 to discovery to dynamic registration to authorization URL to code exchange to connected with tools cached, plus an unreachable host that must not be reported as a sign-in requirement)
- [x] `bun test` on the MCP, HTTP and system-page suites (65 pass)
- [x] `bun run knip`, `ultracite check apps/server/src apps/web/src packages/core/src`, `bun run --filter @nakama/web build`
- [x] `git merge-tree` against `origin/main` and my only other open PR (#882): clean, though both touch `app.ts`, `shared.ts`, `index.ts` and `contract.ts`
- [x] `bun test apps/web/src` and the MCP/HTTP server suites (469 pass)
- [x] Add `https://mcp.notion.com/mcp` from the **Sign-in** type on an install with a public URL set, approve in the tab, and watch the row flip from Sign-in required to connected with the dialog closing itself


ahmadrosid (commented):
Two small simplifications to reduce the state bookkeeping and reuse an existing runtime helper.

apps/web/src/components/soul-tools/mcp-tab/use-mcp-server-dialog-state.ts
```diff
@@ -271,16 +294,12 @@ export function useMcpServerDialogState({
   const [url, setUrl] = useState("");
   const [headers, setHeaders] = useState<McpHeaderRow[]>([emptyHeaderRow()]);
   const [command, setCommand] = useState("");
+  const [signIn, setSignIn] = useState(false);
```
ahmadrosid, line 297: Could we store the selected McpServerKind directly and derive transport and signIn from it? The three tabs currently become two state values that resets, imports, and selectKind must keep in sync, then mcpServerKind converts them back. One kind value would keep the same three choices with less bookkeeping.

apps/server/src/http/shared.ts
```diff
@@ -441,6 +441,52 @@ export function errorResponse(
...
+</body>
+</html>`,
+    {
+      headers: { "Content-Type": "text/html; charset=utf-8" },
+      status,
+    }
+  );
+}
+
+function escapeHtml(value: string): string {
```
ahmadrosid, line 481: Bun.escapeHTML already handles these five characters. We can use it at the template interpolations and drop this replacement chain; the rendered text and attribute escaping stay the same.

yusupsupriyadi: @ahmadrosid Done in 28093b77. The four interpolations call `Bun.escapeHTML` and the local helper is gone. Only the apostrophe changes entity spelling, `&#39;` to `&#x27;`, which renders the same character, so the text and the href attribute come out identical.

apps/web/src/components/soul-tools/mcp-tab/use-mcp-server-dialog-state.ts
yusupsupriyadi: @ahmadrosid Done in 77a32309. The dialog now holds one `kind`, and `transport` and `signIn` are derived from it at the top of the hook, so `buildMcpServerRequest` and the edit-mode import check still take the same inputs while the form receives `kind` directly instead of converting back. `selectKind` is one `setKind` plus clearing the field the other type owns, the reset sets a single value, and the JSON import no longer clears `signIn` by hand.

One behavior change falls out of that, and it is the better one: a paste whose JSON parses but fails validation used to flip the form off the Sign-in tab before the import was rejected. It now leaves the tab where it was.

`bun test apps/web/src` (479 pass), `bun run --filter @nakama/web build` (tsc included), `bun run knip` and `ultracite check` are clean on the rebased branch.

PR #963, opened 2026-09-11
fix(workflows): resolve assert templates once

isnaenihidayat:
## Summary

Workflow asserts compare template-shaped input strings after exactly one resolution.

**Before:** `WorkflowRunner` resolved `expected`, then `executeAssert` resolved the resulting value again; a literal such as `{{literal}}` became a missing reference.
**After:** `executeAssert` owns the single resolution and the run receipt records that resolved value.

Why safe:
1. Compare, tool, template, and summarize steps are unchanged.
2. The regression failed on current `main`, passes with the fix, and runs through the real runner lifecycle.

Only revisit if workflows intentionally gain recursive template expansion.

## Test plan
- [x] `bun test packages/plugins/workflows/src apps/server/src/services/official-workflows.test.ts` (19 pass)
- [x] Plugin build, `bun run check`, and `bun run knip`
- [ ] Run an assert where both runtime values equal the literal `{{literal}}` — workflow completes

Follow-up to #826

### DIFF
--- a/packages/plugins/workflows/actions/actions.js
+++ b/packages/plugins/workflows/actions/actions.js
@@ -418,17 +418,16 @@ class WorkflowRunner {
       return { input: { left, op: step.op, right }, output: result };
     }
     if (step.kind === "assert") {
-      const expected = resolveWorkflowValue(step.expected, bag);
       const result = executeAssert({
         bag,
-        expected,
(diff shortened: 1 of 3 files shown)


ahmadrosid:
Nice thanks 👍

PR #964, opened 2026-09-11
fix(server): remove session attachments on purge

isnaenihidayat:
## Summary

Purge a chat session → its stored attachment bytes and metadata are removed with it.

**Before:** session purge deleted messages and the session row but left attachment files and detached metadata behind.
**After:** `purgeSession` finds attachments for the session, removes their bytes, then deletes their metadata before deleting the session.

Why safe:
1. Organization ownership is checked before any attachment cleanup starts
2. Cleanup is limited to attachment rows linked to the exact session; missing files remain harmless on retry

Only follow up if non-purge chat clearing should also erase attachments.

## Test plan
- [x] `bun test packages/db/src apps/server/src/services/session-persistence.test.ts` (120 pass)
- [x] `bun run check && bun run knip && bun run build`
- [ ] Upload one attachment, purge its session, and confirm the attachment path no longer exists

Progresses #368

### DIFF
--- a/apps/server/src/services/agent-service.ts
+++ b/apps/server/src/services/agent-service.ts
@@ -124,6 +124,7 @@ import {
   DEFAULT_THINKING_ENABLED,
   defaultOllamaBaseUrl,
   deleteArtifactFile,
+  deleteAttachmentBytes,
   extractImageParts,
   findProviderInstance,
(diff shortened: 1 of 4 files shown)


ahmadrosid (commented):
LGTM

Issue #430, opened 2026-08-22
Enterprise compliance: email infrastructure (password reset, address verification, invite delivery)


vincentri:
## Problem

There is no outbound system mailer anywhere in the server:

- **No password-reset flow at all.** Auth routes are setup / login / me / change-password / accept-invite / local-token-rotate only (`apps/server/src/http/routes/auth.ts`). A forgotten password requires someone with server access to intervene manually.
- **Invites return the raw token in the admin API response** for manual copy-paste (`createInvite` → `{ invite, token }`, `org-service.ts:655-660`).
- **No email ownership verification.** Combined with `PATCH /v1/auth/me` accepting an email change without current-password re-auth (`org-service.ts:313-343`), a hijacked session can silently take over the identity.

Note: the agent-facing email tool keeps its own SMTP/IMAP config (`packages/core/src/email-config.ts`) — that is separate; there is no transactional mailer for auth/org events.

## Why it matters

Company deployments expect self-service credential recovery and clean delivery of invites. Hand-copied tokens end up in chats, clipboards, and logs; lockout recovery via "SSH in and fix the DB" is both a support burden and a compliance problem. SSO (#366) reduces reliance on local passwords but does not remove the need for account recovery.

## Proposed scope

- Platform-admin–configured system SMTP settings, stored under the same private-permission conventions as other secrets.
- Transactional templates: invite, password reset, and (optionally) email-change verification.
- Reset flow mirrors invites: high-entropy token stored hashed, single-use, short expiry.
- Preserve the manual-token fallback when SMTP is not configured (self-hosters without mail access).

## Acceptance criteria

- [ ] Forgot-password endpoint issues a hashed, single-use, expiring token; email is delivered when SMTP is configured, manual-token fallback otherwise
- [ ] Invites can be delivered by email instead of copy-paste
- [ ] Optional verification step for email changes
- [ ] No secret material logged; tests cover token single-use and expiry

isnaenihidayat:
Taking the invite-delivery slice: send organization invitations through the configured SMTP mailbox, add a working browser acceptance page, and return the raw token only as a manual fallback when delivery is unavailable. I’ll keep password reset and address verification as separate follow-ups.

isnaenihidayat:
[...]

PR #969, opened 2026-09-11
fix(auth): require reauthentication for email changes

isnaenihidayat:
## Summary

Change an account email → the current password must verify before the identity changes.

**Before:** any valid browser session could replace its account email without reauthentication.
**After:** updateOwnProfile verifies the current password for real email changes, and the profile dialog sends it only when needed.

Why safe:
1. Missing or wrong passwords fail before duplicate lookup or database mutation
2. Name and phone updates still work without a password; password changes keep their existing flow

Only follow up here if passwordless sign-in lands; reset and invite delivery remain separate #430 work.

## Test plan
- [x] bun test apps/server/src/services/org-service.test.ts apps/server/src/http/org-invites.test.ts (38 pass)
- [x] Targeted Ultracite check, bun run knip, and bun run build
- [ ] Change only the account email in Profile and confirm the current password is required

Progresses #430

### DIFF
--- a/apps/server/src/http/routes/auth.ts
+++ b/apps/server/src/http/routes/auth.ts
@@ -64,6 +64,7 @@ export function registerAuthRoutes(app: HonoApp, options: ServerOptions): void {
     .openapi("AuthUserResponse");
   const updateAuthProfileSchema = z
     .object({
+      currentPassword: z.string().optional(),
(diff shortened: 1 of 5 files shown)


ahmadrosid (commented):
LGTM

Issue #364, opened 2026-08-21
Enterprise compliance: secrets-at-rest strategy (external secret managers, encrypted config)


vincentri:
> **Enterprise compliance gap** — identified in the nakama security & compliance audit (2026-08-21). Required for SOC 2 CC6.7 / ISO 27001 A.8.24 and general breach blast-radius containment.

# Summary

Every secret — provider API keys, channel bot tokens, the local auth token — is written to disk in plaintext. The one encryption scheme present derives its key from another plaintext-stored secret. There is no KMS/Vault integration.

# Why this blocks enterprise compliance

- Encryption of secrets at rest is a baseline audit finding; "file permissions" is not an accepted compensating control for credentials.
- Enterprises standardize on Vault/Cloud KMS/Secrets Manager and expect apps to consume secrets from there.
- Breach blast radius: any filesystem read = total credential compromise including all LLM accounts.

# Current state (verified)

- Provider `api_key` values written verbatim into `config.ini` (`user-config.ts:641` builds `api_key: provider.apiKey`; `:707` emits raw INI lines). Same pattern for Telegram/Discord/WhatsApp tokens and Composio key. Mitigation is chmod 0600/0700 only (`fs.ts:13-14`).
- Local auth token stored plaintext at `{configDir}/local-auth-token`; config.ini holds a bare SHA-256 hash (not a KDF) (`local-auth.ts:30-32,65-69`).
- Composio AES-256-GCM (`composio-secret.ts:8-24`) keys off SHA-256 of the deployment's own Composio API key — stored plaintext one directory over (`composio-service.ts:965-971`). Covers only `session_id_enc` columns.
- Grep `kms|vault|secretsmanager|keychain` repo-wide: zero integrations.
- SQLite opened unencrypted, no WAL (`sqlite.ts:406`); transcripts/attachments plaintext.

# Proposed approach

1. Secret injection: support `${ENV_VAR}` interpolation or `_FILE` convention in config.ini so operators can mount secrets from Vault/KMS without them touching disk in cleartext.
2. At-rest encryption mode: encrypt secret values in config.ini under a master key supplied via env/file (documented KMS integration examples).
3. Interim docs: require encrypted volumes for the data dir; upgrade local-token hash to a slow KDF.
4. Longer term: evaluate SQLCipher/SEE for the DB (ties into managed-DB issue).

# Acceptance criteria

- [ ] In master-key mode, no plaintext provider/channel tokens exist on disk (automated test greps the data root).
[...]

PR #971, opened 2026-09-11
feat(config): support mounted provider API keys

isnaenihidayat:
## Summary

Mount a provider API key as a file → Nakama reads it without persisting the secret in settings.

**Before:** Provider startup accepted direct API-key environment variables only.
**After:** Every supported provider key also accepts a `_FILE` companion path; direct values still win.

Why safe:
1. Bootstrap coverage confirms the mounted secret never reaches `config.ini`
2. Missing files fail explicitly, while lazy filesystem access keeps the web bundle buildable

Only follow up if channel or other non-provider secrets need the same convention.

## Test plan
- [x] `bun test packages/core/src/provider-resolution.test.ts apps/server/src/setup.test.ts` (38 pass)
- [x] Targeted Ultracite check, `bun run knip`, `bun run build`, and `bun run build:docs`
- [ ] Start with `NAKAMA_PROVIDER=openai` and `OPENAI_API_KEY_FILE=/run/secrets/openai_api_key`

Progresses #364

### DIFF
--- a/apps/server/src/setup.test.ts
+++ b/apps/server/src/setup.test.ts
@@ -1,5 +1,5 @@
 import { afterEach, describe, expect, test } from "bun:test";
-import { mkdtemp, rm } from "node:fs/promises";
+import { mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import {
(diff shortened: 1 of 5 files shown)


ahmadrosid (commented):
LGTM

PR #972, opened 2026-09-12
fix(auth): protect install-wide settings

### LINKED ISSUE #967: Define org isolation for shared providers, Discord, WhatsApp, and Composio configuration
Supersedes #307 after rechecking main at `097563e785ce90c410c2827452d9889579c30265`.

An org admin can still change provider credentials and endpoints used by other organizations. Discord, WhatsApp, and the Composio API key also remain shared across the install. The tenant configuration boundary still needs a decision.

### Current state

| Configuration | Scope |
| --- | --- |
| Provider instances, API keys, endpoints, default provider | Install-wide `userConfig` |
| Timezone and thinking defaults | Install-wide |
| Vision, transcription, image-generation model settings | Install-wide workspace settings |
| Discord bot configuration | Install-wide |
| WhatsApp configuration/session | Install-wide |
| Composio API key | Install-wide |
| Telegram configuration | **Org-scoped; excluded from the remaining work** |

Composio's shared API key should not be confused with its org-scoped toolkits and per-user connections.

### Evidence

- [Provider update route](https://github.com/ahmadrosid/nakama/blob/097563e785ce90c410c2827452d9889579c30265/apps/server/src/http/routes/models.ts#L1373) accepts an org admin or platform admin and calls the service without an org ID.
- [Provider persistence](https://github.com/ahmadrosid/nakama/blob/097563e785ce90c410c2827452d9889579c30265/apps/server/src/services/agent-service.ts#L2430) updates shared `userConfig`.
- [Telegram configuration](https://github.com/ahmadrosid/nakama/blob/097563e785ce90c410c2827452d9889579c30265/packages/core/src/telegram-config.ts#L53) now uses an org-specific directory.
- Discord and WhatsApp use install-level directories in `packages/core/src/discord-config.ts` and `packages/core/src/whatsapp-config.ts`; `setComposioSettings` also has no org scope.
- Ran `bun test apps/server/src/http/routes/settings-rbac.test.ts`: **47 passed, 0 failed**, including the test that an org admin can update provider settings. These tests verify permissions, not complete cross-org isolation.

The old issue's claim that all channel credentials live in `WORKSPACE_SETTINGS_ID` is outdated. Shared channel configuration is stored in configuration files.

### Decision needed

Choose the intended boundary before implementation:

1. Keep these settings shared, restrict their mutations to platform admins, and document what organizations share.
2. Make the remaining configuration org-scoped, including the necessary channel/session routing and credential migration.
3. Add org-level provider overrides with an explicit fallback to install defaults, while handling shared channel settings separately.

### Completion criteria

- Document which settings belong to the install and which belong to an org.
- Enforce that boundary in reads, writes, and runtime configuration resolution.
- An org admin cannot alter another org's effective configuration unless explicitly authorized as a platform admin.
- Add cross-org regression coverage and define migration/fallback behavior for existing installs.

This replaces the outdated scope of #307; it does not mean the remaining isolation gap is fixed.

isnaenihidayat:
## Summary

Keep install-wide credentials under platform-admin control → an org admin can no longer change another organization's effective providers or shared bridges.

**Before:** An org admin could update provider defaults, shared channel credentials, and other installation settings used by every organization.
**After:** Platform admins own installation writes; Telegram and organization-level Composio operations keep their existing tenant scope.

Why safe:
1. Authorization runs before configuration or worker side effects across every protected route
2. The UI hides installation controls from org admins, while regression coverage preserves Telegram and Composio organization workflows

Only revisit this boundary if these settings gain organization-scoped storage and migration.

## Test plan
- [x] `bun test apps/server/src/http/routes/error-tracking.test.ts apps/server/src/http/routes/settings-rbac.test.ts apps/server/src/http/routes/composio.test.ts` (77 pass)
- [x] Targeted Ultracite check, `bun run knip`, `bun run build`, and `bun run build:docs`
- [ ] Sign in as an org admin and confirm shared provider, Discord, WhatsApp, and Composio credential controls are unavailable

Closes #967

### DIFF
--- a/apps/server/src/http/routes/models.ts
+++ b/apps/server/src/http/routes/models.ts
@@ -1636,7 +1636,7 @@ export function registerModelRoutes(
   });
 
   app.put("/v1/settings/email", async (c) => {
-    requireOrgAdminFromContext(c);
+    requirePlatformAdminFromContext(c);
     const body = await readJson<UpdateEmailSettingsRequest>(c.req.raw);
 
     try {
@@ -1651,7 +1651,7 @@ export function registerModelRoutes(
   });
 
   app.post("/v1/settings/email/test", async (c) => {
-    const auth = requireOrgAdminFromContext(c);
+    const auth = requirePlatformAdminFromContext(c);
     const body = await readOptionalJson<SendEmailTestRequest>(c.req.raw, {});
 
     try {
@@ -1697,7 +1697,7 @@ export function registerModelRoutes(
   });
 
   app.post("/v1/settings/agent-browser/install", async (c) => {
-    requireOrgAdminFromContext(c);
+    requirePlatformAdminFromContext(c);
 
     return streamAgentBrowserInstall(
       async (send, signal) => {
--- a/apps/server/src/http/routes/models.ts
+++ b/apps/server/src/http/routes/models.ts
@@ -67,6 +67,7 @@ import {
   requireNotViewerFromContext,
   requireOrgAdminFromContext,
   requireOrgAdminOrPlatformAdminFromContext,
+  requirePlatformAdminFromContext,
 } from "../org-guards";
 import {
   errorResponse,
@@ -1355,7 +1356,7 @@ export function registerModelRoutes(
   });
 
   app.post("/v1/providers", async (c) => {
-    requireOrgAdminOrPlatformAdminFromContext(c);
+    requirePlatformAdminFromContext(c);
     const body = await readJson<CreateProviderRequest>(c.req.raw);
 
     try {
@@ -1371,7 +1372,7 @@ export function registerModelRoutes(
   });
 
   app.patch("/v1/providers/:providerId", async (c) => {
-    requireOrgAdminOrPlatformAdminFromContext(c);
+    requirePlatformAdminFromContext(c);
     const body = await readJson<UpdateProviderRequest>(c.req.raw);
 
     try {
@@ -1392,14 +1393,14 @@ export function registerModelRoutes(
   });
 
   app.delete("/v1/providers/:providerId", async (c) => {
-    requireOrgAdminOrPlatformAdminFromContext(c);
+    requirePlatformAdminFromContext(c);
     return json<DeleteProviderResponse>(
       await agent.deleteProvider(decodeURIComponent(c.req.param("providerId")))
     );
   });
 
   app.post("/v1/xai-oauth/device/start", async (c) => {
-    const auth = requireOrgAdminOrPlatformAdminFromContext(c);
+    const auth = requirePlatformAdminFromContext(c);
     const owner = JSON.stringify([auth.user.id, auth.activeOrgId]);
 
     try {
@@ -1415,7 +1416,7 @@ export function registerModelRoutes(
   });
 
   app.post("/v1/xai-oauth/device/complete", async (c) => {
-    const auth = requireOrgAdminOrPlatformAdminFromContext(c);
+    const auth = requirePlatformAdminFromContext(c);
     const owner = JSON.stringify([auth.user.id, auth.activeOrgId]);
     const body = await readJson<{ sessionId?: string }>(c.req.raw);
     const sessionId =
@@ -1448,7 +1449,7 @@ export function registerModelRoutes(
   });
 
   app.post("/v1/chatgpt-oauth/device/start", async (c) => {
-    requireOrgAdminOrPlatformAdminFromContext(c);
+    requirePlatformAdminFromContext(c);
 
     try {
       return json(await startChatgptOAuthDeviceSession());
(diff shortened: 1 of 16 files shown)


ahmadrosid (commented):
The guard changes are straightforward. One small simplification in the authorization tests.

apps/server/src/http/routes/settings-rbac.test.ts
```diff
@@ -196,6 +200,14 @@ describe("workspace-global settings writes require an org admin", () => {
       expect(response.status).toBe(403);
       expect(calls).toEqual([]);
     });
+
+    test(`${route.method} ${route.path} -> 403 for an org admin`, async () => {
```
ahmadrosid: These three tests only differ by role. You could loop over `["viewer", "member", "admin"] as const` inside the route loop and keep one test body, with the role in its name. That keeps every route/role case and both assertions while dropping the repeated setup.

isnaenihidayat: Updated. The three roles now share one nested loop while keeping every route/role case and both assertions. The focused suite still passes all 69 tests.

apps/server/src/http/routes/models.ts
```diff
@@ -1355,7 +1356,7 @@ export function registerModelRoutes(
   });
 
   app.post("/v1/providers", async (c) => {
-    requireOrgAdminOrPlatformAdminFromContext(c);
+    requirePlatformAdminFromContext(c);
```
fajarhide, line 1359: Four install-wide writes did not move. I logged in as an org admin (no platform-admin flag) against this branch and called them:

```
PUT  /v1/settings/email                  -> 400
POST /v1/settings/email/test             -> 400
POST /v1/settings/agent-browser/install  -> 200
PUT  /v1/settings/coding-harnesses       -> 400
```

The 400s are my throwaway bodies, not the guard: all four are past it. Email is the one I would not leave, `saveEmailConfig` writes `getUserConfigPath()` (`packages/core/src/email-config.ts:315`), the same install-wide file as the provider config this PR just moved, and invite and password-reset mail routes through it, so an org admin can still repoint the install's SMTP. `agent-browser/install` runs a host-level install on `requireOrgAdminFromContext`.

Telegram staying org-scoped is right and the regression test for it is the right call.

apps/web/src/pages/chat/use-chat-page.ts
```diff
@@ -322,9 +323,9 @@ export function useChatPage() {
 
   const readOnlySession = isReadOnlySessionChannel(sessionChannel);
   const showThinking = shouldShowThinkingEffort(activeModelSupportsThinking);
-  const thinkingEffortVisible = shouldShowThinkingEffort(
-    activeModelSupportsThinking
-  );
+  const thinkingEffortVisible =
```
fajarhide: This hides the thinking-effort selector in the composer from everyone who is not a platform admin, which is a visible feature loss for every member and org admin, not just a credential lockdown. The server guard is right because the value is install-wide, but was removing the control the intended trade, or should it stay visible and read-only until thinking gets per-profile storage? Gating the auto-enable effect the same way is good, that avoids a 403 loop.

fajarhide (commented):
Swapping the guard at 22 call sites in one file and gating the UI to match is the right shape, and `PUT /v1/settings/thinking` sitting on `getRequestAuth` was a real hole. The 22 by 3 role table is the kind of coverage this needs. Two things: four install-wide writes stayed on the org-admin side, and @ahmadrosid this picks option 1 of the three #967 asked you to choose between, so a multi-org install loses org-admin access to provider settings on upgrade. Worth your word before it merges.

apps/server/src/http/routes/models.ts
isnaenihidayat: Fixed in `50c5e4d0`. The email update/test, agent-browser install, and coding-harness update routes now require platform admin. I added all four to the shared denial matrix for viewer, member, and org-admin roles, and aligned the related controls so non-platform admins cannot trigger these writes.

apps/web/src/pages/chat/use-chat-page.ts
isnaenihidayat: Good catch. Fixed in `50c5e4d0`: the selector stays visible when the model supports thinking, but is read-only for non-platform admins. The change handler and auto-enable effect remain platform-admin gated, so users can inspect the current effort without causing a 403 loop.

fajarhide:
Re-ran both checks on `50c5e4d0`. As an org admin with no platform-admin flag:

```
PUT  /v1/settings/email                  -> 403
POST /v1/settings/email/test             -> 403
POST /v1/settings/agent-browser/install  -> 403
PUT  /v1/settings/coding-harnesses       -> 403
PUT  /v1/settings/telegram               -> 200
```

Telegram still passing is the right half of that. `thinkingEffortVisible` is back to the model check and `thinkingEffortDisabled` carries `!canManageInstallSettings`, so the selector stays where people expect it. Full settings-rbac file green here, 86 pass.

The open item is not yours: #967 asked @ahmadrosid to pick between three boundaries and this takes option 1, so a multi-org install loses org-admin access to provider settings on upgrade.

ahmadrosid:
Option 1 is the intended boundary. Shared installation settings should require platform admin. We can handle org-level provider overrides separately. Good to merge.

PR #973, opened 2026-09-12
feat(security): add append-only audit events

### LINKED ISSUE #369: Enterprise compliance: append-only audit log of admin & security events

isnaenihidayat:
## Summary

Inspect a security event → see who changed what, in which organization, and under which request ID.

**Before:** Login, role, provider, profile, organization, import/export, and session changes left no durable audit trail.
**After:** Covered mutations append immutable `audit_events` rows, and platform admins can query them through `GET /v1/platform/audit-events`.

Why safe:
1. SQLite triggers reject updates and deletes; no foreign keys means account or organization erasure cannot remove history
2. Event metadata contains only response outcome and status, never request bodies, credentials, or tokens
3. Coverage tests exercise every security-event class and the read API rejects organization admins

Only follow up if a new security-sensitive route is added; add its rule beside the coverage checklist and test table.

## Test plan
- [x] `bun test packages/db/src/audit-events.test.ts apps/server/src/http/audit-log.test.ts` (16 pass)
- [x] Targeted Ultracite check, `bun run knip`, and `bun run build`
- [ ] Make one successful and one failed login, then query `/v1/platform/audit-events?action=auth.login`

Closes #369

### DIFF
--- a/apps/server/src/http/audit-log.test.ts
+++ b/apps/server/src/http/audit-log.test.ts
@@ -1,0 +1,130 @@
+import { describe, expect, test } from "bun:test";
+import { setupTestConfigDir } from "../test-config-dir";
+import { describeAuditEvent } from "./audit-log";
+import { createMinimalHonoApp } from "./test-app-helpers";
+import {
+  type AppFetch,
+  loginPlatformAdminSession,
+  loginUserSession,
+  seedOrgAdmin,
+} from "./test-session-helpers";
+
+setupTestConfigDir("nakama-audit-log-test-");
+
+describe("audit event coverage", () => {
+  test.each([
+    ["POST", "/v1/auth/login", "auth.login"],
+    ["POST", "/v1/auth/logout", "auth.logout"],
+    ["POST", "/v1/auth/change-password", "auth.password_change"],
+    ["POST", "/v1/auth/local-token/rotate", "auth.token_rotate"],
+    ["POST", "/v1/orgs/org_1/invites", "invite.create"],
+    ["PATCH", "/v1/orgs/org_1/members/user_1", "member.role_update"],
+    ["PUT", "/v1/settings/email", "settings.update"],
+    ["PATCH", "/v1/providers/provider_1", "provider.update"],
+    ["DELETE", "/v1/profiles/profile_1", "profile.delete"],
+    ["GET", "/v1/profiles/profile_1/pack/export", "profile.export"],
+    ["POST", "/v1/platform/data/import/restore", "data.import"],
+    ["DELETE", "/v1/sessions/session_1", "session.revoke"],
+  ])("classifies %s %s", (method, path, action) => {
+    expect(describeAuditEvent(method, path)?.action).toBe(action);
+  });
+});
+
+describe("audit event API", () => {
+  test("records actor, organization, outcome, and request", async () => {
+    const { app, authService, databaseAdapter } = createMinimalHonoApp();
+    const fetchApp = app as unknown as AppFetch;
+    const session = await loginPlatformAdminSession(
+      fetchApp,
+      authService,
+      databaseAdapter
+    );
+    const platformUser = await databaseAdapter.getUserByEmail(
+      "platform@example.com"
+    );
+
+    const createResponse = await app.fetch(
+      new Request("http://localhost:4310/v1/platform/orgs", {
+        body: JSON.stringify({ name: "Audit Org", slug: "audit-org" }),
+        headers: session.headers({ "X-CSRF-Token": session.csrfToken }),
+        method: "POST",
+      })
+    );
+    expect(createResponse.status).toBe(201);
+    const created = (await createResponse.json()) as {
+      organization: { id: string };
+    };
+
+    const failedLogin = await app.fetch(
+      new Request("http://localhost:4310/v1/auth/login", {
+        body: JSON.stringify({
+          email: "platform@example.com",
+          password: "wrong-password",
+        }),
+        headers: { "Content-Type": "application/json" },
+        method: "POST",
+      })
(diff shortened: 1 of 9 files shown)


apps/server/src/http/audit-log.ts
```diff
@@ -0,0 +1,297 @@
...
+ * Security-event coverage checklist:
+ * - authentication: setup, login success/failure, logout, password change, token rotation
+ * - authorization: organization/member/invite lifecycle and role changes
+ * - configuration: providers and installation settings
+ * - sensitive resources: profile/org CRUD, export/import, session revocation
+ *
+ * Add a rule here whenever a new route joins one of those classes. Request
+ * bodies are intentionally never recorded, so credentials cannot enter metadata.
+ */
+export function describeAuditEvent(
```
fajarhide, line 38: I built the app in a test, read `app.routes`, and fed every mutating path under /v1/auth, /v1/orgs, /v1/platform, /v1/providers, /v1/settings, /v1/profiles and /v1/sessions through `describeAuditEvent`. 50 come back null. Most are fine (session messages, memory search), these are not:

```
POST /v1/auth/setup/import/restore         full-install restore, while /v1/platform/data/import/restore is audited
POST /v1/platform/plugins/releases         installs code
POST /v1/profiles/:profileId/tools         same for mcp-servers, skills, and their DELETEs
PUT  /v1/profiles/:profileId/soul/files/:fileKey   rewrites the agent instructions
POST /v1/auth/active-org
```

Also `invite.revoke` cannot fire: there is no DELETE invites route in the app today, only `POST /v1/orgs/:orgId/invites` and its platform twin.

Worth noting the repo already has the service-layer shape #369 asked for in `recordProfileChangeEvent` (`apps/server/src/services/profile-change-history.ts`), called from profile-service and agent-service. Middleware buys you one place to look; it costs this drift. Your call which one, but the gaps above need closing either way.

apps/server/src/http/audit-log.test.ts
```diff
@@ -0,0 +1,130 @@
...
+    ["POST", "/v1/auth/local-token/rotate", "auth.token_rotate"],
+    ["POST", "/v1/orgs/org_1/invites", "invite.create"],
+    ["PATCH", "/v1/orgs/org_1/members/user_1", "member.role_update"],
+    ["PUT", "/v1/settings/email", "settings.update"],
+    ["PATCH", "/v1/providers/provider_1", "provider.update"],
+    ["DELETE", "/v1/profiles/profile_1", "profile.delete"],
+    ["GET", "/v1/profiles/profile_1/pack/export", "profile.export"],
+    ["POST", "/v1/platform/data/import/restore", "data.import"],
+    ["DELETE", "/v1/sessions/session_1", "session.revoke"],
+  ])("classifies %s %s", (method, path, action) => {
```
fajarhide, line 28: This asserts the classifier against string literals the classifier itself owns, so it stays green if a route path moves underneath it, and it is why the gaps above are invisible. Walking `app.routes` and asserting every mutating security route classifies would catch both. I have that probe locally if you want it.

apps/server/src/http/audit-log.ts
```diff
@@ -0,0 +1,297 @@
...
+
+    let attemptedUserId: string | null = null;
+    if (descriptor.action === "auth.login") {
+      attemptedUserId = await resolveLoginUserId(c.req.raw, databaseAdapter);
+    }
+
+    await next();
+
+    const status = c.res.status;
+    if (status >= 400 && descriptor.action !== "auth.login") {
```
fajarhide, line 173: Non-login failures never land, so a denied role change or a 403 on an org route leaves no row, and that denial is usually what an auditor is looking for. Deliberate for the first pass?

fajarhide (commented):
The trigger pair plus no foreign keys is the right shape for a ledger, and the read route is platform-admin guarded. What I would fix before merge is coverage: I walked the real route table against the classifier and several routes in the classes #369 lists come back null.

PR #977, opened 2026-09-12
fix(whatsapp): fresh reports, no scratch spam, full retry sets

### LINKED ISSUE #974: [whatsapp] Freshness request serves yesterday's file without running the agent
## What happens

`tolong kirim laporan harian` with a stale registry artifact serves the old file and never runs the agent. The user gets yesterday's report with no freshness signal.

## Repro

1. Day 1: `buatkan laporan` fills the registry with `laporan-01.csv`.
2. Day 2: `tolong kirim laporan harian`.
3. The day-1 file arrives; the agent turn never runs (`sendStream` 0 calls).

Root: `createsArtifact` in `apps/platform/whatsapp/src/chat-handler.ts` is start-anchored and has no `kirim` verb, while `isAttachIntent` matches `kirim` + `laporan` — so the freshness request hits the pre-turn path and `return`s after serving (`if (attached) return`).

Proven with a failing test on `origin/main` (expected `sendStream` 1, got 0).

## Expected

Time-bound report requests (`harian` / `daily` / `hari ini` / `today`) run the agent first so the report is built fresh.

### LINKED ISSUE #975: [whatsapp] Post-turn delivery sends every write including scratch files
## What happens

After an agent turn, every successful `write_file` under `artifacts/` is pushed into the group as a document — including scratch and intermediate files. A turn writing `laporan-final.csv` plus `_scratch-debug.json` delivers 2 documents for a "send me the csv" request.

## Repro

1. Ask: `save it and send me the csv`.
2. Agent writes `laporan-final.csv` and `_scratch-debug.json` in the same turn.
3. Both files arrive in the group (expected 1, got 2).

Root: the post-turn loop in `handleChatMessage` (`apps/platform/whatsapp/src/chat-handler.ts`) sends all of `extractPairedTurnArtifacts` with no intent check and no already-sent dedup. #911 widened this from sidecar-paired writes to any successful write with `bytesWritten`.

Proven with a failing test on `origin/main`.

## Expected

Dotfiles, underscore-prefixed, and temp-looking artifacts (`*.tmp`, `*.bak`, `*~`) are skipped for document delivery. Residual: agent-named junk without such a marker still goes out.

### LINKED ISSUE #976: [whatsapp] Retry delivers only the newest artifact; older set members unreachable
## What happens

A retry over a multi-file set delivers only the newest artifact. Registry `[part1, part2]` + `send it to this group` sends part2 only — part1 is then unreachable, because a text-only retry extracts zero new turn artifacts.

## Repro

1. Registry holds `laporan-part1.csv` and `laporan-part2.csv`.
2. Ask: `send it to this group`.
3. 1 document arrives (expected 2).

Root: `maybeSendRequestedWhatsAppArtifactAttachment` (`apps/platform/whatsapp/src/channel-artifact-flow.ts`) sends only `getMostRecentDeliverableArtifact`, while the post-turn path that could complete the set only fires on new writes.

Proven with a failing test on `origin/main`.

## Expected

The natural-language retry delivers the full deliverable set (`/attach` stays most-recent-only). Residual: a registry full of stale files could send several docs at once — a cap may follow.

itw-code:
## Summary

WhatsApp report requests get fresh files, no scratch spam, and complete retry sets.

**Before:** `kirim laporan harian` served yesterday's file without running the agent; every write including scratch blasted into the group; a retry delivered only the newest artifact.
**After:** freshness markers force an agent turn first (`isFreshReportRequest`); scratch-looking writes skip post-turn delivery; a natural-language retry sends the full deliverable set.

Why safe:
1. Plain `kirim file X` and `/attach` still short-circuit exactly as before — marker words inside a filename are stripped before the freshness test, so `kirim file daily-report.csv` stays an attach
2. Scratch filter is filename-pattern only; the registry still records everything
3. All pre-existing delivery tests pass with unchanged assertions

Retry fan-out is bounded: `pushDeliverableArtifact` caps the registry at 5 entries (`channel-artifact-delivery.ts:178`), so the worst case is 5 documents, not an unbounded blast.

Scratch filtering is WhatsApp-only on purpose. Telegram and Discord share `deliverTurnArtifactShares`, and moving the filter into `extractPairedTurnArtifacts` would drop scratch from the registry entirely, killing the `/attach` retrieval of an earlier turn's file that issue #975 asks to keep.

Fixes #974, Fixes #975, Fixes #976.

## Test plan
- [x] `bun test apps/platform/whatsapp/src packages/core/src/channel-artifact-delivery.test.ts packages/core/src/channel-artifacts.test.ts` (135 pass, 4 skip)
- [x] `bun test apps/platform/telegram/src/chat-handler.test.ts apps/platform/discord/src/chat-handler.test.ts` (117 pass)
- [x] `bun x ultracite check` on the 7 changed files (clean; 2 failures are pre-existing on untouched files)
- [x] `bun run knip` (clean)
- [x] New cases fail without the change: with `chat-handler.ts` and `channel-artifact-flow.ts` reverted to `main`, 49 pass / 3 fail; with the filename-token strip reverted, the 2 new filename cases fail
- [ ] Ask the validator to request a daily CSV in the WhatsApp group; confirm a fresh document arrives

### DIFF
--- a/apps/platform/whatsapp/src/channel-artifact-flow.ts
+++ b/apps/platform/whatsapp/src/channel-artifact-flow.ts
@@ -3,6 +3,7 @@ import {
   formatMissingAttachArtifactMessage,
   getMostRecentDeliverableArtifact,
   isAttachIntent,
+  isScratchArtifactPath,
 } from "@nakama/core";
 import type { ChannelSessionStore } from "@nakama/core/channel-session-store";
 import type { WASocket } from "@whiskeysockets/baileys";
@@ -16,6 +17,10 @@ import {
  * When the user asks to attach/send a file and a registry artifact exists,
  * sends the WhatsApp document. Returns true when attach was attempted so the
  * chat handler can skip the agent turn (avoids invented "can't attach" replies).
+ *
+ * Natural-language retry sends the whole deliverable set (minus scratch):
+ * the post-turn path only fires on new writes, so older set members would
+ * otherwise be unreachable. `/attach` stays most-recent-only.
  */
 export async function maybeSendRequestedWhatsAppArtifactAttachment(input: {
   client: NakamaClient;
@@ -32,20 +37,22 @@ export async function maybeSendRequestedWhatsAppArtifactAttachment(input: {
     return false;
   }
 
-  const artifact = getMostRecentDeliverableArtifact(
-    input.sessionStore.getDeliverableArtifacts(input.conversationKey)
-  );
-  if (!artifact) {
+  const artifacts = input.sessionStore
+    .getDeliverableArtifacts(input.conversationKey)
+    .filter((artifact) => !isScratchArtifactPath(artifact.path));
+  if (artifacts.length === 0) {
     return false;
   }
 
-  await sendArtifactDocumentForPath({
-    ...input,
-    filename: artifact.filename,
-    mimeType: artifact.mimeType,
-    path: artifact.path,
-    sizeBytes: artifact.sizeBytes,
-  });
+  for (const artifact of artifacts) {
+    await sendArtifactDocumentForPath({
+      ...input,
+      filename: artifact.filename,
+      mimeType: artifact.mimeType,
+      path: artifact.path,
+      sizeBytes: artifact.sizeBytes,
+    });
+  }
   return true;
 }
 
--- a/apps/platform/whatsapp/src/chat-handler.ts
+++ b/apps/platform/whatsapp/src/chat-handler.ts
@@ -3,6 +3,7 @@ import {
   extractPairedTurnArtifacts,
   isAttachOnlyCommand,
   isFreshReportRequest,
+  isScratchArtifactPath,
   pushDeliverableArtifact,
 } from "@nakama/core";
 import { formatClientError } from "@nakama/core/api-error";
@@ -550,11 +551,16 @@ export function createChatHandler(deps: ChatHandlerDeps) {
       });
       await sessionStore.save();
 
+      // Scratch-looking writes stay retrievable via /attach but never blast
+      // into the group unasked.
+      const deliverable = artifacts.filter(
+        (artifact) => !isScratchArtifactPath(artifact.path)
+      );
       const postTurnSocket = getSocket();
       if (!postTurnSocket) {
         return;
       }
-      for (const artifact of artifacts) {
+      for (const artifact of deliverable) {
         await sendArtifactDocumentForPath({
           ...artifact,
           client,
--- a/apps/platform/whatsapp/src/chat-handler.ts
+++ b/apps/platform/whatsapp/src/chat-handler.ts
@@ -2,6 +2,7 @@ import type { NakamaClient, RemoteChatSession } from "@nakama/client";
 import {
   extractPairedTurnArtifacts,
   isAttachOnlyCommand,
+  isFreshReportRequest,
   pushDeliverableArtifact,
 } from "@nakama/core";
 import { formatClientError } from "@nakama/core/api-error";
@@ -435,7 +436,11 @@ export function createChatHandler(deps: ChatHandlerDeps) {
     const socket = getSocket();
 
     // ponytail: imperative phrases only; use an explicit send tool for richer requests.
+    // Freshness markers (harian/today/…) mean "build it now": never serve the
+    // registry without an agent turn, or yesterday's file goes out as today's.
+    const wantsFreshReport = isFreshReportRequest(attachUserText);
     const createsArtifact =
+      wantsFreshReport ||
       /^\s*(?:(?:please|tolong)\s+)?(?:collect|create|generate|save|buat(?:kan)?|rekap(?:kan)?)\b/i.test(
         attachUserText
       );
--- a/packages/core/src/channel-artifact-delivery.ts
+++ b/packages/core/src/channel-artifact-delivery.ts
@@ -45,6 +45,20 @@ export interface ListedArtifactCandidate {
   updatedAt: string;
 }
 
+const FRESHNESS_MARKERS =
+  "harian|hari\\s+ini|terbaru|today|daily|latest|yang\\s+baru|update|diperbarui";
+
+export function isFreshReportRequest(text: string): boolean {
+  const normalized = text.trim();
+  if (!normalized) {
+    return false;
+  }
+
+  return new RegExp(String.raw`\b(?:${FRESHNESS_MARKERS})\b`, "i").test(
+    normalized
+  );
+}
+
 export function isAttachIntent(text: string): boolean {
   const normalized = text.trim();
   if (!normalized) {
--- a/apps/platform/whatsapp/src/chat-handler.test.ts
+++ b/apps/platform/whatsapp/src/chat-handler.test.ts
@@ -1826,6 +1826,34 @@ describe("createChatHandler artifact delivery", () => {
     );
   });
 
+  test("delivers the full set on a natural-language retry, not just newest", async () => {
+    await withArtifactChat(
+      {
+        deliverableArtifacts: [
+          {
+            ...SAMPLE_ARTIFACT,
+            filename: "laporan-part1.csv",
+            path: "laporan-part1.csv",
+          },
+          {
+            ...SAMPLE_ARTIFACT,
+            filename: "laporan-part2.csv",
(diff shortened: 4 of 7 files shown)


packages/core/src/channel-artifact-delivery.ts
```diff
@@ -45,6 +45,20 @@ export interface ListedArtifactCandidate {
...
+const FRESHNESS_MARKERS =
+  "harian|hari\\s+ini|terbaru|today|daily|latest|yang\\s+baru|update|diperbarui";
+
+export function isFreshReportRequest(text: string): boolean {
+  const normalized = text.trim();
+  if (!normalized) {
+    return false;
+  }
+
+  return new RegExp(String.raw`\b(?:${FRESHNESS_MARKERS})\b`, "i").test(
```
fajarhide, line 61: `\b` matches inside a filename too, so `kirim file daily-report.csv` comes out `attach=true fresh=true` and skips the attach path for an agent turn. Same for `send me update-harga.csv` and `kirim file laporan-terbaru.csv`, which is the naming a daily report ends up with. Stripping filename-looking tokens first keeps all four of your positive cases green and turns those three back to false:

```ts
const normalized = text.trim().replace(/\S+\.[a-z0-9]{2,5}\b/gi, " ");
```

I would fix this one before merge.

apps/platform/whatsapp/src/chat-handler.ts
```diff
@@ -545,11 +551,16 @@ export function createChatHandler(deps: ChatHandlerDeps) {
       });
       await sessionStore.save();
 
+      // Scratch-looking writes stay retrievable via /attach but never blast
+      // into the group unasked.
+      const deliverable = artifacts.filter(
```
fajarhide, line 556: The filter sits on the two WhatsApp send sites, so telegram and discord still run scratch through `deliverTurnArtifactShares`. I ran it with `laporan-final.csv` plus `_scratch-debug.json` and got both in `delivered` and both in the footer, so the scratch file gets a public share link in the chat. Filtering inside `extractPairedTurnArtifacts` would cover all three channels but also drop scratch from the registry, killing the `/attach` retrieval you kept on purpose. Deliberate scope for this PR?

apps/platform/whatsapp/src/channel-artifact-flow.ts
```diff
@@ -32,20 +37,22 @@ export async function maybeSendRequestedWhatsAppArtifactAttachment(input: {
...
   }
 
-  await sendArtifactDocumentForPath({
-    ...input,
-    filename: artifact.filename,
-    mimeType: artifact.mimeType,
-    path: artifact.path,
-    sizeBytes: artifact.sizeBytes,
-  });
+  for (const artifact of artifacts) {
```
fajarhide, line 47: On the residual in your description: `pushDeliverableArtifact` already caps the registry at 5 (`channel-artifact-delivery.ts:178`), so the worst case is 5 documents, not an unbounded blast. Worth saying in the body so nobody adds a second cap on top.

fajarhide (commented):
Three reports in one pass reads right, and the tests are real: with `chat-handler.ts` and `channel-artifact-flow.ts` reverted to main your three new cases fail (49 pass, 3 fail) and pass again with them back. One thing to fix, the freshness markers also match inside filenames.

fajarhide:
Filed the telegram and discord half as #981 so it does not ride on this PR. Same probe output, measured on c4572e6f, with the tradeoff between filtering in `deliverTurnArtifactShares` and in `extractPairedTurnArtifacts` written out there. Nothing for you to do here unless you want it in scope.

packages/core/src/channel-artifact-delivery.ts
itw-code: Fixed in 962603cc. Filename-looking tokens are stripped before the marker test:

```ts
const FILENAME_TOKEN = /\S+\.[a-z0-9]{2,5}\b/gi;
const normalized = text.trim().replace(FILENAME_TOKEN, " ").trim();
```

Your three cases are now pinned in `packages/core/src/channel-artifact-delivery.test.ts` (`daily-report.csv`, `update-harga.csv`, `laporan-terbaru.csv`), and the four positive cases still pass. Without the strip the new test fails; with it the three files are green (85 pass).

apps/platform/whatsapp/src/chat-handler.ts
itw-code: Deliberate scope. #975 is `[whatsapp]`-scoped, and the filter can't move into `extractPairedTurnArtifacts`: the registry push loop in `handleChatMessage` consumes that same array, so filtering there drops scratch from the registry and kills the `/attach` retrieval of an earlier turn's file that #975 asks to keep. Telegram and Discord go through `deliverTurnArtifactShares` (`packages/core/src/channel-artifact-delivery.ts`), untouched here.

Parity for those two is doable but needs the registry to keep recording scratch while delivery skips it — happy to do it as a follow-up rather than widen this PR.

apps/platform/whatsapp/src/channel-artifact-flow.ts
itw-code: Right — the body now states the cap instead of inviting a second one. `pushDeliverableArtifact(registry, artifact, maxEntries = 5)` bounds the registry at 5, and the retry path iterates exactly that registry, so the worst case is 5 documents.

ahmadrosid:
Thanks for the fixes! Can you try requesting a daily report in a WhatsApp group and confirm it sends a fresh file? That’s the last thing before we merge. Let’s keep Telegram/Discord in #981.

itw-code:
Thanks! Verified the group fresh-report routing with a harness proof — ready for your live-device confirm.

## Summary

Group `kirim laporan harian` now forces a fresh agent turn instead of re-sending yesterday's registry file.

**Before:** request short-circuited to the registry and re-sent yesterday's file with no agent turn.
**After:** `harian` marker forces `session.sendStream` via `isFreshReportRequest` (`packages/core/src/channel-artifact-delivery.ts`).

Why safe:
1. Filename tokens stripped before the freshness test, so `kirim file daily-report.csv` stays an attach
2. Scratch filtered by filename pattern only; registry still records everything so `/attach` retrieval is intact
3. Retry fan-out capped at 5 (`pushDeliverableArtifact`), worst case is 5 documents

Only re-check if a freshness word inside a filename with no marker intent should attach — `jangan`/`tidak usah` guard + `FILENAME_TOKEN` strip cover the reported phrasings.

Keeping Telegram/Discord to #981 as noted.

## Test plan
- [x] `bun test apps/platform/whatsapp/src/chat-handler.test.ts` (53 pass)
- [x] `bun test packages/core/src/channel-artifact-delivery.test.ts` (16 pass)
- [x] `bun test packages/core/src/channel-artifacts.test.ts` (16 pass)
- [x] Harness: `@bot kirim laporan harian` → `sendStream=1`, registry-doc=0; `daily-report.csv` / `update-harga.csv` / `laporan-terbaru.csv` → `sendStream=0`, attach attempted
- [x] Harness: retry sends full set minus scratch (2 docs, 0 scratch); 7 pushes → registry length 5; `/attach` sends exactly 1 (most recent)
- [ ] Live device: `@bot kirim laporan harian` in a paired test group — typing first, today's file, no `_scratch*` doc (checklist in `docs/qa/whatsapp-fresh-report-live-checklist.md`)

packages/core/src/channel-artifact-delivery.ts
fajarhide: Checked on e27eeba4: the three filename cases come back `fresh=false` now, and the four positives are still `true`. With `channel-artifact-delivery.ts` put back to 962603cc^, your new case goes red. nice

apps/platform/whatsapp/src/chat-handler.ts
fajarhide: Makes sense, the registry loop reads the same array, so filtering there would break `/attach`. Telegram and Discord stay in #981.

fajarhide:
Round two looks right to me. On e27eeba4 the three touched suites pass (85 pass, 0 fail), and tsc shows nothing on the files this PR changes.

The only box still open is the live group check ahmadrosid asked for, and I can't run it either, since I don't have a paired WhatsApp group. Your comment points to `docs/qa/whatsapp-fresh-report-live-checklist.md`, but that file isn't on the branch. If you want someone else to run the check, paste the checklist here.

PR #986, opened 2026-09-12
feat(providers): honour a custom model's own context window

### LINKED ISSUE #980: Custom provider models always compact at 128k, because CustomModelEntry has no context window
## Problem

Every model from a custom provider gets a hardcoded 128k context window, no matter what the model actually supports.

`resolveCompactionConfig` looks the model up in the built-in catalog only:

```ts
// apps/server/src/services/agent-service.ts:4078
const model = getModelById(resolved.model);

return {
  contextWindow: model?.contextWindow ?? 128_000,
  maxOutputTokens: model?.maxOutputTokens ?? 8192,
};
```

and `getModelById` only searches `AVAILABLE_MODELS`:

```ts
// apps/server/src/providers/models.ts:897
export function getModelById(modelId: string): ProviderModelOption | undefined {
  return AVAILABLE_MODELS.find((model) => model.id === modelId);
}
```

A custom model id is never in that catalog, so the lookup always misses and the fallback always wins.

There is also nowhere to put the real value. `CustomModelEntry` carries pricing and capability flags but no context size:

```ts
// packages/core/src/contract.ts:1652
export interface CustomModelEntry {
  default?: boolean;
  id: string;
  inputPerMillionUsd?: number;
  name?: string;
  outputPerMillionUsd?: number;
  supportsThinking?: boolean;
  supportsVision?: boolean;
}
```

That is why the provider setup UI offers pricing and vision toggles but no context field.

## Impact

`resolveCompactionConfig` feeds both history compaction and the context chip in chat, so one wrong number causes two problems:

- **Large models lose history early.** A 1M-context model compacts at roughly 128k, discarding turns that still fit.
- **Small models overflow.** A 32k model is treated as 128k, so nothing compacts and the provider rejects the request.
- **The chat chip reports a number that was never measured.** `ChatContextUsage.contextWindow` and `usableContextTokens` both come from this config.

## Near miss in the current code

Model discovery already reads the context size and then drops it. In `apps/web/src/hooks/use-provider-setup-form.ts:335` the browsed row carries `contextWindow: row.context`, but it only lands in the `extraModels` UI state used to populate the dropdown. It is never written into the persisted `customModels`, because the type has no field for it.

## Proposed fix

1. Add `contextWindow?: number` and `maxOutputTokens?: number` to `CustomModelEntry`, validated in `validateCustomModels` the way the pricing rates already are (positive finite numbers, both optional).
2. In `resolveCompactionConfig`, read the entry before falling back:

```ts
const custom = findCustomModel(resolved.instance.customModels, resolved.model);

return {
  contextWindow: custom?.contextWindow ?? model?.contextWindow ?? 128_000,
  maxOutputTokens: custom?.maxOutputTokens ?? model?.maxOutputTokens ?? 8192,
};
```

`resolveProfileProviderSelection` already returns the instance, so `customModels` is in scope.

3. Persist the discovered `row.context` into the saved custom model, and expose a context column in `ModelListEditor` for providers with no `/models` endpoint to discover from.

The fallback stays at 128k for entries that leave the field blank, so existing configs keep behaving exactly as they do today.

yusupsupriyadi:
Closes #980

## What was wrong

`resolveCompactionConfig` looked the model up in the built-in catalog only, and a custom model id is never in it:

```ts
const model = getModelById(resolved.model);

return {
  contextWindow: model?.contextWindow ?? 128_000,
  maxOutputTokens: model?.maxOutputTokens ?? 8192,
};
```

So every custom provider, every OpenRouter slug and every self-hosted build resolved to a hardcoded 128k. That number drives history compaction **and** the context chip in chat, so a 1M model discarded turns that still fit, and a 32k model never compacted and was rejected by the provider instead.

There was also nowhere to record the real size: `CustomModelEntry` had pricing and capability flags but no context field, which is why the setup UI offered no way to set one.

## What this does

- `CustomModelEntry` gains `contextWindow` and `maxOutputTokens`, validated in `validateCustomModels` as positive whole token counts, the same shape as the existing pricing rates.
- New `resolveModelLimits(modelId, customModels)` in `apps/server/src/providers/models.ts` prefers the instance entry, then the catalog, then the fallback. `resolveCompactionConfig` is now a single call to it.
- `ModelListEditor` gets a **Context** column, so endpoints with no `/models` to discover from can be told the real window. Blank means auto, and the helper line under the table says what auto resolves to.
- The models.dev browse row already read `limit.context` and then dropped it. It now lands in the saved entry for both the OpenAI-compatible and OpenRouter paths.
- `appendOpenRouterModelRow` was rebuilding each row key by key and silently losing anything it did not name, which is why the browsed context never survived. It spreads now, which also removes about 60 lines.

Nothing changes for an entry that leaves the sizes blank: catalog value first, then the same 128k fallback as before.

## Test plan

- `bun test packages/core/src/compatible-provider-config.test.ts apps/server/src/providers/models.test.ts apps/web/src/lib/models.test.ts apps/web/src/components/model-list-editor.shared.test.ts` — 108 pass, 0 fail. New coverage: size validation and rejection of non-integer or non-positive values, `resolveModelLimits` precedence (instance over catalog, catalog over fallback, fallback when nobody declares one), and `appendOpenRouterModelRow` keeping the browsed context while moving the default.
- `bun run --filter @nakama/web build` (runs `tsc --noEmit`) — exit 0.
- `bun run check` — clean for every file in this diff. Two pre-existing `useSortedKeys` findings remain in `packages/core/src/document-content.ts` and `packages/core/src/provider-label.ts`, both untouched here.
- `bun test apps/server/src apps/web/src` — 1794 pass / 54 fail on this branch against 1784 pass / 54 fail on the same commit without these changes. The 54 are pre-existing failures in the plugin suites and the cassette replays, identical on both sides; the difference is the 10 tests added here.

## Not covered

Providers whose browse path does not persist `customModels` at all (the catalog providers and `opencode_go`) still fall back for an unrecognised model id. The Context column in **Settings → LLM providers** covers those by hand; wiring their save path is a larger change and did not belong here.

### DIFF
--- a/apps/server/src/providers/models.ts
+++ b/apps/server/src/providers/models.ts
@@ -803,10 +803,6 @@ const BASE_MODELS: ProviderModelOption[] = withVisionDefaults([
   },
 ]);
 
-/** Used when neither the instance entry nor the catalog knows the model. */
-const FALLBACK_CONTEXT_WINDOW = 128_000;
-const FALLBACK_MAX_OUTPUT_TOKENS = 8192;
-
 export const AVAILABLE_MODELS: ProviderModelOption[] = [
   ...BASE_MODELS,
   ...deriveChatgptModels(BASE_MODELS),
@@ -916,14 +912,9 @@ export function resolveModelLimits(
   const custom = findCustomModel(customModels, modelId);
 
   return {
-    contextWindow:
-      custom?.contextWindow ??
-      catalog?.contextWindow ??
-      FALLBACK_CONTEXT_WINDOW,
+    contextWindow: custom?.contextWindow ?? catalog?.contextWindow ?? 128_000,
     maxOutputTokens:
-      custom?.maxOutputTokens ??
-      catalog?.maxOutputTokens ??
-      FALLBACK_MAX_OUTPUT_TOKENS,
+      custom?.maxOutputTokens ?? catalog?.maxOutputTokens ?? 8192,
   };
 }
 
--- a/apps/server/src/providers/models.ts
+++ b/apps/server/src/providers/models.ts
@@ -803,6 +803,10 @@ const BASE_MODELS: ProviderModelOption[] = withVisionDefaults([
   },
 ]);
 
+/** Used when neither the instance entry nor the catalog knows the model. */
+const FALLBACK_CONTEXT_WINDOW = 128_000;
+const FALLBACK_MAX_OUTPUT_TOKENS = 8192;
+
 export const AVAILABLE_MODELS: ProviderModelOption[] = [
   ...BASE_MODELS,
   ...deriveChatgptModels(BASE_MODELS),
@@ -896,6 +900,31 @@ export function validateOpenCodeGoCustomModels(
 
 export function getModelById(modelId: string): ProviderModelOption | undefined {
   return AVAILABLE_MODELS.find((model) => model.id === modelId);
+}
+
+/** Compaction sizing for one model.
+ *
+ * A custom model id is never in the built-in catalog, so without the instance
+ * entry every custom provider would silently compact at the fallback window.
+ * Entries that leave the sizes blank keep the catalog value, then the fallback.
+ */
+export function resolveModelLimits(
+  modelId: string,
+  customModels?: CustomModelEntry[]
+): { contextWindow: number; maxOutputTokens: number } {
+  const catalog = getModelById(modelId);
+  const custom = findCustomModel(customModels, modelId);
+
+  return {
+    contextWindow:
+      custom?.contextWindow ??
+      catalog?.contextWindow ??
+      FALLBACK_CONTEXT_WINDOW,
+    maxOutputTokens:
+      custom?.maxOutputTokens ??
+      catalog?.maxOutputTokens ??
+      FALLBACK_MAX_OUTPUT_TOKENS,
+  };
 }
 
 export function getModelsForProvider(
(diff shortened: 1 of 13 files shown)


apps/server/src/providers/models.ts
```diff
@@ -803,6 +803,10 @@ const BASE_MODELS: ProviderModelOption[] = withVisionDefaults([
   },
 ]);
 
+/** Used when neither the instance entry nor the catalog knows the model. */
+const FALLBACK_CONTEXT_WINDOW = 128_000;
+const FALLBACK_MAX_OUTPUT_TOKENS = 8192;
```
ahmadrosid: These constants are each used once. You can inline `128_000` and `8192` in `resolveModelLimits` and drop the declarations, keeping each fallback next to the custom → catalog lookup.

ahmadrosid (commented):
One small simplification; the rest looks lean.

apps/server/src/providers/models.ts
yusupsupriyadi: @ahmadrosid Done in d9bad828. Both constants had exactly one use, so the declaration was only moving the number away from the lookup it belongs to. The fallback now sits at the end of each chain, which makes the custom → catalog → default precedence readable in one line.

`bun test` on the four affected suites is still 108 pass / 0 fail, and `bun run --filter @nakama/web build` exits 0.

PR #989, opened 2026-09-12
feat(config): support mounted channel bot tokens

isnaenihidayat:
## Summary

Mount Telegram and Discord bot tokens as files → channel workers start without putting those secrets in environment values or saved settings.

**Before:** `TELEGRAM_BOT_TOKEN` and `DISCORD_BOT_TOKEN` had to contain the token directly, or the token lived in `config.ini`.
**After:** `TELEGRAM_BOT_TOKEN_FILE` and `DISCORD_BOT_TOKEN_FILE` accept readable secret paths; direct token values still win.

Why safe:
1. Reuses the existing mounted-secret resolver, including trimming and explicit file-read failures
2. Shared regression coverage exercises both channel resolvers and direct-value precedence
3. Existing env variables and saved channel configs keep their behavior

Only follow up if WhatsApp auth state or other integration secrets need an equivalent external-secret path.

## Test plan
- [x] `bun test packages/core/src/telegram-config.test.ts packages/core/src/discord-config.test.ts` (45 pass)
- [x] Targeted Ultracite check and `bun run knip`
- [x] `bun run build` and `bun run build:docs`
- [ ] Start each channel worker once with only its `_FILE` variable set

Progresses #364

### DIFF
--- a/docs/website/content/docs/discord.mdx
+++ b/docs/website/content/docs/discord.mdx
@@ -195,6 +195,9 @@ Environment-based setup is also supported. The main env var is:
 DISCORD_BOT_TOKEN
 ```
 
+For mounted secrets, set `DISCORD_BOT_TOKEN_FILE` to the path of a readable
+file containing the token. A directly set `DISCORD_BOT_TOKEN` takes priority.
+
 Nakama also supports:
 
 ```text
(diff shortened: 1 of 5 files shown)


ahmadrosid (commented):
LGTM

PR #990, opened 2026-09-12
feat(orgs): deliver member invitations by email

isnaenihidayat:
## Summary

Invite an organization member → they receive a working acceptance link through the configured mailbox.

**Before:** creating an invite always exposed a raw token for manual sharing.
**After:** successful email delivery returns no raw token; failed or unavailable delivery keeps the manual-token fallback.

Why safe:
1. Existing token hashing, expiry, single-use acceptance, and role checks remain unchanged
2. SMTP failures preserve the created invite and return its token instead of blocking onboarding
3. The acceptance page uses the existing public endpoint and refreshes the authenticated organization session

Only move delivery to a background job if SMTP latency starts delaying invite creation.

## Test plan
- [x] `bun test apps/server/src/services/org-service.test.ts --timeout 15000` (38 pass)
- [x] `bun run --cwd packages/core build && bun run --cwd packages/client build && bun run --cwd apps/server build && bun run --cwd apps/web build`
- [x] `bun run knip`
- [ ] Configure SMTP, invite one new address, and accept from the emailed link

Part of #430

### DIFF
--- a/apps/server/src/services/org-service.test.ts
+++ b/apps/server/src/services/org-service.test.ts
@@ -1,7 +1,11 @@
 import { describe, expect, test } from "bun:test";
 import { readFile } from "node:fs/promises";
 import { join } from "node:path";
-import { getProfileSoulDir, NakamaApiError } from "@nakama/core";
+import {
+  type EmailOutboundAdapter,
+  getProfileSoulDir,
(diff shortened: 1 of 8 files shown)


ahmadrosid (commented):
LGTM

PR #998, opened 2026-09-13
fix(channels): omit scratch artifacts from Telegram/Discord share footers

### LINKED ISSUE #981: [telegram][discord] Turn share links include scratch writes
### Summary

Telegram and Discord publish a public share link for every write in a turn, scratch files included, and post them in the chat footer. `deliverTurnArtifactShares` (`packages/core/src/channel-artifact-delivery.ts:241`) feeds `extractPairedTurnArtifacts` straight into `mintDeliverableArtifacts` with no filter.

This is the sibling of #975 on the other two channels. #977 fixes the WhatsApp document path by filtering at the two WhatsApp send sites, so the shared publish path keeps the behaviour.

### Measured on

c4572e6f (neither `channel-artifact-delivery.ts` nor `channel-artifacts.ts` has changed since #911)

### Reproduction

Call `deliverTurnArtifactShares` with a turn that writes `artifacts/laporan-final.csv` and `artifacts/_scratch-debug.json`, a fake `publish` and a `sendFooter` that collects. Both reach the footer.

### Expected vs actual

Expected one link. Actual:

```
delivered: laporan-final.csv, _scratch-debug.json
footer:
laporan-final.csv: https://app.example/s/tok
_scratch-debug.json: https://app.example/s/tok
```

No scratch filter on the shared path. #977 adds `isScratchArtifactPath` in `packages/core/src/channel-artifacts.ts`, so the fix is to apply it to `paired` inside `deliverTurnArtifactShares`.

Filtering inside `extractPairedTurnArtifacts` instead would cover every caller in one line, but it also drops scratch from the WhatsApp registry, which removes the `/attach` retrieval #977 deliberately kept there. Pick the first unless that retrieval is not wanted.

Correction: on Telegram and Discord the `/attach` registry is built from the minted shares, so filtering in `deliverTurnArtifactShares` keeps scratch out of their registry as well. `/attach` there is most-recent-only, so it then returns the last real file instead of a scratch one.

akmalsyrf:
## Summary

Telegram/Discord turn footers only share real deliverables — scratch/`_`/`.`/`.tmp` writes stay out of the link list.

**Before:** `deliverTurnArtifactShares` minted and footered every paired write, including `_scratch-debug.json`.
**After:** paired artifacts are filtered with `isScratchArtifactPath` before mint/publish/footer.

Why safe:
1. Filter is local to the shared publish path; WhatsApp still fills its registry separately (#977)
2. On Telegram/Discord the registry is built from `delivered`, so `/attach` also skips scratch and returns the last real file
3. Same scratch heuristics as #977; real-file mint/footer behavior unchanged

Only re-add / follow up if a real deliverable uses a scratch-like filename on purpose.

## Test plan
- [x] `bun test packages/core/src/channel-artifacts.test.ts packages/core/src/channel-artifact-delivery.test.ts` (29 pass)
- [x] Existing share fixture includes `_scratch-debug.json` — footer + registry only keep `report.md`
- [ ] Telegram/Discord live turn that writes `artifacts/report.md` + `artifacts/_scratch.json` — footer has only `report.md`

Fixes #981


packages/core/src/channel-artifact-delivery.ts
```diff
@@ -224,7 +225,9 @@ export async function deliverTurnArtifactShares(input: {
   };
 }): Promise<DeliverableChannelArtifact[]> {
   const messages = await input.session.getMessages();
-  const paired = extractPairedTurnArtifacts(messages);
+  const paired = extractPairedTurnArtifacts(messages).filter(
```
fajarhide, line 228: This also keeps scratch out of the `/attach` registry on Telegram and Discord, because the registry further down is built from `delivered`. My issue said filtering here keeps the `/attach` retrieval #977 kept, and that is only true on WhatsApp, where the registry is filled separately. Here `/attach` is most-recent-only, so in practice it now returns the last real file instead of a scratch one. I think that's what we want, it's just worth a line in the PR body so nobody reads it as a regression later.

packages/core/src/channel-artifact-delivery.test.ts
```diff
@@ -345,4 +345,128 @@ describe("deliverTurnArtifactShares", () => {
 
     expect(delivered).toEqual([]);
   });
+
+  test("omits scratch writes from share footers", async () => {
```
fajarhide: The test at L216 already builds this fixture and tracks the registry. Adding a `_scratch-debug.json` write and its meta to that message list, then asserting the footer and `registry` only hold `report.md`, covers the same path in far fewer lines. It also pins the registry side effect, which this test can't see because `updateArtifactState` is a no-op here.

fajarhide (commented):
Filter is in the right place, and the new test fails without it (29 pass, 1 fail with the filter removed). The test can fold into the existing fixture, and one side effect is worth stating. I also checked it merges cleanly with #977 (`git merge-tree` exit 0, one `isScratchArtifactPath` after the merge); I didn't run Telegram or Discord live.

packages/core/src/channel-artifact-delivery.ts
akmalsyrf: Agreed — updated the PR body. On Telegram/Discord the registry is built from `delivered`, so `/attach` now returns the last real file instead of scratch. WhatsApp still fills its registry separately (#977).

packages/core/src/channel-artifact-delivery.test.ts
akmalsyrf: Done — folded the scratch write into the existing fixture and asserted footer + registry only keep `report.md`.

ahmadrosid:
LGTM

PR #999, opened 2026-09-13
feat(org-memory): record optional source document ids on proposals

### LINKED ISSUE #984: memory: bullets record the session they came from but never the source document
### Summary

An org memory bullet records which session it came from but never which document. `OrgMemoryProposal` (`packages/core/src/contract.ts:743`) carries `profileId`, `sessionId`, `proposedByUserId`, `reviewerUserId` and `reviewedAt`, and no document reference. `git grep documentId` across `packages/core/src/soul/` and the memory service returns nothing.

### Why it matters

The knowledge base already records where a document came from: every extracted file starts with `# source:`, `# mediaType:` and `# uploadedAt:`, and the manifest holds a SHA-256 per document. The memory side records where a bullet came from as well. The two halves are never joined, so the question an operator actually asks cannot be answered: this bullet claims X, which uploaded document made the agent believe it?

Reading the session transcript is the only route today, and it only works while the session still exists and the bullet text has not been edited since. The published bullet in `MEMORY.md` is a plain line with no id back to its proposal row either, so the link is by text match through `normalizeOrgMemoryDedupKey`. Edit the bullet and even that breaks.

This gets worse as the knowledge base grows, because a bullet summarising a document is exactly the case where the source matters most.

- `org_memory_proposals` in `packages/db/sql/schema.sql`, and `OrgMemoryProposal` in `packages/core/src/contract.ts:743`
- `propose_org_memory` in `apps/server/src/tools/org-memory-tools.ts:87`, whose input is only `bullet`
- `apps/server/src/services/org-memory-service.ts`
- `apps/web/src/components/settings/OrgMemoryProposalsPanel.tsx`

### Acceptance criteria

- [ ] `propose_org_memory` accepts optional source document ids and the agent is told to pass them when the bullet came from a document
- [ ] the ids are stored on the proposal and returned by `GET /v1/orgs/:orgId/memory/proposals`
- [ ] the proposals panel links each cited document, and says plainly when a bullet cites none
- [ ] a source that no longer exists renders as a removed document rather than a broken link or a raw id
- [ ] proposals created before this change keep loading with no source
- [ ] a test proposes a bullet with a document id and asserts it survives approval

akmalsyrf:
## Summary

Org memory proposals can cite knowledge-base documents; Settings shows filenames (or “Removed document”) and empty-source when none were passed.

**Before:** proposals only recorded session/profile/user — no link back to the uploaded doc that justified the bullet.
**After:** optional `sourceDocumentIds` on `propose_org_memory` (id or filename), resolved to real KB ids, stored on the proposal, returned by list/get, kept through approval.

Why safe:
1. Column is nullable; legacy rows map to `[]`
2. Tool resolves id/filename against the profile KB and drops invented values
3. Catalog prints `[id: …]` so agents can pass either form; UI waits for KB load before “Removed document”

Only re-add / follow up if we later need org-scoped KB lookup when `profileId` is null.

## Test plan
- [x] `bun test apps/server/src/services/org-memory-service.test.ts apps/server/src/tools/org-memory-tools.test.ts` (34 pass)
- [x] Tool resolves `refund-policy.txt` → real id; invented names dropped; approval keeps ids
- [ ] Propose a bullet with a filename cite, open Settings → Org Memory proposals — filename links; delete the doc — shows Removed document

Fixes #984


apps/server/src/tools/org-memory-tools.ts
```diff
@@ -109,6 +126,7 @@ export function createOrgMemoryTools(
           profileId: context.profileId ?? null,
           proposedByUserId: context.userId ?? null,
           sessionId: context.sessionId ?? null,
+          sourceDocumentIds: readStringArray(input, "sourceDocumentIds"),
```
fajarhide: I think this blocks. The agent is told to pass document ids, but it never sees one. I uploaded `refund-policy.txt` on this branch and printed what the agent gets:

```
DOCUMENT ID: kb_e9368a54-6e04-4d60-9e92-fd21e4cfca9f
CATALOG:
- refund-policy.txt (text/plain)
SEARCH: [{"file":".../knowledge-base/kb_e9368a54-6e04-4d60-9e92-fd21e4cfca9f.extracted.txt","line":4,...}]
```

The catalog and the `filename` parameter on `knowledge_base_search` both speak filenames, and the id only shows up inside a match path. So the likely value here is `refund-policy.txt`. It gets stored as-is, and `OrgMemoryProposalsPanel.tsx:112` then renders "Removed document" for a file that still exists.

Cheapest fix: resolve each entry here against `listKnowledgeBaseDocuments(context.orgId, context.profileId)` by id or filename, store the real id and drop anything that doesn't match. That also keeps invented ids out. Printing the id on the catalog line is the other option. I haven't checked what a real model passes, this comes from the prompt and tool output only.

apps/server/src/tools/org-memory-tools.ts
```diff
@@ -42,6 +42,17 @@ function readString(input: unknown, key: string): string | null {
   return typeof value === "string" && value.trim() ? value.trim() : null;
 }
 
+function readStringArray(input: unknown, key: string): string[] {
```
fajarhide: `normalizeSourceDocumentIds` in the service already skips non-strings, so this could pass the raw array through and drop the helper.

apps/server/src/services/org-memory-service.test.ts
```diff
@@ -250,6 +251,51 @@ describe("OrgMemoryService", () => {
...
+      "admin_user"
+    );
+    expect(approved.sourceDocumentIds).toEqual(["kb_handbook", "kb_faq"]);
+    expect(
+      (await service.getProposal("org_a", proposed.proposalId!))
+        .sourceDocumentIds
+    ).toEqual(["kb_handbook", "kb_faq"]);
+  });
+
+  test("proposals without source document ids load as an empty list", async () => {
```
fajarhide: The propose test above already asserts `sourceDocumentIds: []` for a proposal created without ids (line 210), so this one doesn't add much on top of that.

apps/web/src/components/settings/OrgMemoryProposalsPanel.tsx
```diff
@@ -69,6 +71,72 @@ function resolveProposer(
...
 }
 
+function ProposalSources({
+  proposal,
+  variant = "compact",
+}: {
+  proposal: OrgMemoryProposal;
+  variant?: "compact" | "detail";
+}) {
+  const sourceDocumentIds = proposal.sourceDocumentIds ?? [];
```
fajarhide: `sourceDocumentIds` is `string[]` in the contract, so the `?? []` can go.

fajarhide (commented):
Migration, the five SELECTs and keeping ids through approval all look right. The ids the tool asks for are never shown to the agent, so I think that part blocks, and a few spots can shrink. I ran the tests and a probe on 7c99dfdc; I didn't try it with a real model or in a browser.

apps/server/src/tools/org-memory-tools.ts
akmalsyrf: Valid block — fixed. `propose_org_memory` now resolves each entry against `listKnowledgeBaseDocuments` by id or filename and drops unknowns; the KB catalog line also prints `[id: …]` so the agent can pass either form.

apps/server/src/tools/org-memory-tools.ts
akmalsyrf: Done — dropped `readStringArray`; the raw array goes through resolve → service normalize.

apps/server/src/services/org-memory-service.test.ts
akmalsyrf: Removed the redundant empty-list test.

apps/web/src/components/settings/OrgMemoryProposalsPanel.tsx
akmalsyrf: Dropped the `?? []`.

PR #1000, opened 2026-09-13
fix(org-memory): surface history revision cap and truncation

### LINKED ISSUE #982: memory: org memory history drops revisions past 50 with no signal to the reader
### Summary

Every append to org memory history prunes the directory to the 50 most recent revisions. Nothing in the API response or the UI says the cap exists, so an org that edits memory regularly loses restore points with no signal.

### Why it matters

The revision history is what makes org memory auditable: who changed a bullet, when, and what the file looked like before. An org reviewing a handful of proposals a week crosses 50 revisions in a couple of months, and the oldest restore points disappear silently. Someone hunting for the change that introduced a wrong bullet just finds a list that starts later than expected, with no way to tell whether that is the beginning of history or the edge of the cap. The cap itself is a reasonable default. Being invisible is the problem.

- `packages/core/src/soul/org-memory-history.ts:11` defines `ORG_MEMORY_HISTORY_MAX_ENTRIES = 50`, `:77` prunes on every append, `:137` is `pruneOrgMemoryHistory`
- `apps/server/src/http/routes/org-memory.ts:494` serves `GET /v1/orgs/:orgId/memory/history`
- `ListOrgMemoryHistoryResponse` in `packages/core/src/contract.ts`
- `apps/web/src/components/settings/OrgMemoryHistoryPanel.tsx`

### Acceptance criteria

- [ ] `GET /v1/orgs/:orgId/memory/history` reports the cap and whether pruning has already discarded revisions
- [ ] `OrgMemoryHistoryPanel` tells the reader when older revisions have been dropped, rather than rendering a list that looks complete
- [ ] the cap is stated in the docs where an operator will find it
- [ ] existing clients that ignore the new field keep working
- [ ] a test appends past the cap and asserts the response reports it

akmalsyrf:
## Summary

Org memory history API + Settings tell admins the 50-revision soft cap exists, and when older snapshots were already pruned.

**Before:** prune silently kept 50 newest revisions; UI looked like a complete timeline.
**After:** `GET …/memory/history` returns `maxEntries` + `truncated`; panel and docs state the cap.

Why safe:
1. New fields are additive — old clients that only read `changes` keep working
2. `truncated` is durable (`truncated.json` marker) and ignored by history listing
3. Cap value unchanged (`ORG_MEMORY_HISTORY_MAX_ENTRIES = 50`)

Only re-add / follow up if orgs pruned before this ship need a backfill signal (marker starts false until the next prune).

## Test plan
- [x] `bun test packages/core/src/soul/org-memory-history.test.ts` (7 pass)
- [x] Append past the soft cap via `appendOrgMemoryHistory` — `truncated: true` and length 50
- [ ] Settings → Org Memory → History after >50 edits — banner says older snapshots were removed

Fixes #982


fajarhide (commented):
Marker logic is right: it is only written after an actual unlink, and the listing skips it. The new test doesn't pin that yet, and a couple of exports can shrink. I ran the tests, a mutation and a probe on 63fd219e; I didn't open the panel in a browser.

packages/core/src/soul/org-memory-history.test.ts
```diff
@@ -241,14 +243,43 @@ describe("org memory history", () => {
...
+    ).resolves.toBeFalsy();
     await expect(
       access(path.join(historyDir, `${malformedId}.md`))
-    ).resolves.toBeNull();
+    ).resolves.toBeFalsy();
     await expect(
       access(path.join(historyDir, "omh_00000001_valid.json"))
     ).rejects.toThrow();
   });
+  test("reports the revision cap after pruning discards older entries", async () => {
```
fajarhide: If `markOrgMemoryHistoryTruncated` moves above the `stale.length === 0` return, so every append writes the marker, both test files still pass (29 pass, 0 fail). This test calls `pruneOrgMemoryHistory(orgId, 2)` directly, and #982 asks for appending past the cap. Going through `appendOrgMemoryHistory` catches it, this fails on that mutation with `Expected: false / Received: true`:

```ts
for (let i = 0; i < 3; i += 1) await append(orgId, i);
expect((await listOrgMemoryHistoryWithCap(orgId, 100, tempDir)).truncated).toBe(false);

for (let i = 3; i <= ORG_MEMORY_HISTORY_MAX_ENTRIES; i += 1) await append(orgId, i);
const listing = await listOrgMemoryHistoryWithCap(orgId, 100, tempDir);
expect(listing.changes).toHaveLength(ORG_MEMORY_HISTORY_MAX_ENTRIES);
expect(listing.truncated).toBe(true);
```

`append` is the `appendOrgMemoryHistory` call you already have, with a distinct `createdAt` per index.

packages/core/src/soul/org-memory-history.ts
```diff
@@ -10,6 +10,54 @@ import { assertConfigPathSegment, getOrgMemoryHistoryDir } from "./resolve";
...
+const ORG_MEMORY_HISTORY_TRUNCATED_MARKER = "truncated.json";
+
+function historyTruncatedMarkerPath(orgId: string, configDir?: string): string {
+  return join(
+    getOrgMemoryHistoryDir(orgId, configDir),
+    ORG_MEMORY_HISTORY_TRUNCATED_MARKER
+  );
+}
+
+export async function isOrgMemoryHistoryTruncated(
```
fajarhide: `listOrgMemoryHistoryWithCap` already returns `truncated`, so this is only exported for the test. It could stay private and the test asserts `listing.truncated`.

packages/core/src/soul/org-memory-history.ts
```diff
@@ -10,6 +10,54 @@ import { assertConfigPathSegment, getOrgMemoryHistoryDir } from "./resolve";
...
+    `${JSON.stringify({ truncatedAt: new Date().toISOString() }, null, 2)}\n`,
+    { ensureDir: historyDir }
+  );
+}
+
+export async function listOrgMemoryHistoryWithCap(
+  orgId: string,
+  limit = ORG_MEMORY_HISTORY_MAX_ENTRIES,
+  configDir?: string
+): Promise<{
```
fajarhide: This is `ListOrgMemoryHistoryResponse`, and the file already imports from `../contract`.

apps/web/src/components/settings/OrgMemoryHistoryPanel.tsx
```diff
@@ -312,10 +314,19 @@ export function OrgMemoryHistoryPanel({ orgId }: { orgId: string }) {
...
-        <p className="text-muted-foreground text-xs">
-          Timeline of every change. View snapshots or revert to an earlier
-          revision.
-        </p>
+        <div className="min-w-0 space-y-1">
+          <p className="text-muted-foreground text-xs">
+            Timeline of every change. View snapshots or revert to an earlier
+            revision.
+          </p>
+          {typeof maxEntries === "number" ? (
```
fajarhide: Web and server ship together, so `maxEntries` is always a number here. You could drop the guard.

packages/core/src/soul/org-memory-history.test.ts
akmalsyrf: Done — the test now appends via `appendOrgMemoryHistory` past `ORG_MEMORY_HISTORY_MAX_ENTRIES` and asserts `truncated: true` (no direct `prune` call).

packages/core/src/soul/org-memory-history.ts
akmalsyrf: Made `isOrgMemoryHistoryTruncated` private; the test asserts `listing.truncated`.

packages/core/src/soul/org-memory-history.ts
akmalsyrf: Switched the return type to `ListOrgMemoryHistoryResponse`.

apps/web/src/components/settings/OrgMemoryHistoryPanel.tsx
akmalsyrf: Dropped the typeof guard; panel always reads `maxEntries`.

PR #1004, opened 2026-09-13
fix(orgs): permanently delete archived organizations

isnaenihidayat:
## Summary

Permanently delete an archived organization → its tenant rows and on-disk data are removed while user accounts remain available for separate erasure.

**Before:** `DELETE /v1/platform/orgs/:orgId` only archived the organization, leaving its database and filesystem data indefinitely.
**After:** platform admins can call `DELETE /v1/platform/orgs/:orgId/permanent` after archival for a full tenant purge.

Why safe:
1. Active organizations return 409, non-platform users return 403, and the existing archive endpoint keeps its behavior
2. The validated org directory is removed before the database cascade; active session pointers are cleared and migrated org-owned resources are deleted in one transaction

Only add a dashboard action when archived organizations have a reviewable management view with an explicit permanent-delete confirmation.

## Test plan
- [x] `bun test packages/db/src/organization-archive.test.ts apps/server/src/services/org-service.test.ts apps/server/src/http/platform-orgs.test.ts --timeout 15000` (57 pass)
- [x] `bun test packages/db/src --timeout 15000`, `bun run knip`, and database/server builds (113 database tests pass)
- [ ] Archive a disposable organization, call the permanent endpoint, and confirm its org directory no longer exists

Part of #368


ahmadrosid (commented):
LGTM

PR #1007, opened 2026-09-14
feat(auth): add secure password reset flow

isnaenihidayat:
## Summary

Recover a forgotten local password → receive a one-hour reset link and return to sign-in without operator database access.

**Before:** forgotten passwords required direct server or database intervention.
**After:** `/reset-password` requests and completes a hashed, expiring, single-use reset; existing sessions and outstanding reset tokens are revoked.

Why safe:
1. Token consumption, password replacement, outstanding-token invalidation, and session revocation commit in one transaction
2. Public responses do not reveal account existence; raw fallback tokens require platform-admin authentication plus CSRF, and both endpoints use the strict credential rate limit

Only change the neutral public response if delivery telemetry can remain free of account-enumeration signals.

## Test plan
- [x] `bun test packages/db/src/migrate.test.ts apps/server/src/services/org-service.test.ts apps/server/src/http/org-invites.test.ts apps/server/src/http/rate-limit-middleware.test.ts --timeout 15000` (82 pass)
- [x] `bun test packages/db/src --timeout 15000`, `bun run knip`, targeted formatting, and database/core/client/server/web builds (115 database tests pass)
- [ ] Configure SMTP, request a reset from sign-in, open the email link, reset, then sign in with the new password

Part of #430


packages/db/src/adapters/sqlite.ts
```diff
@@ -1677,6 +1677,52 @@ function createSqliteDatabaseAdapter(db: Database): DatabaseAdapter {
...
+      if (!token) {
+        return false;
+      }
+
+      const consumed = consumePasswordResetTokenStmt.run(
+        consumedAt,
+        tokenHash,
+        consumedAt
+      );
+      if (consumed.changes !== 1) {
```
ahmadrosid: You can drop `consumePasswordResetTokenStmt` and this `consumed.changes` check. This runs in an immediate transaction, so another writer cannot consume the token between the lookup and update. `consumePasswordResetTokensForUserStmt` already consumes this token along with the others; keeping the lookup guard and the remaining updates preserves the single-use behavior with one fewer statement.

ahmadrosid (commented):
One small cut in the token-consumption transaction.

packages/db/src/adapters/sqlite.ts
isnaenihidayat: Updated in 63b60475. Removed the redundant single-token update and changes check; the immediate transaction now uses the usable-token lookup as its guard, then consumes every outstanding token for that user before changing the password and revoking sessions. The focused suite still passes (82 tests).

apps/server/src/services/org-service.ts
```diff
@@ -953,6 +957,80 @@ export class OrgService {
...
+    const email = normalizeEmail(requestedEmail);
+    if (!EMAIL_PATTERN.test(email)) {
+      throw new NakamaApiError("A valid email address is required.", 400);
+    }
+
+    const user = await this.databaseAdapter.getUserByEmail(email);
+    if (!user) {
+      // Match the successful-delivery shape so configured deployments do not
+      // disclose whether an address has an account.
+      return { delivered: true, token: null };
```
ahmadrosid, line 973: Unknown emails return immediately here, while existing accounts wait for SMTP. Even with the same response body, that timing difference can reveal whether an account exists. Can we make the public response independent of email delivery and add a check for this?

apps/server/src/services/org-service.ts
```diff
@@ -953,6 +957,80 @@ export class OrgService {
...
+      id: `password_reset_${crypto.randomUUID().replace(/-/g, "")}`,
+      tokenHash: this.authService.hashToken(token),
+      userId: user.id,
+    };
+    await this.databaseAdapter.createPasswordResetToken(record);
+
+    const webPublicUrl = this.getWebPublicUrl();
+    const resetInstruction = webPublicUrl
+      ? `Reset your password: ${webPublicUrl}/reset-password?token=${encodeURIComponent(token)}`
+      : `Password reset token: ${token}`;
```
ahmadrosid, line 993: If no public URL is configured, the email only contains a token, but the reset page has no field to paste it into. Can we add a way to enter that token so this fallback works without manually editing the URL?

apps/server/src/services/org-service.ts
isnaenihidayat: Updated in d8f958d0. Public requests now dispatch delivery without awaiting SMTP and return the same neutral response before delivery settles; the authenticated platform-admin fallback still awaits delivery so it can return a manual token when needed. Added a regression test with a deliberately blocked sender to verify the public response completes independently.

apps/server/src/services/org-service.ts
isnaenihidayat: Updated in d8f958d0. The reset page now offers “Use reset token” and renders a required token field for manual paste; tokens from email links still prefill that field. Added render tests for both the manual-entry option and a token-bearing link.

Issue #1011, opened 2026-09-14
Artifact links from another profile request the active profile and return 500


tansilandre:
## Summary

An artifact path mentioned in chat is resolved against the currently active profile, not the profile that owns the artifact. The resulting `GET /v1/profiles/:profileId/artifacts/content` returns 500 when the file is absent.

Observed at `apps/web/src/lib/chat-artifacts.ts:484` and `apps/server/src/http/routes/profiles.ts:784`.

## Measured on

`61bb5072` (`main`)

## Reproduction

1. In one organization, create two profiles, for example `Super Bot` and `Video Producer`.
2. Create an artifact only in the Video Producer workspace:
   `~/.nakama/orgs/<orgId>/profiles/video-producer/artifacts/ramadhan-nulife/shots/review-frames.html`
3. Let a Super Bot chat message or its memory reference the relative artifact path `artifacts/ramadhan-nulife/shots/review-frames.html`.
4. Open that inferred artifact while Super Bot is the active profile.
5. The web app builds this URL with Super Bot's profile ID:
   `GET /v1/profiles/<super-bot-id>/artifacts/content?path=ramadhan-nulife%2Fshots%2Freview-frames.html&inline=1`

The same failure can be observed by requesting any absent artifact path from a valid profile.

## Expected vs actual

Expected: an artifact reference retains its owning profile and loads the file from that workspace. If a file is genuinely absent, the API returns a 404 response.

Actual: the URL uses the active profile ID, so the server tries to read a nonexistent path. The route returns HTTP 500:

```text
GET http://localhost:3003/v1/profiles/kOKP1lN2AXCKUsWOOftsg/artifacts/content?path=ramadhan-nulife%2Fshots%2Freview-frames.html&inline=1 500 (Internal Server Error)
```

The file exists under `profiles/video-producer`, but not under `profiles/kOKP1lN2AXCKUsWOOftsg` (Super Bot).

## Root cause or proposed fix

`apps/web/src/lib/chat-artifacts.ts` extracts bare relative paths from assistant text and `buildArtifactContentUrl(profileId, artifactPath)` supplies the active profile ID. The path has no source-profile provenance.

`packages/core/src/artifacts.ts:readArtifactFile` then calls `stat(filePath)` without translating `ENOENT`; `apps/server/src/http/app.ts` converts that plain error into a generic 500.

[...]

PR #1018, opened 2026-09-14
fix(artifacts): answer 404 for a missing artifact instead of 500

fajarhide:
## Summary

Opening an artifact whose file is gone now shows "Artifact not found: <name>" instead of a server error.

**Before:** `readArtifactFile` let `ENOENT` and `ENOTDIR` from `realpath` and `stat` escape as plain errors, so `GET /v1/profiles/:profileId/artifacts/content` answered 500 and logged a stack for every stale link.
**After:** a missing file, a profile with no `artifacts/` dir yet, or a path through a file throws `NakamaApiError(404)`, which `app.onError` already maps. The message names only the basename.

| Before (`main`) | After (this branch) |
| --- | --- |
| before | after |

A document that still exists opens as before (this branch, same page):

existing

All three are Files, then View on `review-frames.html`, on a throwaway instance. For the first two the file is deleted from disk after the list loads, which is what a stale chat or list link hits.

Why safe:
1. Only `ENOENT` and `ENOTDIR` are translated. Any other fs error still surfaces as a 500.
2. The share publish and refresh paths call the same function, so a missing source there is a 404 too.
3. The Files page passes the absolute path `listArtifacts` returns. The first commit echoed that path into the 404 body and the screenshot caught it; the second commit names only the basename, and a test pins that.

Only follow up if the owner-profile half of #1011 is built (agreed in the issue: owner-qualified references). `writeArtifactFile` has the same `realpath` on a missing `artifacts/` dir, so a `PUT` there is also a 500; left alone because the editor only saves artifacts that already exist.

Part of #1011

## Test plan
- [x] Browser run above: `/artifacts/content` status 500 on `main`, 404 on this branch, 200 for an existing file.
- [x] Live API against the same kind of instance, same script on `main`'s `artifacts.ts` and this branch:

```
BEFORE (main)
no artifacts dir yet: shots/review-frames.html -> 500 {"error":"An unexpected server error occurred."}
missing file: shots/review-frames.html         -> 500 {"error":"An unexpected server error occurred."}
through a file: report.md/nested.md            -> 500 {"error":"An unexpected server error occurred."}
existing file: report.md                       -> 200 # Report
server log lines mentioning ENOENT/ENOTDIR: 6

AFTER (this branch)
no artifacts dir yet: shots/review-frames.html -> 404 {"error":"Artifact not found: review-frames.html"}
missing file: shots/review-frames.html         -> 404 {"error":"Artifact not found: review-frames.html"}
through a file: report.md/nested.md            -> 404 {"error":"Artifact not found: nested.md"}
existing file: report.md                       -> 200 # Report
server log lines mentioning ENOENT/ENOTDIR: 0
```

- [x] `bun test packages/core/src/artifacts.test.ts` (13 pass). The new case fails on `main` with `ENOENT: no such file or directory, stat '.../artifacts/shots/review-frames.html'`. Removing the `ENOTDIR` check, the `realpath` catch, or the basename each fails it again (12 pass, 1 fail).
- [x] `bun run test` on the first commit: every workspace 0 fail except `@nakama/server` (1319 pass, 55 fail). None touch artifacts: 54 are plugin tests (42 `package_unavailable`, no npm registry in this sandbox) and 1 is the SIGTERM harness probe that times out in a worktree. The second commit only changes the message and its test.


ahmadrosid (commented):
LGTM

PR #1020, opened 2026-09-14
feat(discord): list and resume a chat's earlier sessions

### LINKED ISSUE #991: Add Discord commands to list and resume assistant sessions
## Problem

The Discord integration currently does not provide a way to switch between existing sessions. Users who start multiple conversations, or return after a previous conversation, cannot select an earlier session and continue it.

## Proposed feature

Add session management commands to the Discord integration:

- /sessions lists the user’s available sessions.
- Each session shows a short identifier, title or first message, and last activity time.
- /resume <session> switches the current Discord conversation to the selected session.
- New messages after resuming are appended to that session.
- The active session is shown or otherwise discoverable in the Discord UI.
- Invalid, expired, or inaccessible session IDs return a clear error.
- Users can only list and resume sessions they are authorized to access.

## Acceptance criteria

- A Discord user can list their existing sessions.
- A user can select a session from the list and resume it.
- The resumed session retains its previous conversation history and context.
- Starting a new session still works.
- Session access is isolated between users and workspaces where applicable.
- The feature works with Discord threads and direct messages, if both are supported.
- Automated tests cover listing sessions, resuming sessions, invalid session IDs, and unauthorized access.
- Documentation includes the available commands and examples.

## Open questions

- Should session selection use slash-command arguments, Discord buttons, or a select menu?
- Should sessions be scoped to the Discord user, channel, thread, or connected Nakama account?
- How many sessions should /sessions display at once?
- Should inactive sessions expire or remain resumable indefinitely?

fajarhide:
## Summary

Discord users pick an earlier conversation from `/sessions`, or run `/resume session:<id>`, and keep talking in it. Closes #991.

**Before:** `/new` replaced the chat's session and nothing in Discord could reach the old one again.
**After:** each Discord chat keeps the IDs of its last 25 sessions (`ChatSessionRecord.sessionIds`). `/sessions` and `/resume` only accept those, and `/status` shows the active one.

Why safe:
1. Every Discord user reaches the API as `discord@nakama.internal`, and `GET /v1/sessions` filters by profile and channel only. The chat's own ID list is therefore the access check, and removing it makes both new tests fail.
2. The list is kept per DM, thread or channel. A DM session never shows up in a server thread, and switching org still deletes the record.
3. Resume only rebinds the ID. The next message goes through `resolveSession`, which already starts a new session if the old one was deleted.

Follow up only if people want to continue a DM conversation inside a thread. That needs a per-user list and a call on showing DM history in a shared thread.

## Test plan
- [x] `bun run test`: 3496 pass, 54 fail. All 54 are server plugin tests that need npm network, and `apps/server` on the base commit fails the same 54.
- [x] Returned the unfiltered server list from `listChatSessions`: both new tests fail (0 pass, 2 fail). Restored: 2 pass
- [ ] DM the bot, send a message, `/new`, send another, then `/sessions` and pick the first one. The next reply continues it. I have not run this in a Discord client, so this PR has no screenshot.


ahmadrosid (commented):
LGTM

PR #1025, opened 2026-09-14
feat(chat): cognito mode for sessions that are never saved or written to memory

### LINKED ISSUE #373: Enterprise compliance: configurable data retention policies
> **Enterprise compliance gap** — identified in the nakama security & compliance audit (2026-08-21). Required for GDPR Art. 5(1)(e) storage limitation and operational cost/eDiscovery control.

# Summary

Nothing expires automatically. Chat transcripts, artifacts, attachments, usage stats, and automation runs persist forever unless someone manually purges a session. Docs state it outright: "Nothing is deleted." (`docs/website/content/docs/token-optimiser.mdx:67`).

# Why this blocks enterprise compliance

- Storage-limitation principle requires defined retention periods for personal data.
- Unbounded growth of transcripts containing private conversations expands breach and eDiscovery scope.

# Current state (verified)

- Grep `retention|purge|prune|ttl|cleanupExpired|deleteOlderThan` across `apps/server/src`: only user-initiated `purgeSession` (`agent-service.ts:1775`) and coding-agent log pruning to 10 files (`bash.ts:29`).
- All `setInterval` uses are SSE keepalives, not cleanup jobs.
- Channel workers additionally log message content to stdout indefinitely (filed as #354).
- Docs confirm permanence by design; skill curator archives rather than deletes.

# Proposed approach

1. Configurable retention (days) per data class: chat messages/sessions, artifacts, attachments, automation runs, usage stats, (future) audit events.
2. Background sweeper honoring the single-node design (runs inside the existing server process), with dry-run/report mode.
3. Disabled by default; documented defaults for regulated deployments.

# Acceptance criteria

- [ ] Operator can set per-class TTLs via config/env.
- [ ] Sweeper deletes DB rows AND corresponding files (attachments/artifacts) — no orphans.
- [ ] Dry-run report lists what would be deleted.

### LINKED ISSUE #1024: Cognito chat mode: sessions that are never persisted and never written back to memory
## Summary

A **cognito** toggle in the web chat, in the top right corner of the chat column. A cognito session lives only in server memory: it never reaches `sessions` or `session_messages`, never shows up in History, and nothing it produces is written back to memory.

Two sub-modes, picked from a dropdown that appears next to the toggle once it is on:

1. **Personalized.** The chat still uses profile memory, org memory, plugins, skills, MCP, Composio and the agent's own instructions. It just leaves no trace behind.
2. **Not personalized.** None of that is loaded. A neutral system prompt, the neutral builtin tools, and the model you picked.

Both sub-modes block every write-back path. The difference is only what gets read in.

## Why

Every web chat today is durable by construction, and there is no way to opt out of any layer of it.

`AgentService.createSession()` (`apps/server/src/services/agent-service.ts:1648`) writes a `sessions` row before the first token, and `wrapPersistedSession()` (`apps/server/src/services/session-persistence.ts:181`) replays the whole history into `session_messages` after every turn. On top of that, each finished turn schedules `scheduleSessionTitleGeneration()` and `schedulePostTurnSkillReview()` from `apps/server/src/http/routes/sessions.ts:606`, so the session also earns a generated title and feeds the skill suggestion pipeline.

That is the right default for the work you want the agent to remember. It is the wrong default for the rest:

- Testing a prompt, a model or a new profile, where twenty throwaway sessions bury the real ones in History.
- Asking something personal or sensitive on a shared org server, where the operator can read `~/.nakama` and every admin can read org memory.
- Checking what the agent does **without** your accumulated context, which is impossible right now. There is no way to ask the same question with the soul stack, the skills catalog and `MEMORY.md` switched off, so you cannot tell which answers come from the model and which come from what it has been told about you.

Deleting the session afterwards does not cover it. By then the title is generated, the skill review has run, and whatever the agent chose to write into `MEMORY.md` is already there.

## Proposed scope and non-goals

**In scope:** web dashboard only. The server takes a generic `cognito` field on `CreateSessionRequest`, so CLI and the messaging channels can adopt it later without touching the server.

**Pure in-memory, deliberately.** No `ephemeral` column on `sessions`, no row that gets deleted later, nothing written and then cleaned up. The cost is accepted and should be stated in the UI: a browser refresh ends a cognito chat, and so does restarting the server. If it is worth persisting, it is not a cognito chat.

**Not in scope for a first pass:**

- CLI, Telegram, WhatsApp, Discord.
- Any cross-device or cross-tab resume of a cognito session.
- Files the agent writes into the profile workspace on your explicit instruction. Asking for a document in a cognito chat still produces a real file under `~/.nakama`. That is an action you asked for, not a trace the session left on its own.

**Contract.** `CreateSessionRequest` in `packages/core/src/contract.ts:902` gains `cognito?: { personalized: boolean }`. Absent means a normal session, so every existing caller is unaffected.

**Server.** Every session path funnels through `getSessionRecordForOrg()` (`agent-service.ts:3341`) and `resolveSession()` (`agent-service.ts:1947`), so an in-memory map checked ahead of those two covers `beginSessionTurn`, `getSessionMessages`, `updateSessionModel` and the rest without scattering branches through the file. `createSession()` skips `db.upsertSession()`, and `buildChatSession()` returns the raw session instead of wrapping it in `wrapPersistedSession()`. `listSessions()` needs no change at all: a session that was never written cannot be listed.

The seven `this.sessions.clear()` calls that fire on config changes must not touch that map, or editing a provider setting would kill a chat in progress.

**Blocking write-back.** Memory here is `MEMORY.md` and `memory-archive/YYYY-MM.md` under the profile soul dir (`apps/server/src/services/memory-backend-service.ts:70`), written through the ordinary file tools rather than a dedicated one. So it needs a `forbidMemoryWrites` flag on `ToolContext` and a guard in `packages/core/src/tools/builtin.ts`, following `refuseProfileSkillMarkdownWrite()` at line 192. Alongside that: drop the `org_memory_*` tools, drop `skill_manage` so `/learn` is inert, and skip the two post-turn schedulers.

**Not personalized** additionally skips the soul stack, the skills catalog, the agent-browser capability, the knowledge base catalog, the org memory section and the user context in `resolveProfileSystemPrompt()` (`agent-service.ts:3922`), and filters plugin, MCP, Composio, skill, org memory, automation and sub-agent tools out of `resolveProfileTools()` (`agent-service.ts:3407`).

**Attachments.** `attachments.session_id` is a foreign key into `sessions`, and a cognito session has no row there, so the column goes `NULL` and an `ephemeral` column marks the rows for cleanup. Purged when the session closes, when an idle session is evicted, and swept at server start for whatever a hard restart left behind.

**Web.** A new control at the top right of the chat column, portalled into `[data-page-header-actions]` on mobile, since `Layout.tsx:118` hides the app header on desktop chat. The session id is never written to `localStorage`, so a refresh cannot pick it back up. The sub-mode dropdown locks once the first message is sent, because changing it mid-session would swap the system prompt and the tool list underneath the conversation.

## Acceptance criteria

- [ ] A cognito session sends and receives turns with no row in `sessions` and none in `session_messages`
- [ ] A cognito session never appears in `GET /v1/sessions` or in the History page
- [ ] No title is generated and no post-turn skill review runs for a cognito session
- [ ] `write_file` and `edit_file` refuse `MEMORY.md` and `memory-archive/YYYY-MM.md` in both sub-modes, and the refusal names the reason
- [ ] `org_memory_*` and `skill_manage` are absent from a cognito session's tool list
- [ ] In the non-personalized sub-mode the system prompt contains no soul content, no skills catalog, no knowledge base catalog, no org memory section and no user context
- [ ] In the non-personalized sub-mode no plugin, MCP, Composio, skill or sub-agent tool is offered
- [ ] The personalized sub-mode still resolves the same prompt layers and tools as a normal session
- [ ] Attachments uploaded in a cognito session are removed from disk and from `attachments` when the session ends, and a server restart sweeps any that survived
- [ ] Toggling cognito off deletes the session server side and returns to the previous normal session
- [ ] Changing provider settings mid-chat does not kill a live cognito session
- [ ] A normal session is unchanged on every one of these paths

## Open questions

1. @ahmadrosid Web only for the first pass, or should the messaging channels get a `/cognito` command in the same change? The server side is channel agnostic either way, so this is only about how much UI lands at once.
2. @ahmadrosid Should a cognito chat also be stopped from writing artifacts into the profile workspace? I lean no, because that write is something the user explicitly asked for rather than a trace the session leaves behind, but it does mean a cognito chat can leave a file on disk.

Happy to implement this once the scope question is settled.

yusupsupriyadi:
## Summary

Toggle cognito in the web chat and the conversation lives only in server memory: nothing it produces is written back. It still uses profile memory, org memory, plugins, skills and the agent's own instructions. It just leaves nothing behind.

Fixes #1024 (part 1 of 2).

**Before:** every web chat writes a `sessions` row before the first token, replays into `session_messages` each turn, then earns a generated title and a post-turn skill review. Deleting it afterwards is too late.
**After:** `cognito` on `CreateSessionRequest` skips the row entirely and `buildChatSession` returns the session unwrapped, so no turn can reach `session_messages`.

Scope follows @ahmadrosid's call in #1025: web only, explicitly requested workspace files still persist, and personalized cognito ships with the attachment lifecycle in this PR. The non-personalized sub-mode is out, and lands next.

Why safe:
1. `cognito` is optional and absent means the old path, so every existing caller, channel and generated client is byte-identical.
2. The cognito branches sit at `getSessionRecordForOrg` and `resolveSession`, the two functions every session path already funnels through, so no path is half-converted.
3. Write-back is closed at each source rather than per caller: `forbidMemoryWrites` on the tool context, `propose_org_memory` and `skill_manage` filtered out, both post-turn schedulers guarded inside `AgentService`, and `profile_skill_usage` recording gated.

Only re-add persistence if losing a cognito chat on reload turns out to be the wrong trade. It is deliberate here, and stated in the UI: no row means no half-saved transcript to leak.

Residual, and deliberate: `llm_usage_stats` and `llm_turn_usage` still count a cognito turn. They are aggregate counters keyed by org, model and bucket with no session or content, and quota enforcement needs them. Files the agent writes to the profile workspace on request also persist, which is the answer to the second question on #1024.

Related, not fixed: #373 and #368, which cover retention and erasure for data that *is* stored.

## Not in this PR

The non-personalized sub-mode: the neutral prompt, the tool filtering, and the dropdown and welcome variants. `cognito` is a flag here rather than `{ personalized: boolean }`, because an object would accept `personalized: false` and silently do nothing. The follow-up widens the field when the sub-mode arrives, and `refactor(cognito): defer the non-personalized sub-mode` is the commit it reverts.

## Test plan
- [x] `bun run test`: server 1414 pass, core 852, web 573, db 116. The 56 server and 2 core failures are the `origin/main` baseline (plugin and provider tests need network), measured on `dcd21605`: 1391/56 and 849/2. Same 58 failures, name for name. 31 tests added.
- [x] `bun run check` clean (1401 files), `bun run knip` clean
- [x] Rebased on `dcd21605`. The `use-chat-page.ts` conflict is resolved by leaving the history-list invalidation to ordinary sessions: a cognito session is not in `sessions`, so there is nothing to refetch.
- [x] Revert experiment, production reset to `dcd21605` with the tests kept: 15 of the 20 server cognito cases go red, the memory-write guard goes red in core, the `ephemeral` column case goes red in db, and the web hook case goes red. The 5 that stay green are the controls, and are supposed to be: `an ordinary session still persists`, `an ordinary session keeps both write paths`, `an ordinary turn still records it`, and the two cases pinning that a cognito session reads every layer an ordinary one reads.
- [x] Adversarial cases pinned, not just the happy path: an ordinary session still gets `skill_manage`, `propose_org_memory`, a title, a skill review and a usage row; `notes/MEMORY.md` one directory down stays writable while `MEMORY.md` and `memory-archive/2026-09.md` are refused
- [ ] In the browser: turn cognito on, send a message, reload, the chat is gone and History never showed it; turn it off and the previous chat is back


fajarhide (commented):
Keeping cognito out of SQLite by never writing the session row is the right shape. The tests and the second sub-mode can shrink a lot.

I checked out the branch at b4dad0b2 (the two commits since are web only), ran the new tests, then reverted the production files to the merge base and kept the tests: 12 of the new cases stay green. Not checked: a live browser run, branchSession and compaction on a cognito session.

apps/server/src/http/routes/sessions.cognito.test.ts
```diff
@@ -0,0 +1,200 @@
...
+    const { sessions } = (await response.json()) as {
+      sessions: Array<{ id: string }>;
+    };
+    const ids = sessions.map((entry) => entry.id);
+
+    expect(ids).toContain(normalId);
+    expect(ids).not.toContain(cognitoId);
+  });
+
+  test("the turn is readable over HTTP while the session is alive", async () => {
```
fajarhide: With the production code reverted this still passes, since an ordinary session is readable while alive too. The service test covers the same path, so you could drop this and keep the same confidence.

apps/server/src/services/agent-service.cognito.test.ts
```diff
@@ -0,0 +1,608 @@
...
+    );
+    const session = await service.resolveSession(sessionId, ORG_ID);
+    await session?.send({ message: "hello" });
+
+    expect(await db.getSession(sessionId)).not.toBeNull();
+    expect((await db.listMessagesForSession(sessionId)).length).toBe(2);
+    expect((await db.listSessions()).length).toBe(1);
+  });
+
+  test("resolveSession returns the same object so history is never rebuilt", async () => {
```
fajarhide: This one, the delete-makes-it-unreachable case and the readable-while-alive case all stay green with production reverted to the merge base. Could they collapse into one case that fails without cognito?

apps/server/src/services/ephemeral-session-store.ts
```diff
@@ -0,0 +1,103 @@
...
+      return false;
+    }
+
+    this.entries.delete(sessionId);
+    await this.onEvict(entry);
+    this.stopSweepWhenEmpty();
+    return true;
+  }
+
+  /** Exposed for tests; the timer calls it with the current clock. */
```
fajarhide: Says exposed for tests, but no test calls `evictIdle`, so the 12h eviction never runs under test. A Map on AgentService plus one unref'd interval would do the same with fewer moving parts.

packages/core/src/tools/builtin.ts
```diff
@@ -241,15 +279,24 @@ export function refuseSkillLocalToolFileWrite(resolvedPath: string): void {
   );
 }
 
-function buildFileGuardOptions(
+function resolveWorkspaceRoot(
```
fajarhide: `tools/paths.ts:130` already exports an async `resolveWorkspaceRoot` with a different signature, and `bash.ts:464` has its own. A third one here is easy to import by mistake. Could this get a distinct name?

packages/db/src/migrate.ts
```diff
@@ -1429,12 +1429,26 @@ function migrateAttachmentsTable(db: Database): void {
...
       storage_path TEXT NOT NULL,
       created_at TEXT NOT NULL,
+      ephemeral INTEGER DEFAULT 0 NOT NULL,
       FOREIGN KEY (org_id) REFERENCES organizations (id) ON DELETE CASCADE,
       FOREIGN KEY (profile_id) REFERENCES profiles (id) ON DELETE CASCADE,
       FOREIGN KEY (session_id) REFERENCES sessions (id) ON DELETE SET NULL
     );
   `);
 
+  const columns = db.prepare("PRAGMA table_info(attachments)").all() as Array<{
```
fajarhide, line 1439: #1024 says pure in-memory, nothing written and then cleaned up. A new column plus a startup sweep is that. Question: could cognito attachments skip the column, or is there a case a `session_id IS NULL` check misses?

apps/server/src/services/agent-service.ts
```diff
@@ -332,6 +341,25 @@ import {
   type ServerToolOverrides,
 } from "./tool-resolver";
 
+/**
+ * The whole prompt for a non-personalized cognito chat. Deliberately bare: the
+ * point of the mode is to see the model without the soul, skills, memory or
+ * profile instructions layered on top. chat-prompt.ts still adds the tool
+ * instructions and the channel style on top of this.
+ */
+const NON_PERSONALIZED_SYSTEM_PROMPT = [
```
fajarhide, line 350: Question on size: would this land faster as personalized-only first, with this prompt, the tool filtering and the second welcome variant as a follow-up PR? That is roughly 250 lines out of this one.

fajarhide:
@yusupsupriyadi The feature is well argued in #1024, and most of this diff follows its touch points closely. At +1912 across 16 commits it is a lot to review in one go, so a suggested split along the acceptance criteria:

1. **Personalized cognito.** The contract field, the in-memory session (no rows, not listed, no title or skill review), the memory write guard, dropping `org_memory_*` and `skill_manage`, surviving provider changes, and the toggle.
2. **Non-personalized sub-mode.** The neutral prompt, the tool filtering, and the dropdown and welcome variants.
3. **Attachment lifecycle.** The `ephemeral` column, the purge and the restart sweep. Until it lands, the composer could hide attachments in a cognito chat so nothing is left behind.

Each one is small enough to test against its own criteria, and the test cuts from my review get easier per piece.

A correction to my inline note on `migrate.ts`: #1024 does ask for the `ephemeral` attachments column. The "nothing written and then cleaned up" line is about `sessions`, so ignore that framing.

@ahmadrosid the two open questions on #1024 are still unanswered. Does this split and order work for you before more lands here?

yusupsupriyadi:
@fajarhide The revert experiment was the right way to read these tests, and it found a real problem. Rebased on `bcd791bb` and pushed.

**The tests.** I reran your experiment on the current branch: 10 of the cognito cases passed against an untouched main, not 12, and the split matters.

Five were weak, exactly as you said. An ordinary session is also readable while alive, also resolves to the same cached object, and also stops resolving once deleted, so those assertions proved nothing. They now assert the cognito half alongside, that no row and no messages exist, which an ordinary session cannot satisfy. Three collapse into one case covering readability, the stable object and the absent rows together.

The other five are deliberate. `an ordinary session still persists`, `an ordinary session keeps both write paths`, `an ordinary turn still records it`, and the two `the personalized sub-mode keeps...` cases are the control arm: they assert the paths this PR touches are unchanged for everyone else, so they are supposed to be green on main. If they ever go red, the feature has leaked. Re-measured after the fix: 10 green before, 5 after, and those 5 are the controls.

**`evictIdle`.** Fair at the commit you read. The timer is gone since then, along with the start/stop lifecycle around it. Eviction now rides on the next cognito session, the ceiling is marked in the code, and three tests cover the 12-hour threshold, the collect-on-next-session trigger, and delete purging exactly once.

**`resolveWorkspaceRoot`.** Renamed to `fileToolWorkspaceRoot`, with a comment naming the two neighbours so the next person does not re-add the collision.

**The `ephemeral` column.** Your question stands even after your correction, so here is the measurement. On my install 0 of 36 attachment rows have a null `session_id`, so `session_id IS NULL` would work today. I would still keep the column, because `attachments.session_id` is declared `ON DELETE SET NULL`: the schema is built to produce exactly that state for ordinary attachments. Nothing reaches it today since `purgeSession` deletes them before the row, but that is an invariant held by one call site, not by the schema, and the failure mode if it ever breaks is a startup sweep deleting someone's real files. One integer to make the sweep say what it means seemed the better side to be wrong on.

**The split.** Three PRs along the acceptance criteria is a better shape than this one, and your ordering is the right one. Two things before I cut it up: the `ephemeral` column and the purge are what make attachments safe in a cognito chat, and hiding the composer's attachment button until part 3 lands means part 1 ships a mode that silently drops a capability, so I would rather keep part 3 with part 1 or hide the button as you suggest, whichever you prefer. Second, it is already written and green, so splitting is re-sequencing, not rework, and I would rather do it once you and @ahmadrosid agree the order, than guess and redo it.

@ahmadrosid the two questions on #1024 are still open, and the second one decides part 3's scope: should a cognito chat also be stopped from writing artifacts into the profile workspace?

apps/server/src/http/routes/sessions.cognito.test.ts
yusupsupriyadi: @fajarhide Right, and it was worse than one case: five of them passed against an untouched main. Dropped this one and folded its assertion into `a full turn leaves nothing in the database`, which already fails without cognito. Now one case covers the turn being readable over HTTP and nothing reaching the tables. e6233990

apps/server/src/services/agent-service.cognito.test.ts
yusupsupriyadi: @fajarhide Collapsed into one, `the turn is readable, and exists nowhere but memory`: it sends a turn, reads it back, resolves again to check it is the same object, then asserts no session row, no messages and nothing in `listSessions`. Those last three are what an ordinary session cannot satisfy, so it now fails without cognito. `deleting a cognito session makes it unreachable` kept its own case but asserts the row was already absent before the delete, which makes it cognito-specific too. e6233990

apps/server/src/services/ephemeral-session-store.ts
yusupsupriyadi: @fajarhide Fair at the commit you read. The timer is gone since then, and so is the lazy start/stop lifecycle around it: eviction now rides on the next cognito session, so an abandoned entry survives until someone opens another cognito chat or until a restart, whose attachment sweep collects what it left. The ceiling is marked with a `ponytail:` comment rather than left implicit. `evictIdle` kept its injectable clock because there are now three tests on it: the 12-hour threshold, the collect-on-next-session trigger, and delete running its purge exactly once. 4620cc72

On the alternative: a plain Map on `AgentService` is roughly what this is, but the purge callback has to reach `db`, `deleteAttachmentBytes` and three session-state objects, so inlining it puts that back into a 4000-line file. Happy to inline it if you would rather see it there.

packages/core/src/tools/builtin.ts
yusupsupriyadi: @fajarhide Good catch, three of them and all with different signatures. Renamed this one to `fileToolWorkspaceRoot`, with a comment naming the other two so the next person does not re-add the collision. 3c596a6f

packages/db/src/migrate.ts
yusupsupriyadi: @fajarhide Taking your correction, but the underlying question is still worth answering, so I measured it. On my install 0 of 36 attachment rows have a null `session_id`, so a `session_id IS NULL` sweep would be correct today.

I would still keep the column. `attachments.session_id` is declared `ON DELETE SET NULL`, so the schema is built to produce exactly that state for ordinary attachments; nothing reaches it only because `purgeSession` deletes them before the session row. That is an invariant held by one call site, not by the schema, and if it ever breaks the failure is a startup sweep deleting a user's real files. One integer to make the sweep say what it means seemed the better side to be wrong on.

apps/server/src/services/agent-service.ts
yusupsupriyadi: @fajarhide Answered on the conversation tab rather than here since it decides the shape of the whole PR: your split and its ordering are better than this one, and the only thing I would change is where the attachment lifecycle sits. Waiting on you and @ahmadrosid to agree the order before I cut it, since it is re-sequencing work that is already written and green, and I would rather not guess and redo it.

ahmadrosid:
Web only for now. Allow explicitly requested workspace files to persist. Keep personalized mode and attachment cleanup together in the first PR, then non-personalized mode separately.

fajarhide:
@yusupsupriyadi I re-ran the revert experiment on `e6233990`, one file at a time. With production reset to `bcd791bb`, only the five controls you listed stay green, and every cognito case goes red. The `ephemeral` column holds up too, since `attachments.session_id` is `ON DELETE SET NULL`. Nothing is left from my inline notes, so I'm resolving those threads.

For the re-cut: `apps/web/src/pages/chat/use-chat-page.ts` conflicts with main now, which has moved 9 commits since `bcd791bb`.

yusupsupriyadi:
@ahmadrosid All three answered, and the PR is re-cut to part 1. Rebased on `dcd21605` and pushed.

**Web only.** Nothing changes: the toggle is a React component and the server field is channel agnostic, so the messaging channels get no `/cognito` and no UI. They can adopt it later without touching the server.

**Workspace files persist.** Already the behaviour, now stated rather than left as my lean. A cognito chat that is asked for a document still writes it under `~/.nakama`, because that is an action you asked for rather than a trace the session left.

**Personalized plus attachment cleanup in this PR, non-personalized next.** Out of this branch: the neutral system prompt, the `personalized` gate through `resolveProfileTools` and `resolveProfileSystemPrompt`, the sub-mode entry on the in-memory session, and the dropdown next to the toggle. It is one commit, `refactor(cognito): defer the non-personalized sub-mode`, so part 2 is a revert of it rather than a rewrite.

One shape decision that goes with the split: `cognito` is now a flag, not `{ personalized: boolean }`. Kept as an object this PR would accept `personalized: false` and silently do nothing, which is a worse API than not offering the choice yet. Part 2 widens the field when the sub-mode arrives.

The two sub-mode tests went with it. In their place is a control describe pinning that a cognito session reads every layer an ordinary one reads, the profile's own prompt, the skills catalog and the org-taught tools, which is the whole of what this PR now claims.

I force-pushed over `4206b71e`. Nothing of yours is lost: that merge was main plus the cognito commits with no original work, and it was already four commits behind. A rebase onto current main serves the same purpose and keeps the history granular for the re-cut.

@fajarhide Thanks for running the revert one file at a time, that is the read that made the test split worth doing. Re-measured on the re-cut branch, production reset to `dcd21605` with the tests kept: 15 of the 20 server cognito cases go red, the memory-write guard goes red in core, the `ephemeral` column case goes red in db, and the web hook case goes red while `chat-page.shared.test.ts` no longer loads at all. Five stay green and all five are controls, the three you named plus the two new ones above.

On `use-chat-page.ts`: the conflict is main's new `invalidateQueries` on the sessions list landing next to the cognito branch. It is gated with `syncChatUrl` now, since a cognito session is never in `sessions` and there is nothing to refetch. Say the word if you would rather it stay unconditional.

Full suite against `dcd21605`: server 1414 pass, core 852, web 573, db 116, and the 58 failures are main's own, name for name.

ahmadrosid:
LGTM

PR #1027, opened 2026-09-14
fix(server): accept the loopback origin after a desktop relaunch

### LINKED ISSUE #1026: Desktop: chat returns "Origin is not allowed." after the app is relaunched
### Summary

After the desktop app is relaunched, sending a chat message fails with `400 {"error":"Origin is not allowed."}`. So does creating an artifact share link. Chat is dead from the second launch on, for every desktop install.

Image

The desktop app starts the server with `NAKAMA_PORT=0` (`apps/desktop/main.mjs`), so each launch binds a new port. The setup wizard sends `window.location.origin` as `webPublicUrl` (`apps/web/src/context/auth-context.tsx:84`), and `/v1/auth/setup` saves it as `web_public_url`. That saves `http://127.0.0.1:<first port>`.

Once a public URL is saved, `isAllowedClientOrigin` (`apps/server/src/services/composio-callback-url.ts:72`) only accepts that exact host and port. The relaunched window's origin carries the new port, so every route that runs `resolveRequestClientOrigin` refuses it: `POST /v1/sessions/:id/messages`, artifact shares, Composio connect.

### Measured on

9605ae7e

### Reproduction

Two server runs on one config dir, the way desktop launches it:

```sh
D=$(mktemp -d)
NAKAMA_CONFIG_DIR=$D NAKAMA_HOST=127.0.0.1 NAKAMA_PORT=4391 bun apps/server/src/index.ts
# POST /v1/auth/setup with Origin and webPublicUrl http://127.0.0.1:4391, then stop it
NAKAMA_CONFIG_DIR=$D NAKAMA_HOST=127.0.0.1 NAKAMA_PORT=4392 bun apps/server/src/index.ts
# log in, create a session, POST /v1/sessions/:id/messages with Origin http://127.0.0.1:4392
```

### Expected vs actual

Expected: the message is accepted.

Actual:

```
== send a chat message from the relaunched window
{"error":"Origin is not allowed."}
HTTP 400
```

A saved loopback URL names a port, not a public host, and on desktop that port changes on every launch. Proposed fix, in `composio-callback-url.ts` only:

- when the saved URL is loopback, accept any loopback origin
- when the saved URL is loopback and the browser origin passed the check, build callback and share links from the browser origin, so they do not point at the dead port

A non-loopback `webPublicUrl` stays the whole allowlist, so the PR 821 policy for deployed installs does not change.

fajarhide:
## Summary

Relaunched desktop app can send chat messages again. Closes #1026.

**Before:** setup saves `http://127.0.0.1:<first port>` as `webPublicUrl`, desktop binds a new port on every launch, and every send returns `400 Origin is not allowed.`
**After:** a saved loopback URL accepts any loopback origin, and callback and share links use the verified browser origin instead of the dead port.

Why safe:
1. A non-loopback `webPublicUrl` is still the whole allowlist, so deployed installs keep the PR 821 policy.
2. A loopback URL now accepts loopback origins only, which is what an install with no saved URL already accepts.
3. The new test also asserts `https://evil.example.com` is still refused.

Only re-add a port check if a loopback install ever needs to tell two local ports apart.

## Test plan
- [x] Live, two server runs on one config dir (4391 for setup, 4392 after relaunch), `POST /v1/sessions/:id/messages`: `HTTP 400 {"error":"Origin is not allowed."}` on 9605ae7e, `HTTP 200` with this branch. Same env as desktop spawns; not clicked through the Electron window.
- [x] New test fails with the fix reverted (`9 pass, 1 fail`), passes with it.
- [x] `bun run test`: every workspace 0 fail except `@nakama/server` 1366 pass / 54 fail, all plugin suites with `package_unavailable` (no npm in this sandbox, same 54 on upstream `42160af4`).


ahmadrosid:
Nice catch.

PR #1037, opened 2026-09-14
test(desktop): send a message after the runtime relaunch

### LINKED ISSUE #1036: Desktop runtime test never sends a message after the relaunch
### Summary

`bun run --cwd apps/desktop test:runtime` restarts the bundled server and logs in again, but it never sends a message after the restart. So #1026 (every chat send answered `400 Origin is not allowed.` from the second launch on) passed it.

The test's setup also sends no `webPublicUrl`, while the web wizard always sends `window.location.origin`. That saved origin, plus a new port on the next launch (`NAKAMA_PORT=0` in `apps/desktop/main.mjs`), is what broke.

### Why it matters

Desktop is the one install that changes its port on every start. Anything persisted with the port in it is stale on the next launch, and the only automated desktop check does not exercise it.

`apps/desktop/tests/smoke.mjs`, the `--runtime-test` block.

### Acceptance criteria

- [ ] setup sends `webPublicUrl` set to the first launch's URL, like the wizard
- [ ] after the restart, the test creates a session and sends a message with `Origin` set to the new URL, and asserts it is not refused
- [ ] the test fails against a runtime carrying the resolver from before #1027, and passes on main

fajarhide:
## Summary

The desktop runtime test now sends a chat message after the relaunch, so #1026 cannot come back unnoticed. Closes #1036.

https://github.com/user-attachments/assets/ba788401-c16a-4a43-ab6c-72e0cc237f06

**Before:** `test:runtime` restarted the server and logged in, with no `webPublicUrl` at setup and no send, and passed with the #1026 bug present.
**After:** setup sends `webPublicUrl` like the wizard, and after the restart the test creates a session and sends a message with the new `Origin`.

Why safe:
1. Test-only change in `apps/desktop/tests/smoke.mjs`.
2. No provider needed: the send answers from offline mode.

## Test plan
- [x] `bun run --cwd apps/desktop test:runtime` on main: `Passed: bundled runtime, first setup, persistent account, chat after relaunch, shutdown, and parent disconnect cleanup.`
- [x] Same test with `dist/runtime` carrying the resolver from `9605ae7e` (before #1027): `AssertionError: {"error":"Origin is not allowed."} actual: 400, expected: 200`.
- [x] `bun run --cwd apps/desktop test`: 3 pass, smoke passed. The full `bun run test` was not rerun, since only this test file changed.


yusupsupriyadi (approved):
@fajarhide LGTM.

Reproduced both directions here on Linux, driving the runtime block against a stubbed runtime instead of the macOS bundle:

- Server built from a tree without #1027: `AssertionError: {"error":"Origin is not allowed."} actual: 400, expected: 200`.
- Same run with the #1027 resolver in place: passes, and the relaunch really did take a fresh port (setup `:45537`, relaunch `:41999`), so the origin check is what the assert is exercising.

Sending `webPublicUrl` at setup is the part that makes this bite: without it nothing is saved, the allowlist falls back to the request origin, and the test can never fail.

PR #1041, opened 2026-09-14
feat: add organization-scoped shared knowledge base

### LINKED ISSUE #983: knowledge base: locked to one profile, no org scope and no configurable sources
### Summary

The knowledge base is scoped to a single profile with no way to share it. `getKnowledgeBaseDir(orgId, profileId)` puts every document under that one profile's soul dir, and all four endpoints live under `/v1/profiles/:profileId/knowledge-base`. Giving five agents the same reference set means five uploads, five extractions, five manifest entries and five copies of the same bytes.

### Why it matters

Document sets are usually a property of the work, not of one agent. An org that runs a researcher, a writer and a reviewer over the same corpus has to upload it three times and re-upload three times whenever a document is revised. Nothing detects that the copies are the same file even though the manifest already stores a SHA-256 per document.

There is no escape hatch through URL sources either. `KnowledgeBaseSource` looks configurable, but `DEFAULT_KNOWLEDGE_SOURCES` in `knowledge-base/sources.ts:6` is a one-element constant holding the Nakama docs index, and both call sites assign it directly with no merge:

```
apps/server/src/services/profile-service.ts:831   const sources = DEFAULT_KNOWLEDGE_SOURCES;
packages/core/src/knowledge-base/catalog.ts:9     const sources = DEFAULT_KNOWLEDGE_SOURCES;
```

So `enabled` and `inherited` on that type cannot be set by anyone, and no endpoint adds a source.

- `packages/core/src/knowledge-base/paths.ts:8`, and `:17`, `:31`, `:42` which all derive from it
- `apps/server/src/http/routes/profiles.ts:861-904`, the four knowledge base endpoints
- `packages/core/src/tools/knowledge-base-search.ts`, which resolves its root from `getProfileSoulDir`
- `apps/web/src/components/soul-tools/KnowledgeTab.tsx`

### Acceptance criteria

- [ ] a document can be uploaded once at org scope and read by more than one profile
- [ ] `knowledge_base_search` searches org-scoped documents alongside profile-scoped ones, and a match says which scope it came from
- [ ] deleting an org-scoped document is refused or clearly reported while profiles still reference it
- [ ] existing profile-scoped documents keep working untouched, with no migration required of the operator
- [ ] the knowledge tab shows which documents are shared and which belong to the profile
- [ ] a test covers a search that matches one document in each scope

puspoaditya:
## Summary

Platform admins upload a shared document once → it is searchable from every profile it is attached to, including with Supermemory on, and the upload route answers with real status codes.

**Before:** the org upload route called core directly (duplicate → 500, bad body → 500, attached replace → 500) and an attached org document vanished from normal queries whenever the memory backend answered.
**After:** the four org endpoints go through `ProfileService` (409 / 400 / 404 like the profile route) and `runKnowledgeBaseSearch` merges a ripgrep pass over the organization root into backend results.

Why safe:
1. Store paths now take an explicit root (`getOrgKnowledgeBaseDir(orgId)`), so `profileId: undefined` can no longer silently mean "organization scope"
2. Delete-in-use still returns 409 with `profileIds`, and the guard now also covers replace uploads
3. Citations of attached org documents resolve in `propose_org_memory` instead of being dropped

Scope per review:
- Management (list / upload / delete / content / attach / detach) stays platform-admin only. All six routes call `requirePlatformAdminFromContext`, and a test sends an org-admin session through all six and expects `403`.
- The six routes are registered in the OpenAPI document, with the `409` + `profileIds` body documented on delete.
- Dashboard org upload/delete is a follow-up PR; `KnowledgeTab` keeps list plus attach/detach.

Only re-add a fuller backend sync if semantic recall (not ripgrep) is required for org documents.

## Test plan
- [x] `bun run check`
- [x] `bun test packages/core/src/knowledge-base packages/core/src/tools/knowledge-base-search.test.ts` (20 pass)
- [x] `bun test apps/server/src/tools/org-memory-tools.test.ts apps/server/src/services/profile-service.test.ts` (54 pass)
- [x] `bun test apps/server/src/http/routes/profiles-organization-knowledge-base.test.ts` (6 pass)
- [x] `bun run --filter @nakama/core build`, `@nakama/client`, `@nakama/server`, `@nakama/web`
- [ ] Manual: attach a shared document, search a phrase that only exists there, confirm the hit is labelled `organization`

Closes #983


fajarhide (commented):
Attaching by reference instead of copying bytes is the right shape, and both new tests fail with production reverted. Two paths still miss org documents, and the org upload route returns 500 where the profile route returns 409/400.

apps/server/src/http/routes/profiles.ts
```diff
@@ -929,6 +949,135 @@ export function registerProfileRoutes(
...
+  });
+
+  app.post("/v1/orgs/:orgId/knowledge-base", async (c) => {
+    requirePlatformAdminFromContext(c);
+    const orgId = requireActiveOrgIdFromContext(c);
+    if (orgId !== decodeURIComponent(c.req.param("orgId"))) {
+      throw new NakamaApiError("Not found", 404);
+    }
+    const body = await readJson<UploadKnowledgeBaseRequest>(c.req.raw);
+    const result = await uploadOrganizationKnowledgeBaseDocument(
```
fajarhide: This calls core directly, so it skips the error mapping `ProfileService.uploadKnowledgeBaseDocument` does. I ran the same three bodies against both routes as a platform admin:

```
/v1/profiles/profile_a/knowledge-base  dup 409  unsupported 400  missing body 400
/v1/orgs/org_a/knowledge-base          dup 500  unsupported 500  missing body 500
```

A duplicate org upload never gets the skip/replace prompt, and each of these lands in error tracking as a 500. Routing it through a service method with the same try/catch fixes all three.

packages/core/src/tools/knowledge-base-search.ts
```diff
@@ -72,7 +77,6 @@ export async function runKnowledgeBaseSearch(
   }
 
   const parsed = parseToolInput(knowledgeBaseSearchInputSchema, input);
-
   const backend = await context.searchKnowledge?.({
```
fajarhide, line 81: With Supermemory enabled this returns before the org scope is searched, and `knowledgeEntries` only lists profile documents. I attached `shared.txt` in your test and added a backend:

```
PROBE backend on: []
PROBE backend regex: 1 organization
```

So an attached org document is invisible to a normal query and only shows up with `regex: true`. Cheapest fix is to still run the org ripgrep target when the backend answers and merge the matches. The fuller one is syncing attached org documents into the backend on attach, detach and upload.

apps/web/src/components/soul-tools/KnowledgeTab.tsx
```diff
@@ -189,6 +226,59 @@ export function KnowledgeTab({ profileId }: { profileId: string | null }) {
...
         ) : null}
 
+        <section className="mb-4 rounded-md border border-border">
+          <div className="border-border border-b px-4 py-3">
+            <h3 className="font-medium text-sm">
+              Shared organization knowledge
+            </h3>
+            <p className="mt-1 text-muted-foreground text-xs">
+              Attach an organization document to make it available to this
+              profile. Shared documents are managed by organization
```
fajarhide, line 236: Nothing in `apps/web` calls `uploadOrganizationKnowledgeBaseDocument` or `deleteOrganizationKnowledgeBaseDocument`, so in the dashboard this section always shows "No shared organization documents yet." and org documents can only be added through the API. Is a management UI planned for this PR or a follow-up?

packages/core/src/knowledge-base/paths.ts
```diff
@@ -1,17 +1,28 @@
...
-export function getKnowledgeBaseDir(orgId: string, profileId: string): string {
-  return join(getProfileSoulDir(orgId, profileId), KNOWLEDGE_BASE_RELATIVE_DIR);
+/** Root for documents shared by profiles in an organization. */
+export function getOrgKnowledgeBaseDir(orgId: string): string {
+  return join(getOrgConfigDir(orgId), KNOWLEDGE_BASE_RELATIVE_DIR);
+}
+
+export function getKnowledgeBaseDir(
+  orgId: string,
+  profileId: string | undefined
```
fajarhide: Before this change, a missing `profileId` was a type error. Now `undefined` means org scope, so a caller holding a nullable profile id reads the org root without any warning. `getOrgKnowledgeBaseDir` already exists. Could the store functions take that root explicitly and keep `profileId: string`?

packages/core/src/tools/knowledge-base-search.test.ts
```diff
@@ -105,6 +105,76 @@ describe("knowledge_base_search tool", () => {
     expect(result.root).toBe(getKnowledgeBaseDir(orgId, profileId));
   });
 
+  test("includes only shared organization documents attached to the profile", async () => {
```
fajarhide, line 171: This query only matches the org document, so the issue's "one document in each scope" case is still untested. A query that hits `private alpha` and `shared alpha` together would cover it, including the merge and the scope labels.

fajarhide:
The branch conflicts with main in `catalog.ts`. #984 (505e3598) added `[id: ...]` to the same catalog line, so that part is quick to resolve.

After the rebase there is one more thing to check. `resolveSourceDocumentIds` in `apps/server/src/tools/org-memory-tools.ts` looks ids up with `listKnowledgeBaseDocuments(orgId, profileId)` and drops anything it doesn't find. If an agent cites an attached org document in `propose_org_memory`, that id will be dropped silently. I found this by reading the code and haven't run it, since it only shows up once both changes are on one branch.

What I ran: the three test files from your description (54 pass), the six new routes through a platform-admin session, and tsc on every touched file, which is clean. I didn't try the knowledge tab in a browser.

puspoaditya:
Thanks — all five points are addressed on the rebased branch (`77d3d4d6` on top of `01a512ff`).

**Conflict (`catalog.ts`)** — rebased, kept both: `- [${scope}] ${filename} (${mediaType}) [id: ${id}]`.

**500s on the org upload route** — the four org endpoints now go through `ProfileService` instead of calling core directly, so they map like the profile route: duplicate `409`, unsupported/missing body `400`, unknown document `404`, in-use `409` with `profileIds`. Replace-on-attached also maps to `409` now (it used to surface as a 500 through the duplicate path). New `apps/server/src/http/routes/profiles-organization-knowledge-base.test.ts` asserts those statuses through a platform-admin session, including the `profileIds` body on delete.

**Supermemory** — took your cheaper suggestion: the organization target is resolved before the backend call and its ripgrep pass is merged into the backend matches, with hits labelled `organization`. Your probe is now a test — `keeps attached organization documents searchable when the memory backend answers` returns the profile hit plus the attached org hit, and the unattached token stays at 0. I did not sync attached org documents into the backend; say the word if you want that instead and I will do it as a follow-up.

**`resolveSourceDocumentIds`** — resolves attached org documents too (profile scope wins a filename collision), so citing one in `propose_org_memory` no longer drops it. Test: `keeps attached organization documents when the agent cites them`.

**`paths.ts`** — helpers take the root explicitly (`getOrgKnowledgeBaseDir(orgId)`, `getKnowledgeBaseManifestPath(dir)`), store internals take `dir`, and `profileId: string` is back on every profile-scoped function.

**Test gap** — `merges profile and organization matches in one query` covers the issue's "one document in each scope" case, including the merge order and both scope labels.

On the dashboard question: `KnowledgeTab` already lists org documents and attaches/detaches them, but there is still no way to upload or delete an org document from the UI — only the API. Happy to add the upload control here if you want it in this PR rather than a follow-up.

Verified locally: `bun run check`, core KB + search tests (20 pass), `org-memory-tools` + `profile-service` (54 pass), new route test (4 pass), and builds for core/client/server/web.

yusupsupriyadi (commented):
@puspoaditya I went through the whole diff at `1f22b5b`. The scope split between the profile manifest and the org manifest is clean, and routing removal through `KnowledgeBaseDocumentInUseError` is the right call. One blocking item and a set of smaller ones inline.

**Blocking:** the four `/v1/orgs/:orgId/knowledge-base*` handlers call `agent.listOrganizationKnowledgeBase` and friends, but those methods are on `ProfileService`, not on `AgentService`. I checked the branch out and typechecked it:

```
apps/server/src/http/routes/profiles.ts(975,29): error TS2339: Property 'listOrganizationKnowledgeBase' does not exist on type 'AgentService'.
apps/server/src/http/routes/profiles.ts(985,32): error TS2551: Property 'uploadOrganizationKnowledgeBaseDocument' does not exist on type 'AgentService'. Did you mean 'uploadKnowledgeBaseDocument'?
apps/server/src/http/routes/profiles.ts(1002,21): error TS2551: Property 'deleteOrganizationKnowledgeBaseDocument' does not exist on type 'AgentService'. Did you mean 'deleteKnowledgeBaseDocument'?
apps/server/src/http/routes/profiles.ts(1025,34): error TS2551: Property 'readOrganizationKnowledgeBaseDocument' does not exist on type 'AgentService'. Did you mean 'readKnowledgeBaseDocument'?
```

`grep -c Organization apps/server/src/services/agent-service.ts` returns 0, so on a real server all four endpoints still throw and answer 500, which is the failure this rebase set out to fix. The new route test passes because `createMinimalHonoApp` casts a hand-written stub with `overrides.agent as ServerOptions["agent"]`.

To be fair about the tooling: the root `tsc --noEmit` is nowhere near green on main either (around 10k errors, most of them `apps/web` hitting a root tsconfig with no `jsx` setting), so this is not something existing CI would have caught. `bun run knip` on the branch is clean.

Everything else is inline, roughly in order of how much it matters.

apps/server/src/http/routes/profiles.ts
```diff
@@ -949,6 +966,121 @@ export function registerProfileRoutes(
     }
   );
 
+  app.get("/v1/orgs/:orgId/knowledge-base", async (c) => {
+    requirePlatformAdminFromContext(c);
+    const orgId = requireActiveOrgIdFromContext(c);
+    if (orgId !== decodeURIComponent(c.req.param("orgId"))) {
+      throw new NakamaApiError("Not found", 404);
+    }
+    return json(await agent.listOrganizationKnowledgeBase(orgId));
```
yusupsupriyadi, line 1189: `agent.listOrganizationKnowledgeBase` does not exist. Same for `uploadOrganizationKnowledgeBaseDocument` (985), `deleteOrganizationKnowledgeBaseDocument` (1002) and `readOrganizationKnowledgeBaseDocument` (1025). All four are defined on `ProfileService` (lines 925, 937, 975, 991); `agent-service.ts` has no `Organization` member at all.

Since `agent` here is the concrete `AgentService`, each of these throws `TypeError: agent.listOrganizationKnowledgeBase is not a function` and the route answers 500. The route test does not catch it because the stub it casts defines all four.

Four delegating methods on `AgentService`, next to `listKnowledgeBase` and `uploadKnowledgeBaseDocument`, cover all of them.

apps/server/src/http/routes/profiles.ts
```diff
@@ -949,6 +966,121 @@ export function registerProfileRoutes(
     }
   );
 
+  app.get("/v1/orgs/:orgId/knowledge-base", async (c) => {
+    requirePlatformAdminFromContext(c);
```
yusupsupriyadi, line 1184: These four routes require platform admin, so an org admin cannot upload, delete, or read their own organization's shared documents. Only the instance operator can populate an org knowledge base. The PR description says "Org admins upload a shared document once", which reads like org-level access was the intent.

Every other `/v1/orgs/:orgId/...` resource gates on the org role: `org-memory.ts`, `org-members.ts`, `org-curator.ts`, `skill-proposals.ts`. `requireOrgAdminOrPlatformAdminFromContext` is already imported in this file and used at 1112 and 1119.

If platform-admin-only is deliberate, worth saying so in the description so the UI work lands on the right assumption.

apps/server/src/http/routes/profiles.ts
```diff
@@ -949,6 +966,121 @@ export function registerProfileRoutes(
     }
   );
 
+  app.get("/v1/orgs/:orgId/knowledge-base", async (c) => {
```
yusupsupriyadi, line 1183: None of the six new routes call `app.openAPIRegistry.registerPath`, while the four profile knowledge-base routes above it do (542, 580, 606, 637).

The generated spec ends up missing the whole shared knowledge base API, including the 409 response carrying `profileIds` that every client has to handle.

apps/server/src/http/routes/profiles.ts
```diff
@@ -949,6 +966,121 @@ export function registerProfileRoutes(
...
+    requirePlatformAdminFromContext(c);
+    const orgId = requireActiveOrgIdFromContext(c);
+    if (orgId !== decodeURIComponent(c.req.param("orgId"))) {
```
yusupsupriyadi, line 1186: This check is repeated at 981, 996 and 1022, and it is byte for byte the body of `resolveOrgId` in `org-memory.ts:69`, `org-curator.ts:36`, `skill-proposals.ts:39` and `skill-suggestions.ts:36`.

Four more copies push against the 3% jscpd gate (`bun run dupes`), and the next change to how cross-org access is decided now has nine places to touch instead of five. Lifting the existing helper into `org-guards.ts` would cover all of them.

apps/server/src/http/routes/profiles.ts
```diff
@@ -949,6 +966,121 @@ export function registerProfileRoutes(
...
+      const orgId = requireActiveOrgIdFromContext(c);
+      const profileId = decodeURIComponent(c.req.param("profileId"));
+      const documentId = decodeURIComponent(c.req.param("documentId"));
+      await agent.getProfile(orgId, profileId);
+      try {
+        await attachSharedKnowledgeBaseDocument(orgId, profileId, documentId);
+      } catch (error) {
+        if (
+          error instanceof Error &&
+          error.message === "Shared knowledge base document not found."
```
yusupsupriyadi, line 1265: Matching on the message string makes this 404 depend on wording. `attachSharedKnowledgeBaseDocument` throws a bare `new Error("Shared knowledge base document not found.")` at `store.ts:542`, so adding a word there silently turns this into a 500 and no test fails.

This PR already introduced `KnowledgeBaseDocumentInUseError` for the delete path; the same treatment here would keep the two consistent.

packages/core/src/tools/knowledge-base-search.ts
```diff
@@ -72,99 +78,175 @@ export async function runKnowledgeBaseSearch(
...
+      glob: target.glob,
+      maxResults: parsed.maxResults,
+      query: parsed.query,
+      regex: parsed.regex,
+      searchRoot: target.root,
+    }),
+    {
+      maxResults: parsed.maxResults,
+      searchRoot: target.root,
+      workspaceRoot,
```
yusupsupriyadi: `workspaceRoot` here is the profile soul dir, but for the organization target `searchRoot` is the org knowledge base dir. `parseMatchLine` does `path.relative(workspaceRoot, absolutePath)` (`ripgrep.ts:243`), so every organization match comes back with a `file` that climbs out of the profile dir with `../`.

The agent then sees a path outside its own workspace: a follow-up `read_file` on it gets rejected by the path guard, and the `root` in the response points at the profile knowledge base dir, which that path is not under.

The new tests assert `.text` and `.scope` but never `.file`, so this is uncovered. Passing the org root as the relativization base for org matches would fix it.

packages/core/src/tools/knowledge-base-search.ts
```diff
@@ -72,99 +78,175 @@ export async function runKnowledgeBaseSearch(
...
+  root: string,
+  scope: KnowledgeBaseScope,
+  documents: { filename: string; id: string; status: string }[],
+  filename: string | null
+): SearchTarget {
   if (!filename) {
+    const ids = documents
+      .filter((document) => document.status === "ready")
+      .map((document) => document.id);
+    if (ids.length === 0) {
```
yusupsupriyadi, line 232: This replaced a directory glob over `*.extracted.txt` with an explicit list of manifest ids, which makes a damaged manifest fail silently instead of degrading.

`readManifestFrom` swallows a JSON parse error and returns `{documents: []}` (`store.ts:215-228`), so a manifest truncated by a crash mid-`writeManifestTo` gives `ids.length === 0`, this returns `missing`, and the search reports nothing found while the extracted files sit untouched on disk. Before this change ripgrep still found them.

Falling back to the directory glob when the manifest yields no ids would keep the old behaviour.

packages/core/src/tools/knowledge-base-search.ts
```diff
@@ -72,99 +78,175 @@ export async function runKnowledgeBaseSearch(
...
+  await ensureKnowledgeBaseDirs(orgId, profileId);
+  const profileResult = await runSearchTarget(
+    await resolveProfileSearchTarget(orgId, profileId, parsed.filename ?? null),
+    parsed,
+    workspaceRoot
+  );
+  const matches = [
+    ...profileResult.matches,
+    ...organizationResult.matches,
+  ].slice(0, parsed.maxResults);
```
yusupsupriyadi: Profile matches come first and the slice is applied to the concatenation, while each `runSearchTarget` call is given the full `parsed.maxResults`. So when the profile knowledge base alone returns `maxResults` hits (50 by default), every organization match is dropped.

For any reasonably common word the shared document becomes unreachable through `knowledge_base_search`, with `truncated: true` as the only hint. Same shape in the backend branch at 109.

Interleaving the two scopes, or splitting the budget between them, would keep at least some org hits visible.

packages/core/src/tools/knowledge-base-search.ts
```diff
@@ -72,99 +78,175 @@ export async function runKnowledgeBaseSearch(
...
+    matchCount: matches.length,
+    matches,
     query: parsed.query,
-    root: searchTarget.root,
-    truncated: searchResult.truncated,
+    root: getKnowledgeBaseDir(orgId, profileId),
+    truncated:
+      profileResult.truncated ||
+      organizationResult.truncated ||
+      matches.length >= parsed.maxResults,
```
yusupsupriyadi, line 153: `matches.length >= parsed.maxResults` reports truncation whenever the count lands exactly on the limit, even when nothing was dropped: 50 matches with both sub-results `truncated: false` still answers `truncated: true`, and the agent burns a turn narrowing a query that was already complete. Same at 118.

The slice already knows whether it cut anything, so comparing the pre-slice length would be exact.

packages/core/src/tools/knowledge-base-search.ts
```diff
@@ -72,99 +78,175 @@ export async function runKnowledgeBaseSearch(
...
-    };
-  }
-
-  await ensureKnowledgeBaseDirs(orgId, profileId);
-
   const workspaceRoot = await resolveWorkspaceRoot(
     options.workspaceRoot ?? getProfileSoulDir(orgId, profileId)
   );
-  const searchTarget = await resolveSearchTarget(
+  const organizationTarget = await resolveOrganizationSearchTarget(
```
yusupsupriyadi, line 88: The organization target is resolved and its ripgrep pass awaited before the profile target is even resolved, so the two independent searches run strictly one after the other. When shared documents are attached that is two serial `rg` spawns per query on the agent's hot path.

Orgs with nothing attached do bail out before ripgrep, which is good, but they still pay two manifest reads plus a `flattenKnowledgeBaseLayout` pass before the profile search starts.

`Promise.all` over the two targets would make this cost the same as before the feature.

packages/core/src/knowledge-base/store.ts
```diff
@@ -358,19 +408,68 @@ export async function uploadKnowledgeBaseDocument(
...
+  if (!(await pathExists(profilesDir))) {
+    return [];
+  }
+  const entries = await readdir(profilesDir, { withFileTypes: true });
+  const profileIds: string[] = [];
+  for (const entry of entries) {
+    if (!entry.isDirectory()) {
+      continue;
+    }
+    const dir = await profileKnowledgeBaseDir(orgId, entry.name);
```
yusupsupriyadi, line 458: `profileKnowledgeBaseDir` is not a pure path helper: it runs `migrateLegacyKnowledgeBaseDir` (a `rename`) and `flattenKnowledgeBaseLayout`, which moves files and then `rm -rf`s `extracted/` and `uploads/`.

Calling it from a read-only "is this document in use?" predicate means one delete check walks every profile in the org and triggers the legacy migration on each one. For already-migrated profiles that is only a stat storm, but any profile still on the legacy layout gets mutated by a question that should not write anything, and a rename failing mid-loop aborts the delete with a half-migrated tree. The same guard also runs on every replace-upload.

`readManifestFrom(getKnowledgeBaseDir(orgId, entry.name))` is enough here.

packages/core/src/knowledge-base/store.ts
```diff
@@ -358,19 +408,68 @@ export async function uploadKnowledgeBaseDocument(
...
+    onDuplicate,
+    (documentId) => guardSharedDocumentRemoval(orgId, documentId)
+  );
+}
+
+export async function findProfilesReferencingSharedDocument(
+  orgId: string,
   documentId: string
+): Promise<string[]> {
+  const profilesDir = join(getOrgConfigDir(orgId), "profiles");
```
yusupsupriyadi, line 448: This enumerates profile directories on disk rather than the profile table, so a leftover directory blocks deletion permanently.

`deleteProfileWithHistoryArchives` removes the soul dir before `db.deleteProfile` and its own comment notes the filesystem cleanup can fail. When it does, the row is gone but the directory keeps its `sharedDocumentIds`, so the delete endpoint answers 409 with a `profileIds` entry that is not in `listProfiles`. The UI cannot detach a profile it cannot see, and there is no way out short of editing `~/.nakama` by hand.

Listing profiles from the database and intersecting with what is on disk would avoid the orphan case.

packages/core/src/knowledge-base/store.ts
```diff
@@ -406,6 +504,67 @@ export async function deleteKnowledgeBaseDocument(
...
+  );
+}
+
+export async function getProfileSharedDocumentIds(
+  orgId: string,
+  profileId: string
+): Promise<string[]> {
+  const dir = await profileKnowledgeBaseDir(orgId, profileId);
+  const manifest = await readManifestFrom(dir);
+  return [...new Set(manifest.sharedDocumentIds ?? [])];
```
yusupsupriyadi, line 532: `readManifestFrom` only validates that `documents` is an array and casts the rest, so a `sharedDocumentIds` that is not an array flows through as one. A manifest with `"sharedDocumentIds": "kb_abc123"` (hand-edited, or written by an older build) gives three different wrong behaviours:

- `manifest.sharedDocumentIds?.includes(documentId)` at 458 becomes a substring match, so deleting `kb_abc` is refused with a phantom 409
- this spread turns the string into single characters
- `ids.filter(...)` at 563 throws `ids.filter is not a function`, a 500 on detach

One `Array.isArray` plus a `typeof entry === "string"` normalization at parse time covers all three.

packages/client/src/client.ts
```diff
@@ -1292,6 +1292,59 @@ export class NakamaClient {
...
+    orgId: string
+  ): Promise<{ documents: ListKnowledgeBaseResponse["documents"] }> {
+    return this.request(`/v1/orgs/${encodeURIComponent(orgId)}/knowledge-base`);
+  }
+
+  async uploadOrganizationKnowledgeBaseDocument(
+    orgId: string,
+    document: DocumentAttachment,
+    onDuplicate?: KnowledgeBaseDuplicateAction
+  ): Promise<UploadKnowledgeBaseResponse> {
```
yusupsupriyadi, line 1305: `UploadKnowledgeBaseResponse` and `DeleteKnowledgeBaseResponse` (1321) both require a `profileId` that the org endpoints never send: the POST returns `{document, outcome}` and the DELETE returns `{deleted, documentId}`. A consumer reading `result.profileId` compiles and gets `undefined`.

`UploadOrganizationKnowledgeBaseResponse` and `DeleteOrganizationKnowledgeBaseResponse` were added in this PR at `contract.ts:2310-2318` for exactly this, and are currently referenced only inside `ProfileService`.

puspoaditya:
Blocker fixed — pushed `02c4590f`.

1. **The four organization methods now exist on `AgentService`** as thin delegations to `ProfileService`, next to the profile knowledge base methods they mirror. You were right about the route test: it hands the handlers the methods through its `agent` stub, so it passed while a real server answered 500. There is now an `agent-service` test that goes through the service, and on the old code it fails with `TypeError: service.uploadOrganizationKnowledgeBaseDocument is not a function` — so `bun test` covers the wiring instead of the stub. `tsc` agrees: the four `TS2339`/`TS2551` are gone (root error count 9285 → 9282). One pre-existing `Buffer` → `BodyInit` error at `profiles.ts:1030` becomes visible now that `bytes` is typed — same class as 835/960/1088 on main, not introduced here, left alone.
2. **Organization hits are resolved against the organization root.** New test, old code: `Expected: "kb_shared.extracted.txt"`, `Received: "../../knowledge-base/kb_shared.extracted.txt"`.
3. **Result budget.** The organization scope gets the remaining budget with one slot reserved, and the profile scope is trimmed to `maxResults - organizationMatches.length`, so the total stays at the cap. The test attaches one shared document alongside a profile KB that fills a 2-hit cap and asserts both scopes come back.

Verified: `knowledge-base-search` 8 pass, `agent-service` 31 pass, `profile-service` + `org-memory-tools` + organization knowledge base routes 58 pass, `memory-backend-service` 15 pass, `@nakama/core` + `@nakama/server` builds, `ultracite check` clean on the four files. `knip` does not run on my box (`oxc-parser` `RangeError: Array buffer allocation failed`, 4 GB RAM), so CI is the check there.

Two of your inline items need a call before I queue the rest:

1. **Who manages an organization knowledge base?** The routes are platform-admin only, but the PR description says org admins upload a shared document. If org admins own their org's documents, that is a guard change plus its own route tests, and I would rather do it deliberately than fold it into this fix. Which is it?
2. **`openAPIRegistry` for the six new routes** — here or a separate PR? It changes the generated spec, not behavior.

On the rest, in your order: dropping `resolveSourceDocumentIds`' dependency on the manifest, `truncated` at the exact limit, and then the store-path side effects (`profileKnowledgeBaseDir`'s migration/flatten running from the org path, disk enumeration in delete, `sharedDocumentIds` validation), then the `client.ts` response types and the dashboard upload control.

One honest note on the serial pass: I did not parallelise the two passes, because the organization budget now depends on how many profile hits came back. Running them concurrently means a fixed split instead of a remainder — say which you prefer and I will switch it.

ahmadrosid:
A couple of decisions:

- Keep shared organization knowledge-base management limited to platform admins.
- Add the six organization knowledge-base routes to the OpenAPI docs.
- Leave dashboard upload and delete for a follow-up PR.

puspoaditya:
All three decisions are in `335ae723`.

**Platform admins only** — the four org routes and the two attach/detach routes already called `requirePlatformAdminFromContext`. That is now pinned by a test that drives an org-admin session through all six and asserts `403`, so a later guard change cannot quietly widen the surface.

**OpenAPI** — the six routes are registered: `listOrganizationKnowledgeBase`, `uploadOrganizationKnowledgeBaseDocument`, `deleteOrganizationKnowledgeBaseDocument`, `getOrganizationKnowledgeBaseDocumentContent`, `attachSharedKnowledgeBaseDocument`, `detachSharedKnowledgeBaseDocument`. Delete documents the `409` body with `profileIds`. One extra line: the spec never declared the `Organizations` tag that the org memory, member, and curator routes were already using, so it is declared now.

**Dashboard upload/delete** — left out of this PR as you asked. `KnowledgeTab` keeps list plus attach/detach, so #983 stays on the backend contract.

Evidence:

- `bun test apps/server/src/http/routes/profiles-organization-knowledge-base.test.ts` → 6 pass (2 new)
- the new OpenAPI assertion fails without the registrations — verified by stashing `profiles.ts` and re-running
- `bun run check` clean, `@nakama/server` build exit 0, CI green on `335ae723`

Still open from my earlier list, in the order I would take them: `resolveSourceDocumentIds` resolving attached org ids from storage instead of the manifest, then the `store.ts` side effects (`profileKnowledgeBaseDir` running the migration and `rm -rf` when handed an org path).

PR #1050, opened 2026-09-16
fix(web): disable OpenCode Zen free models in Browse free models

### LINKED ISSUE #1049: "Browse free models" saves OpenCode Zen free models with the public key, which OpenCode now rejects
### Summary

"Browse free models…" in provider setup still offers OpenCode Zen free models and fills the API key with `public` for them, but OpenCode now rejects that key outside OpenCode. The provider saves fine and every chat turn with it then fails with `400 MissingSessionID`.

- `apps/web/src/hooks/use-provider-setup-form.ts:318` sets `setApiKey("public")` when `row.isZen && row.isFree && !row.deprecated`
- `apps/web/src/components/ModelsBrowseList.tsx:146` sorts those rows to the top, and `:329` shows the public-key badge on them

### Measured on

dcd21605

### Reproduction

1. Settings, add a provider, pick "Browse free models…", choose any OpenCode Zen row marked FREE (for example Big Pickle), save.
2. Send a message on a profile using it.

The same request without the UI:

```sh
curl -s https://opencode.ai/zen/v1/chat/completions \
  -H "Authorization: Bearer public" -H "Content-Type: application/json" \
  -d '{"model":"big-pickle","stream":true,"messages":[{"role":"user","content":"hi"}]}'
```

### Expected vs actual

Expected: the picker only offers a free model that works with the key it fills in.

Actual:

```
OpenCode Zen request failed (400 MissingSessionID): Error from provider (Console): OpenCode's free tier can only be used in OpenCode
```

None of the free `opencode` models on models.dev worked with the `public` key when I tested them (details in #1048).

Drop the `public` key special case: remove the `setApiKey("public")` branch and the public-key badge and sort rule, so Zen rows ask for a real key like any other `openai_compatible` provider. `isFree` can keep its FREE badge, since models.dev still reports those prices. A Zen user with a paid key is unaffected.

fajarhide:
## Summary

Browse free models no longer offers OpenCode Zen free models, since OpenCode only serves them inside OpenCode.

**Before:** picking Big Pickle filled the API key with `public`. The provider saved, then every turn failed with `400 MissingSessionID`.
**After:** Zen free rows are disabled with the tooltip "OpenCode's free tier only works inside OpenCode". The `public` key branch, PUBLIC badge and its sort rule are gone.

| before | after |
| --- | --- |
| before | after |

Before, after clicking the row: form

Why safe:
1. Clearing the key alone is not enough: Zen returns the same 400 for an empty or missing key, and the key field is optional for custom providers
2. Paid Zen models keep their rows unchanged (covered by the new test)

Only follow up if OpenCode reopens its free tier to other clients.

## Test plan
- [x] `bun test apps/web/src/hooks/use-models-dev.test.ts` passes, and fails with the Zen rule removed
- [x] `bun run test`: web 570 pass. Server has 57 failures, and the same 54 plugin failures happen on `upstream/main` here (plugin tests need npm network). The other 3 are 5s timeouts that pass when rerun alone
- [x] Setup wizard step 3 on a throwaway instance, before and after (screenshots above)

Closes #1049


ahmadrosid:
got it, no free lunch anymore 😄

PR #1051, opened 2026-09-16
fix(server): stop seeding an OpenCode Zen provider

### LINKED ISSUE #1048: First-boot seed writes an OpenCode Zen provider that OpenCode now rejects (400 MissingSessionID)
### Summary

The first-boot seed writes an OpenCode Zen provider with `apiKey: "public"` and makes it the default, but OpenCode no longer serves its free tier outside OpenCode. A seeded instance fails its first chat turn:

```
OpenCode Zen request failed (400 MissingSessionID): Error from provider (Console): OpenCode's free tier can only be used in OpenCode
    at streamChatCompletion (apps/server/src/providers/openai-compatible/index.ts:348:15)
```

`apps/server/src/seed.ts:111` (`writeOpenCodeZenProvider`). The second seeded model, `hy3-free`, no longer exists either. The Docker docs (`docs/website/content/docs/docker.mdx:46`) still promise this provider.

### Measured on

dcd21605

### Reproduction

The two models the seed writes, called the way `streamChatCompletion` calls them:

```sh
for m in big-pickle hy3-free; do
  curl -s https://opencode.ai/zen/v1/chat/completions \
    -H "Authorization: Bearer public" -H "Content-Type: application/json" \
    -d "{\"model\":\"$m\",\"stream\":true,\"messages\":[{\"role\":\"user\",\"content\":\"hi\"}]}" \
    -w ' HTTP %{http_code}\n'
done
```

Or boot a fresh volume with the three `NAKAMA_SEED_ADMIN_*` variables and send any message.

### Expected vs actual

Expected: a seeded instance can answer its first message, or at least lands on a provider the admin is told to configure.

Actual:

```
big-pickle: {"type":"error","error":{"type":"MissingSessionID","message":"Error from provider (Console): OpenCode's free tier can only be used in OpenCode"}} HTTP 400
hy3-free:   {"type":"error","error":{"type":"ModelError","message":"Model hy3-free is not supported"}} HTTP 401
```

I also tried every model models.dev lists as free under `opencode` (`nemotron-3-ultra-free`, `deepseek-v4-flash-free`, `minimax-m3-free` and five more). None of them answers with the `public` key: each returns either MissingSessionID or "not supported".

The `public` key route is gone on OpenCode's side, so no model id we could put there would fix it. The simplest fix is to stop writing a provider from the seed: create the admin and org only, update the Docker doc, and let the admin add a provider in Settings. Instances seeded before the fix keep a dead default provider. Worth deciding whether a startup migration should remove it or whether a release note is enough.

The same `public` key is also used by the web "Browse free models" picker, filed separately in #1049.

fajarhide:
## Summary

The first-boot seed now creates only the admin and org, because the OpenCode Zen provider it wrote is rejected by OpenCode.

**Before:** a seeded instance reported a working provider and failed its first turn with `400 MissingSessionID`.
**After:** it reports no provider, and the startup log points the admin to `/setup`, which opens at the provider step.

Same boot with the three `NAKAMA_SEED_ADMIN_*` variables on a fresh config dir:

```
before (dcd21605)  /health {"userConfigured":true,"providerConfigured":true}
                   config: provider "OpenCode Zen", key public, models big-pickle, hy3-free
after  (6cdb9685)  /health {"userConfigured":true,"providerConfigured":false}
                   log: "Provider not configured ... complete the setup wizard at http://127.0.0.1:4342/setup ..."
```

Why safe:
1. The no-provider state already has a path: `SetupWizard` starts at step 3 when the user exists
2. Only the seed changes; installs without the seed variables are untouched

Only follow up if already-seeded instances need the dead provider removed. This PR leaves their config alone.

## Test plan
- [x] `bun test apps/server/src/seed.test.ts`: 5 pass. The two changed cases fail on the old `seed.ts`
- [x] `bun run test`: every workspace green except server, whose 54 failures are the plugin tests that fail identically on `upstream/main` here (they need npm network)
- [x] Seeded boot on a throwaway config dir, before and after (output above)

Closes #1048


ahmadrosid:
Experiencing this issue few times. Thanks for the fix.

PR #1053, opened 2026-09-16
fix(web): keep session navigation on the selected chat

ahmadrosid:
## Summary

Switch chats once and stay on the selected conversation.

**Before:** opening chat B could restore chat A, reload both histories, and leave the wrong conversation selected.
**After:** navigation owns the selected chat; superseded loads cannot replace its state or clear its loading indicator.

Why safe:
1. Sending and resuming still update the URL explicitly.
2. Regression tests cover stale history, errors, status responses, and starting a new chat.

## Test plan
- [x] `cd apps/web && bun test src/pages/chat src/lib/chat-stream-resume.test.ts src/lib/chat-history.test.ts src/context/active-chat-profile-store.test.ts` — 54 pass.
- [x] Web build, TypeScript, touched-file Ultracite, and `bun run knip` pass. React Doctor: 91/100, one false-positive warning on cleanup guarded inside `finally`.
- [x] Browser: two session switches each issue one history request and stay on the selected URL.

---


apps/web/src/pages/chat/use-chat-page.ts
```diff
@@ -623,16 +652,21 @@ export function useChatPage() {
...
         }
 
         setError(formatError(err));
       } finally {
-        streamAbortRef.current = null;
-        setBusy(false);
-        setTurnStartedAt(null);
+        if (isCurrentLoad()) {
+          streamAbortRef.current = null;
+          setBusy(false);
```
ahmadrosid: > Move the reset into a `finally` block

The reset was already inside `finally`; the nested guard caused a false positive. Updated e8f0d404 to use a functional state update directly in `finally`, preserving the current busy state when an older request finishes. Added a regression case for a rejected current load. All 8 targeted tests and TypeScript pass; React Doctor now reports 100/100 with no issues.

PR #1059, opened 2026-09-17
fix(channels): preserve replacement active streams

### LINKED ISSUE #1058: Keep replacement channel streams registered during stale cleanup
## Problem

Replacing an in-flight channel reply can unregister the replacement stream.

`registerActiveStream(chatId)` aborts the previous controller and stores a new one. Each channel handler later calls `clearActiveStream(chatId)` in its `finally` block. If the old handler reaches cleanup after a new handler registers the same chat, the unconditional delete removes the new controller.

## Impact

- `/stop` cannot find and abort the replacement reply.
- Worker shutdown can treat the chat as idle while the replacement reply is still running.
- Telegram, Discord, and WhatsApp are affected because they share the core active-stream registry.

## Reproduction

1. Register stream A for a chat.
2. Register stream B for the same chat; A is aborted and B becomes active.
3. Run A's cleanup.
4. `stopActiveStream(chatId)` returns `false`, even though B is still running.

## Proposed fix

Make cleanup ownership-aware: pass the registering signal/controller to `clearActiveStream` and delete only when it is still the registry's current entry. Add a regression test for the A → B → cleanup A sequence.

rickvian:
## Summary

Replacement channel replies stay stoppable after an older reply finishes cleanup.

**Before:** an older Telegram, Discord, or WhatsApp stream could delete the active-stream registry entry for its replacement.
**After:** `clearActiveStream` removes an entry only when the caller owns its current signal.

Why safe:
1. Existing replacement behavior still aborts the prior stream.
2. The ownership check only prevents stale cleanup from deleting a newer stream.
3. The same core registry is exercised by all three channel handlers.

Only re-add unconditional cleanup if streams no longer replace each other for a chat.

## Test plan
- [x] `bun test packages/core/src/channel-active-stream.test.ts apps/platform/telegram/src/chat-handler.test.ts apps/platform/discord/src/chat-handler.test.ts apps/platform/whatsapp/src/chat-handler.test.ts` (167 pass)
- [x] `bun run check -- packages/core/src/channel-active-stream.ts packages/core/src/channel-active-stream.test.ts apps/platform/telegram/src/chat-handler.ts apps/platform/discord/src/chat-handler.ts apps/platform/whatsapp/src/chat-handler.ts`
- [ ] Send two overlapping messages in the same channel conversation, then use `/stop` while the replacement reply is running.

Raised by Codex; still requires human review.

Closes #1058


ahmadrosid:
LGTM

PR #1066, opened 2026-09-17
fix(automation): pass the schedule org id to the internal run

### LINKED ISSUE #1065: Scheduled automations never run: worker drops orgId and the internal run endpoint answers 404
## Summary

No scheduled automation has run since `4cf3dedc` (22 Aug). The automation worker's scheduler delegate drops the `orgId` argument, so every scheduled run reaches the server as `?orgId=undefined` and is answered `404 Automation not found`.

```ts
// apps/platform/automation/src/scheduler.ts:20-24 (@ 01a512ff)
this.scheduler = new AutomationScheduler({
  getDefaultTimezone: () => this.fetchDefaultTimezone(),
  listScheduledAutomations: () => this.fetchSchedules(),
  runAutomation: (id) => this.runAutomation(id),   // ← orgId dropped
});
```

The delegate contract has taken two arguments since #434, and the core scheduler passes both:

```ts
// packages/core/src/automation-scheduler.ts:11-14, 87, 121
runAutomation(automationId: string, orgId: string): Promise<{ ok: boolean; skipped?: boolean; error?: string }>;
...
void this.delegate.runAutomation(automation.id, automation.orgId)
```

Every cron schedule and every one-shot `runAt` is affected, for every org.

## Measured on

`01a512ff`

## Reproduction

Failing test, added to the existing `apps/platform/automation/src/scheduler.test.ts`:

```ts
test("passes the automation orgId to the internal run endpoint", async () => {
  const calls: Array<[string, string]> = [];
  const client = createMockClient({
    listAutomationSchedules: async () => [
      {
        id: "a1",
        orgId: "o1",
        profileId: "p1",
        runAt: new Date(Date.now() + 50).toISOString(),
        timezone: "UTC",
      },
    ],
    runAutomationInternal: async (id: string, orgId: string) => {
      calls.push([id, orgId]);
    },
  });

  const scheduler = new AutomationWorkerScheduler(client);
  await scheduler.start();
  await Bun.sleep(150);
  scheduler.stop();

  expect(calls).toEqual([["a1", "o1"]]);
});
```

`bun test apps/platform/automation/src/scheduler.test.ts`

## Expected vs actual

Expected: the worker calls `runAutomationInternal("a1", "o1")` and the server answers `204`.

Actual:

```
error: expect(received).toEqual(expected)

@@ -3,3 +3,3 @@
     "a1",
-    "o1",
+    undefined,
   ],

- Expected  - 1
+ Received  + 1

(fail) AutomationWorkerScheduler > passes the automation orgId to the internal run endpoint [151.00ms]
 3 pass
 1 fail
```

Against a running server, `undefined` is serialised into the query string and the route answers 404:

```
POST /v1/internal/automations/auto_xxx/run?orgId=<real org>   → 204
POST /v1/internal/automations/auto_xxx/run?orgId=undefined    → 404 {"error":"Automation not found"}
```

## Root cause or proposed fix

`packages/client/src/client.ts:1549-1559` interpolates the argument it is given, so `undefined` becomes the literal string `"undefined"`:

```ts
`/v1/internal/automations/${encodeURIComponent(automationId)}/run?orgId=${encodeURIComponent(orgId)}`
```

That string is truthy, so the `if (!orgId)` guard at `apps/server/src/http/routes/internal-automations.ts:75-78` does not return 400. Execution reaches `automationService.get(automationId, "undefined")` at `:80`, which finds nothing, and `:83` answers `404 Automation not found`. The route 404s before `agent.runAutomation`, so no run record is written either: the dashboard shows no failed run, just nothing.

Fix is the missing argument:

```ts
runAutomation: (id, orgId) => this.runAutomation(id, orgId),
```

Two things kept this quiet for 26 days:

1. `AutomationScheduler` discards the delegate's result. `automation-scheduler.ts:86-93` and `:120-128` only `.catch()` thrown errors; a returned `{ ok: false, error }` is dropped, so a refused run logs nothing. Worth logging the failed result alongside the fix.
2. The repo has no `tsc` step. `typescript` is a devDependency of `apps/web` only and there is no `typecheck` script, so a call missing a required argument in a worker package is never checked.

This is the caller side of the ownership assertion added in #434. Passing the automation's own `orgId` satisfies that check; it does not relax it.

yusupsupriyadi:
## Summary

Scheduled automations fire again: the worker now sends the schedule's `orgId` with every internal run.

**Before:** the delegate called `runAutomation(id)`, the request went out as `?orgId=undefined`, and every cron and `runAt` run was answered `404 Automation not found`.
**After:** `runAutomation(id, orgId)` in `AutomationWorkerScheduler`, and a refused run is logged instead of discarded.

Why safe:
1. `AutomationSchedulerDelegate` already declared `(automationId, orgId)` and the core scheduler already passed both; only the worker adapter ignored the second argument
2. It satisfies the ownership assertion from #438 rather than relaxing it: the id passed is the automation's own `orgId`, the same value the schedules endpoint returns
3. `tsc` reports `scheduler.ts(23): TS2554 Expected 2 arguments, but got 1` on the old line and nothing on the new one

Only revisit if a schedule can legitimately carry an empty `orgId` — the route answers 400 for that, not 404.

Fixes #1065. Related, not fixed: #438, the merged change that added the assertion this caller never satisfied.

## Test plan
- [x] `bun test apps/platform/automation packages/core/src/automation-scheduler.test.ts` (10 pass)
- [x] `bun test packages/core` (855 pass, 2 pre-existing omni environment failures)
- [x] `bun run knip` (clean) and `ultracite check` on the four changed files (clean)
- [x] On a running worker: a daily cron schedule fired at its exact boundary (`T+0.011s`) and wrote a run record. Before the fix the same schedules produced no run record at all, only the 404


ahmadrosid (commented):
LGTM

PR #1067, opened 2026-09-17
feat(auth): add permanent user erasure

isnaenihidayat:
## Summary

Platform admins can erase a user's login and personal access data while retaining anonymous organization history. Part of #368.

**Before:** Removing membership left the user record, sessions, channel mappings, reset tokens, and per-user integrations intact.
**After:** `DELETE /v1/platform/users/:userId` atomically scrubs PII and identity links, disables the account, and preserves chats and shared artifacts without their actor identity.

Why safe:
1. Erasure runs in one database transaction and leaves organization-owned content intact.
2. The route is platform-admin only and rejects erasing the caller's current account.

Only follow up after #973 lands to append a dedicated user-erasure audit event.

## Test plan
- [x] `bun test apps/server/src/http/platform-orgs.test.ts apps/server/src/services/org-service.test.ts` (59 pass)
- [x] `bun test packages/db/src`; DB/server builds; targeted Ultracite; `bun run knip` (116 pass, all checks clean)
- [x] Erase a test user belonging to two organizations; confirm the old login fails and retained chats show no user owner.


ahmadrosid (commented):
LGTM from a simplification perspective. The two-organization scenario has not been tested yet—the PR's test plan still marks it unchecked. Please confirm once a user belonging to two organizations has been erased and both organizations' retained chats are verified anonymous.

isnaenihidayat:
Confirmed in 8072556: the route test now creates the user membership and a retained chat in two organizations, erases the account once, then verifies both sessions remain with userId set to null and both message histories retained. `bun test apps/server/src/http/platform-orgs.test.ts` passes: 13 tests, 0 failures.

PR #1092, opened 2026-09-18
fix(chat): proxy external response images

### LINKED ISSUE #1070: bug(chat): external images in assistant responses are blocked by CSP
> *This was generated by AI during triage.*

### Summary
External image URLs in assistant responses cannot load under Nakama's current Content Security Policy. The reported response shows "Image not available" for school images. A controlled browser check confirmed that a valid external PNG is blocked by `img-src`, while a same-origin image loads.

This is separate from #1068 / #1069: those concern uploaded images disappearing when history reloads, not external images embedded in assistant Markdown.

### Measured on
Browser verification on a live self-hosted deployment on 2026-09-17. Its response header contains `img-src 'self' data: blob:`. The same restriction is present in `apps/server/src/http/app.ts` on [latest main checked, 252d096](https://github.com/ahmadrosid/nakama/commit/252d096d613df66868184140ee1587b2d09c6558). The deployed commit was not checked.

### Reproduction
1. Open the deployed Nakama web app, which serves the CSP header above.
2. In browser DevTools, run the following read-only image-loading probe (no chat mutation or API key needed):

```js
document.addEventListener('securitypolicyviolation', e =>
  console.log(e.effectiveDirective, e.blockedURI));
await Promise.all([
  '/nakama-logo-dither.png',
  'https://www.google.com/images/branding/googlelogo/2x/googlelogo_color_272x92dp.png'
].map(src => new Promise(resolve => {
  const image = new Image();
  image.onload = () => resolve({src, result: 'loaded', width: image.naturalWidth});
  image.onerror = () => resolve({src, result: 'error', width: image.naturalWidth});
  image.src = src;
})));
```

3. Open the external PNG directly in the same browser as a control.

### Expected vs actual
Expected: a valid external image can be displayed through a supported, privacy-conscious response-image flow, or the UI clearly explains that external loading is blocked.

Actual measured results:
- Same-origin Nakama logo: `loaded`, natural width `1254`.
- External PNG inside Nakama: `error`, natural width `0`; `securitypolicyviolation.effectiveDirective = img-src`, with the external PNG as `blockedURI`.
- The identical external PNG opened directly: decoded successfully at `544 x 184`; HTTP response was `200`, `Content-Type: image/png`.

The exact school-image URLs from the reported conversation were not available for verification. They may also have independent URL or hotlinking problems; the controlled probe establishes the CSP restriction, not their validity.

The server's `img-src` directive excludes external HTTPS origins. Chat responses use Streamdown with no custom image renderer. Decide the external-image policy before implementation: allowing remote loads exposes the reader's IP and may enable tracking; consent-gated loading or a carefully secured proxy have different tradeoffs. Do not disable CSP wholesale or broaden script permissions. A proxy would also require SSRF safeguards.

### Security
- [x] This touches authentication, authorization, tenant isolation, credentials, or untrusted code execution.

This report concerns browser security policy and untrusted model-supplied image URLs; no data exposure was demonstrated. Suggested labels: `bug`, `role: FE`, `role: BE`. Maintainers should choose priority and stage after settling the privacy policy.

mamenesia:
## Summary

Chat users can see supported external images in assistant responses without loosening CSP. Fixes #1070; assignment requested and pending.

**Before:** external Markdown images are blocked by `img-src 'self' data: blob:`.
**After:** an authenticated, org-scoped proxy returns supported HTTPS images; uploaded images and ordinary links retain their existing behavior.

Why safe:
1. Reuses DNS pinning and redirect validation, without forwarding browser credentials or upstream cookies. The five confirmed missing non-public ranges are now blocked before requests and on redirects.
2. Requires HTTPS/443 and matching PNG/JPEG/GIF/WebP signatures, with a 5 MiB cap, 10-second deadline, private/no-store responses, and unchanged CSP.

Residual: the inherited IPv6-literal TLS hostname compatibility limitation is unchanged; real IPv6-literal TLS compatibility was not verified. This is not a claim of a complete security audit or a demonstrated exploit.

## Test plan
- [x] Focused fetch/session tests: 58 pass; message-rendering tests: 2 pass. The new reserved-address regression fails before hardening and passes afterward. Full lint and Knip passed locally after correcting checkout line endings.
- [x] Browser proof: public PNG renders through the proxy with no direct external browser request; local image loads directly; loopback URL is rejected. Web/server builds and split-UTF-8/text-size smoke checks passed before the range-only follow-up. Previous full Windows run: 3,704 pass, 57 fail, 6 skip, one error; all 57 failures also reproduce on the unchanged parent. See CI for the current head.
- [ ] Reviewer: render an external HTTPS PNG and a local attachment in chat; confirm both display and ordinary links retain their confirmation flow.

Security evidence: IANA lists the five previously omitted ranges as not globally reachable ([IPv4 registry](https://www.iana.org/assignments/iana-ipv4-special-registry/iana-ipv4-special-registry.xhtml), [IPv6 registry](https://www.iana.org/assignments/iana-ipv6-special-registry/iana-ipv6-special-registry.xhtml)). Mocked-network tests establish a destination-validation gap, not successful access to internal services or data exfiltration. The correction is included here rather than deferred to a separate issue.

Review: Ponytail self-review found no worthwhile cuts. Fresh Oracle review found the destination-gap blocker resolved and no remaining blockers in the correction, subject to passing validation. Implementation and review were AI-assisted.


ahmadrosid (commented):
LGTM

PR #1104, opened 2026-09-18
feat(telegram): add managed bot QR onboarding

adexaja:
## Summary

Allow users to create and connect a Telegram managed bot from Settings by scanning a QR code.

**Before:** Users had to create a Telegram bot manually and paste its token into Nakama.
**After:** Users scan a QR code, create the managed bot in Telegram, and Nakama automatically receives and saves the managed bot token.

Managed bot flow:
1. User scans the QR code and completes bot creation in Telegram.
2. Telegram sends a `managed_bot` update to the Nakama manager bot.
3. Nakama matches the update to the pending pairing.
4. Nakama calls `getManagedBotToken` with the created bot ID.
5. User identity and bot token are saved to the selected profile.
6. Telegram worker can use the managed bot without manual token entry.

Why safe:
1. Pairings are scoped to the authenticated org and user.
2. Pairings expire after 10 minutes and only waiting pairings are polled.
3. The token is stored through the existing Telegram configuration path; manual token setup remains available.

Only re-add / follow up if Telegram Manager Bot API changes its managed-bot update or token contract.

## Telegram manager bot setup

Required environment variable:

```env
NAKAMA_TELEGRAM_MANAGER_BOT_TOKEN=<manager-bot-token>
 ```
 The server uses this token to call Telegram’s Bot API. Do not expose it in the web app, commit it, or send it to users.

 Create the manager bot:

 1. Open @BotFather (https://t.me/BotFather) in Telegram.
 2. Run /newbot.
 3. Choose a display name, for example Nakama Bot Manager.
 4. Choose a username ending in bot, for example nakama_manager_bot.
 5. Copy the token returned by BotFather.
 6. In the manager bot settings in BotFather, enable the permission to manage/create bots (bot_can_manage_bots).
 7. Configure the token before starting Nakama:

 ```bash
export NAKAMA_TELEGRAM_MANAGER_BOT_TOKEN="123456789:AA..."
bun run dev:server
 ```

 For Docker:

 ```bash
docker run -d \
  -p 4310:4310 \
  -e NAKAMA_TELEGRAM_MANAGER_BOT_TOKEN="123456789:AA..." \
  -v nakama-data:/nakama/data \
  --name nakama \
  ghcr.io/ahmadrosid/nakama:latest
 ```

 The manager bot token is only used server-side. Nakama calls getMe, listens for managed_bot updates, then calls getManagedBotToken to retrieve the token of the bot created by the user.

 References:

 - Telegram managed bots (https://core.telegram.org/api/bots/managed-bots)
 - Telegram Bot API (https://core.telegram.org/bots/api)

## Test plan
- [x] `bun install`
- [x] `bun run check`
- [x] `bun run knip`
- [x] `bun test apps/server/src/services/telegram-managed-bot-pairing.test.ts` (2 pass)
- [x] `bun run --filter @nakama/server build`
- [x] `bun run --filter @nakama/web build`
- [x] `bun run test`
- [x] Manual test. Videos attached

https://github.com/user-attachments/assets/57fb6c9f-a76c-4fe9-a5cd-37149bb45b90

### CODE REVIEW

**Review by ahmadrosid — COMMENTED** [review]
Two small simplifications: reuse the shared response types and keep the active pairing in one state value.

File: apps/server/src/services/telegram-managed-bot-pairing.ts [hunk 1]
```diff
@@ -0,0 +1,261 @@
...
+  orgId: string;
+  ownerUserId?: number;
+  profileId: string;
+  state: PairingState;
+  suggestedUsername: string;
+  userId: string;
+  token?: string;
+};
+
+export interface TelegramPairingStartResponse {
```
ahmadrosid: These two response interfaces already exist in packages/core/src/contract.ts in this PR. Import them from @nakama/core instead of redeclaring them here, so the service and client share one definition.

apps/web/src/components/telegram-qr-setup.tsx
```diff
@@ -0,0 +1,195 @@
...
+import { formatError } from "@/lib/client";
+
+export function TelegramQrSetup({
+  profileId,
+  running,
+}: {
+  profileId: string;
+  running: boolean;
+}) {
+  const [pairingId, setPairingId] = useState<string | null>(null);
```
ahmadrosid: The start result already includes pairingId, and every update sets or clears both state values together. Store it as TelegramPairingStartResponse | null and derive pairingId with pairing?.pairingId ?? null. That removes the extra state, paired setter calls, and handwritten response shape without changing the flow.

apps/server/src/services/telegram-managed-bot-pairing.ts
adexaja: Got it, I missed reviewing this code. Thanks for review.

apps/web/src/components/telegram-qr-setup.tsx
adexaja: Pushed new changes @ahmadrosid. thanks for the review

adexaja:
> Two small simplifications: reuse the shared response types and keep the active pairing in one state value.

Pushed the changes @ahmadrosid 👌

PR #1107, opened 2026-09-18
fix(kb): address search truncation, orphan directories, and client types (#1041 follow-up)

puspoaditya:
## Summary

Fixes search truncation accuracy, orphan directory blocks, and types in the organization knowledge base, addressing remaining review items from PR #1041.

**Before:** Search falsely reported truncation when hits exactly met the budget, and read-only org lookups failed if `findProfilesReferencingSharedDocument` hit an orphan profile directory.
**After:** Truncation is reported only when a match is actually dropped (probe `maxResults + 1`); profile guards read manifests in-place without triggering legacy migration; backend and workspace lookups run concurrently.

Why safe:
1. `maxResults` slicing in `mergeScopedMatches` is kept intact while allowing true detection of dropped items.
2. Read-only profile lookups fallback to legacy paths directly instead of mutating the directory structure (`readProfileManifestForReference`).
3. Corrupted `sharedDocumentIds` string entries fallback gracefully via `normalizeSharedDocumentIds`.

Only follow up if profile knowledge bases are failing to migrate due to skipped read-path migrations.

## Test plan
- [x] `bun test packages/core` (867 pass)
- [x] `bun x ultracite check` (1422 files checked, 0 errors)
- [ ] Upload an organization KB document, delete a profile, then successfully delete the org document without 409.


packages/core/src/knowledge-base/store.ts
```diff
@@ -40,6 +40,14 @@ interface KnowledgeBaseManifest {
   sharedDocumentIds?: string[];
 }
 
+export interface OrganizationKnowledgeBaseOptions {
```
ahmadrosid: This exported options type only carries one value, and every caller immediately wraps or unwraps it. Pass `knownProfileIds?: readonly string[]` directly to the two organization functions and remove the interface plus the object literals.

packages/core/src/knowledge-base/store.ts
```diff
@@ -204,6 +215,18 @@ function sanitizeFilename(filename: string): string {
   return base.replace(/[^\w.\-() ]+/g, "_") || "document";
 }
 
+function normalizeSharedDocumentIds(value: unknown): string[] | undefined {
```
ahmadrosid: This helper has one caller and adds more ceremony than the check itself. Inline the `Array.isArray(...)? .filter(...) : undefined` expression in `readManifestFrom`.

packages/core/src/knowledge-base/store.test.ts
```diff
@@ -379,4 +388,116 @@ describe("knowledge base store", () => {
...
+    );
+    const manifest = JSON.parse(await readFile(manifestPath, "utf8")) as Record<
+      string,
+      unknown
+    >;
+    manifest.sharedDocumentIds = uploaded.document.id;
+    await writeFile(manifestPath, JSON.stringify(manifest, null, 2), "utf8");
+
+    expect(await getProfileSharedDocumentIds(ORG_ID, profileId)).toEqual([]);
+    expect(
```
ahmadrosid: `readManifestFrom` does the normalization for all three paths, so the getter assertion already proves the malformed value is ignored. The detach and delete assertions repeat that same property; drop them and keep this test focused.

ahmadrosid (commented):
The core fixes look sound. These three spots can be smaller without losing behavior or coverage.

PR #1139, opened 2026-09-21
fix(web): read the chat profile list from the shared query

### LINKED ISSUE #1138: web: chat page refetches /v1/profiles on every chat switch and twice on first load
## Summary

`apps/web/src/pages/chat/use-chat-page.ts:441` calls `client.listProfiles()` directly inside `loadProfiles`, bypassing the React Query cache that the sidebar already fills through `useProfilesQuery()`. So the first `/chat` render sends `GET /v1/profiles` twice.

The bigger cost is on navigation. `loadProfiles` is memoised on `enterDraftChat`, which is memoised on `location.pathname`. Every chat switch creates a new `loadProfiles`, `useEffect(() => void loadProfiles(), [loadProfiles])` runs again, and the profile list is fetched again although nothing about it changed.

Small perf issue, not a crash. Filing it because the repro is cheap and the fix is local.

## Measured on

`af246736`, production build (`bun run build` in `apps/web`, served with `vite preview` against the `bun run agent:ui` API), so no StrictMode double effects.

## Reproduction

Browser:

1. Log in, open `/chat`, create two chats.
2. In the DevTools console, mark the current request count, click a chat under Recents, then print what arrived since the mark:

```js
window.__m = performance.getEntriesByType("resource").filter((e) => e.name.includes("/v1/")).length;
// click a chat, then:
const es = performance.getEntriesByType("resource").filter((e) => e.name.includes("/v1/"));
console.log(es.slice(window.__m).map((e) => new URL(e.name).pathname));
window.__m = es.length;
```

Hook in isolation (bun test, `happy-dom`, only `useChatPage` mounted inside `MemoryRouter` + `AuthProvider` + `AppProvider`, `client.listProfiles` spied):

```
mount /chat/p1/s1             listProfiles calls: 3
navigate to /chat/p1/s2       listProfiles calls: 5
navigate to /chat/p1/s1       listProfiles calls: 7
navigate to /chat/p1/s2       listProfiles calls: 9
```

The only `listProfiles` call site in the hook is line 441, and neither provider calls it, so all of these come from `loadProfiles`.

## Expected vs actual

| Action | Expected `GET /v1/profiles` | Actual (browser) | Actual (hook test) |
|---|---|---|---|
| First `/chat` load | 1 | 2 | 3 |
| Switch to chat A | 0 | 1 (of 3 requests) | +2 |
| Switch to chat B | 0 | 1 (of 9 requests) | +2 |

The browser sees one per click because React Query dedupes the in-flight request from the sidebar; the isolated hook shows both.

## Root cause or proposed fix

`loadProfiles` still has real work to do: `syncForOrg` and the `enterDraftChat` fallback when the route profile is not in the org. Only the network call is redundant. Reading the list through the cache keeps that logic and drops the extra requests:

```ts
const { profiles } = await queryClient.ensureQueryData(profilesQueryOptions);
```

That is free within `staleTime` (30s), deduped with the sidebar, and no longer tied to the URL. Happy to open the PR with the hook test above as the regression test if this looks right.

Also seen, separate fix: `/v1/settings/thinking` loads twice on `/chat` because `createSettingsHooks` (`apps/web/src/hooks/use-app-queries.ts:49`) sets no `staleTime`, so the sidebar prefetch is already stale when `useThinkingSettings()` mounts.

AlexMuhammad:
## Summary

Switching chats no longer refetches the profile list. The chat page reads it from the same React Query cache the sidebar fills.

**Before:** `loadProfiles` in `use-chat-page.ts` called `client.listProfiles()` directly and was memoised on `location.pathname`, so `/chat` fetched `/v1/profiles` twice on load and once more on every chat switch.
**After:** `useProfilesQuery()` replaces the local state and raw fetch; `syncForOrg` and the `enterDraftChat` fallback run in an effect keyed on the query data instead of the URL.

Why safe:
1. Same `syncForOrg` / `enterDraftChat` / `setError` behaviour, only the data source changed
2. The hook now follows cache invalidations (profile created or deleted elsewhere), where before it only refreshed by accident on navigation
3. Regression test fails on `main` (3 calls on mount) and passes here (1 call, none on navigation)

Only revisit if a page relies on the chat hook re-reading profiles on route change after a mutation that forgets to invalidate `queryKeys.profiles.all`.

Fixes #1138

## Preview

Production build against `bun run agent:ui`. The box in the top right is a `PerformanceObserver` injected from the console for the recording; it counts `GET /v1/profiles` and chat switches. Not part of the diff.

**Before (`main`):** every click on a chat in Recents sends another `GET /v1/profiles`. 4 switches, 4 fetches.

before

**After (this PR):** 4 switches, 0 fetches.

after

`/v1/*` requests per action:

| Action | Before | After |
|---|---|---|
| First `/chat` load, `GET /v1/profiles` | 2 | 1 |
| Switch to another chat | 3 requests, one is `/v1/profiles` | 2 requests (`messages`, `status`) |
| Hook mounted alone, 3 navigations, `listProfiles` calls | 9 | 1 |

## Test plan
- [x] `bun test src/pages/chat/use-chat-page.test.tsx` in `apps/web` (10 pass; new test fails on `main` with `Expected: 1, Received: 3`)
- [x] `bun test` in `apps/web` (607 pass)
- [x] `bun run check`, `bunx tsc --noEmit` in `apps/web`, `bun run knip` (no new findings)
- [x] Browser: production build against `bun run agent:ui`, request counts in the table above
- [ ] Open `/chat/<other-org-profile>/<session>`: still falls back to a draft chat on this org's profile


ahmadrosid (commented):
LGTM — I don’t see any unnecessary complexity to cut. The production change removes duplicated fetch/state, and the regression test setup is proportionate to exercising the hook through its providers.

PR #1140, opened 2026-09-21
feat(config): support mounted Composio API keys

isnaenihidayat:
## Summary

Mount the installation-wide Composio API key from a secret file, advancing #364 without writing it to Nakama's data directory.

**Before:** Composio could only read its API key from the saved plaintext setting.
**After:** `COMPOSIO_API_KEY_FILE` and `COMPOSIO_API_KEY` override the saved setting and appear as configured in system status and Integrations.

Why safe:
1. Direct environment values win, mounted files come next, and the existing saved setting remains the fallback
2. Resolution reuses the provider and channel secret-file helper, including trimming and fail-fast reads

Only follow up if startup should actively probe the mounted key; this change keeps the existing lazy reachability check.

## Test plan
- [x] `bun test packages/core/src/composio.test.ts apps/server/src/services/composio-service.test.ts apps/server/src/services/system-status-service.test.ts apps/server/src/http/routes/composio.test.ts` (33 pass)
- [x] `bun run --cwd packages/core build`, `bun run --cwd apps/server build`, and `bun run --cwd docs/website build`
- [ ] Start Nakama with `COMPOSIO_API_KEY_FILE` and confirm Integrations shows Composio as connected


fajarhide:
@isnaenihidayat This matches what `discord-config.ts:368` and `telegram-config.ts:330` already do, down to the shape of the line, so Composio stops being the one integration that cannot take its key from the environment. The precedence in the docs matches the code.

All five checks are green because this ran before `3b1b3837` landed on main and broke Unit Tests there (#1144).

ahmadrosid:
I’ve confirmed the manual testing for this PR. Ready to merge.

PR #1141, opened 2026-09-21
fix(web): keep prefetched settings fresh for the default stale time

AlexMuhammad:
## Summary

Opening `/chat` fetches `/v1/settings/thinking` once instead of twice.

**Before:** `createSettingsHooks` built its query options without a `staleTime`, so the sidebar's `prefetchAppData` result was already stale when `useThinkingSettings()` mounted on the chat page and React Query fetched it again.
**After:** the factory sets `staleTime: defaultStaleTime` (30 s), the same value every other prefetched query in `use-app-queries.ts` already uses, so the mount reuses the prefetch.

Why safe:
1. Every save path in the factory writes the response back with `setQueryData`, so the cache is never stale after a save regardless of `staleTime`
2. Matches the 12 other `queryOptions` in the same file; `thinkingSettings` was the only prefetched query without it
3. Regression test fails on `main` (`Received number of calls: 2`) and passes here

Only revisit if a settings page must reflect a change made by another admin in under 30 s without a reload.

Follow-up from #1138, where this second duplicate was noted.

## Test plan
- [x] `bun test src/hooks/use-app-queries.test.tsx` in `apps/web` (1 pass; 1 fail on `main`)
- [x] `bun test` in `apps/web` (611 pass)
- [x] `bun run check`, `bunx tsc --noEmit` in `apps/web`, `bun run knip` (no new findings)
- [ ] Production build, open `/chat`, DevTools Network filtered to `settings/thinking`: one request


yusupsupriyadi (approved):
@AlexMuhammad LGTM.

Checked out the branch and ran the regression test both ways: it fails on the base (`Received number of calls: 2`) and passes with the `staleTime` line, so it really does pin the behaviour and not just the current output. `bun test src/hooks src/pages/chat/use-chat-page.test.tsx` is green (25 pass).

Also traced the other ten settings groups the factory builds, since the line lands in the shared factory and only `thinkingSettings` is prefetched. None of them need a refetch-on-mount inside the 30 s window:

- telegram, discord, whatsapp: every pairing or handshake change either writes back with `setQueryData` or calls `invalidateQueries`, and invalidation refetches active observers regardless of `staleTime`
- whatsapp live connection state comes from `useSystemStatusQuery`, a separate query this does not touch
- `webPublicUrl` invalidates in `onSaveSuccess`, the rest save through `setQueryData`

One small correction for the description: `webPublicUrlSettings.onSaveSuccess` and `thinkingSettings.onSaveError` invalidate rather than `setQueryData`, so "every save path writes the response back with `setQueryData`" is not quite literal. The conclusion still holds, invalidation is the stronger guarantee.

fajarhide:
@AlexMuhammad Ran this one rather than read it. Your test applied to main without the `staleTime` line fails with `Received number of calls: 2`; with the line it passes, and the full web suite is 601 pass, 0 fail.

`createSettingsHooks` was the only query factory in that file without a `staleTime`, while the ten other queries already set `defaultStaleTime`. So this puts the ten settings hooks back in line with the file's own convention instead of adding a new rule.

The red `test` check is not from your change. main broke at `3b1b3837` over a hardcoded plugin version, filed as #1144.

AlexMuhammad:
@yusupsupriyadi thanks for tracing the other settings groups and for the correction. Updated point 1 of the description: save paths either write back with `setQueryData` or call `invalidateQueries` (`webPublicUrlSettings.onSaveSuccess`, `thinkingSettings.onSaveError`), and invalidation refetches active observers regardless of `staleTime`.

@fajarhide thanks for running it both ways and confirming the red `test` check is the `3b1b3837` breakage tracked in #1144, not this change.

PR #1142, opened 2026-09-21
fix(audit): record permanent erasure events

isnaenihidayat:
## Summary

Permanent organization deletion and user erasure now leave append-only security records, completing the audit-coverage slice of #368.

**Before:** Both destructive operations completed without an audit event.
**After:** The audit middleware records `organization.delete` and `user.erase` with actor, target, outcome, and the correct organization scope.

Why safe:
1. The deletion and erasure services are unchanged; only their existing routes gain audit descriptors
2. Route-backed tests execute both operations and assert the persisted ledger entries

Only follow up if the audit action taxonomy adopts different names for permanent deletion or erasure.

## Test plan
- [x] `bun test apps/server/src/http/audit-log.test.ts apps/server/src/http/platform-orgs.test.ts` (41 pass)
- [x] `bun run --cwd apps/server build`
- [ ] Permanently delete one archived organization and erase one test user, then confirm both events appear in the audit ledger


fajarhide:
@isnaenihidayat The switch to `Object.hasOwn` is the part worth checking, since it changes how every route resolves `orgId`. It holds up: `event()` never puts an `orgId` key on the descriptor, so `hasOwn` is false everywhere except the five routes that set it explicitly, and each of those passes `decodePathPart(match[1])` where `([^/]+)` guarantees a non-empty string. No existing route changes behaviour, and `user.erase` gets the deliberate null it needs. Neither new matcher shadows an existing one.

The red `test` check is not from your change. main broke at `3b1b3837` over a hardcoded plugin version, filed as #1144.

Issue #1134, opened 2026-09-21
fix(web): opening a second chat cancels the turn running in the first one


yusupsupriyadi:
## Summary

The web dashboard runs one chat at a time. Opening a second chat cancels the turn still
running in the first one, so two chats can never make progress together.

Every path that leaves a chat aborts the one stream handle the chat page keeps, and that
abort closes the POST streaming the turn, which the server treats as a cancel.

`apps/web/src/pages/chat/use-chat-page.ts:560`
`apps/server/src/http/shared.ts:671`

## Measured on

v0.4.19 (`5e820f82`). The same code path is on `af246736`.

## Reproduction

1. Open a chat and send a prompt that runs for a while (a long tool run, or a long answer).
2. While it is still streaming, click another chat in the sidebar, or click New chat.
3. Reopen the first chat.

The server half is already pinned by a test on main:
`apps/server/src/http/routes/sessions.stream.test.ts:133`, "cancelling the response stream
also ends the turn".

## Expected vs actual

Expected: the first chat keeps running in the background, and its answer is there when the
chat is reopened.

Actual: the turn ends the moment the page switches away. `streamMessage` reports the abort
as a cancel and writes the terminal event:

```
data: {"type":"error","error":"Turn cancelled."}
```

`endTurn` then drops the session from `sessionTurnRegistry`, so
`GET /v1/sessions/:sessionId/status` answers `{"active":false}` and reopening the chat finds
nothing running.

## Root cause or proposed fix

Two things meet.

1. `useChatPage` keeps a single `streamAbortRef`, and every path that leaves a chat aborts it:
   - `use-chat-page.ts:560` `resumeSession`, which is what a sidebar click runs
   - `use-chat-page.ts:411` `enterDraftChat`
   - `use-chat-page.ts:787` the `?new=1` handler behind New chat
   - `use-chat-page.ts:234` the unmount cleanup

   Only `stopStreaming` (`:861`) should be aborting.

2. The server ends a turn as soon as the request streaming it goes away: `streamMessage`
   takes `c.req.raw.signal` (`apps/server/src/http/routes/sessions.ts:742`) and also aborts
   on the response stream's `cancel()` (`apps/server/src/http/shared.ts:671`).

So the abort is not a UI detach. It cancels the agent run.

Proposed fix, web side only, leaving server semantics alone: on a chat switch, detach the
stream instead of aborting it. Drop the UI handlers, leave the request open so the turn keeps
[...]

PR #1143, opened 2026-09-21
feat(web): live session list with generated titles and a running mark

### LINKED ISSUE #1135: feat(web): live session list with generated titles and a running indicator
## Summary

The sidebar session list (`apps/web/src/components/AppSidebar.tsx:145`) only moves when
something invalidates its query. Two things should appear in it without a page refresh:

1. the generated chat title, once `SessionTitleService` has written it
2. a mark on a session whose turn is still running

Both are served by one mechanism: a gated refetch on `useHistorySessionsQuery`, plus one new
field on `SessionSummary`.

## Why it matters

A new chat keeps its fallback label until the page is reloaded, because the title is written
after the turn returns and nothing refetches at that moment. And once #1134 is fixed, a chat
keeps running after the user switches away from it, but the list gives no sign of that, so a
background turn is invisible until its chat is reopened. The running mark is what makes that
fix legible, and the title is the same refetch paying for itself twice.

Title, why it is late today:

- `apps/server/src/services/session-title-service.ts:17` generates the title fire and forget,
  after the turn has already returned
- `apps/web/src/pages/chat/use-chat-page.ts:1052` invalidates the sessions query in the
  turn's `finally`, which runs before the title exists
- so the title lands in the database with nothing left watching for it

Running state, what already exists:

- `apps/server/src/services/session-turn-registry.ts:199` `isActive(sessionId)` already
  answers the question
- `GET /v1/sessions/:sessionId/status` (`apps/server/src/http/routes/sessions.ts:637`)
  exposes it one session at a time, which is the wrong shape for a list
- `SessionSummary` (`packages/core/src/contract.ts:1044`) does not carry it, so the list
  route (`apps/server/src/http/routes/sessions.ts:527`) has nothing to send

Shape of the change:

- add `active: boolean` to `SessionSummary` and fill it from the turn registry in the list route
- give `useHistorySessionsQuery` (`apps/web/src/hooks/use-resource-mutations.ts:510`) a
  `refetchInterval`, gated so it polls only while some listed session is `active` or still has
  `title === null`, and stops otherwise
- render the mark in `AppSidebar.tsx`

Polling here is the existing pattern, not a new one:

- `apps/web/src/components/chat/WorkflowRunToolRow.tsx:95`, `refetchInterval: isRunning ? 800 : false`
- `usePostTurnPolling` in `apps/web/src/hooks/use-post-turn-skill-review-overlay.tsx:71`, a
  bounded window after a turn

Two things to get right:

- `useHistorySessionsQuery` fans out over the history channels with `useQueries`, so one
  interval is several requests per tick. The gate is not optional.
- `sessionTurnRegistry` is in-process memory, so `active` is only true for the process serving
  the request. Fine for a single server process, worth naming before anyone scales it out.

## Acceptance criteria

- [ ] `SessionSummary` carries `active`, and `GET /v1/sessions` reports it from the turn registry
- [ ] a chat started from the sidebar shows its generated title without a page reload
- [ ] a session with a running turn is marked in the list, and the mark clears when that turn ends
- [ ] the sessions query stops polling once no session is active and every listed session has a title
- [ ] switching to another chat while the first is running shows the first one as running in the list

## Related

Related, not fixed here: #1134. That one stops a chat switch from cancelling the turn behind
it. This one makes the surviving turn visible in the list. Sequenced after it, because the
running mark has nothing to show until #1134 lands.

yusupsupriyadi:
## Summary

A chat names itself in the sidebar as soon as its title is generated, and the chat still answering is the one whose title is sweeping colour.

**Before:** `useHistorySessionsQuery` had no interval, so a new chat kept its fallback label until a reload and a running turn was invisible.
**After:** `SessionSummary` carries `active` from the turn registry, and the query refetches on `sessionListPollInterval`, which closes the moment nothing is outstanding.

Screenshot_32

Why safe:
1. The gate is a pure function with its falsifying inputs pinned, including the untitled session whose title never arrives and the unparsable timestamp, both of which must answer "stop polling"
2. Only the web channel follows this tab's own turns, and it is the only channel the chat page can send to (`READ_ONLY_SESSION_CHANNEL` in `apps/web/src/lib/chat-history.ts`); the other three still poll on what their own rows say, so this is not the single-channel gap #910 reported
3. `rollPalette` still sets `--ai-*` inline per rim, which wins over the new `:root` declaration, so the composer renders exactly as before

Only revisit if `active` has to be true across processes: the turn registry is in-process memory, so it answers for the process serving the request.

Fixes #1135

Related, not fixed here: #1134, merged as #1136, which is what lets a turn survive a chat switch and so gives this mark something to show. No other open PR of mine; `git merge-tree` against `origin/main` is clean.

## Test plan
- [x] `bun test` in `apps/web` (618 pass)
- [x] `bun test` on the new server test, `app.test.ts`, both cognito suites and the Discord chat handler (68 + 63 pass)
- [x] `bun x tsc --noEmit` in `apps/web`, `ultracite check` on all 11 changed files, `bun run knip` (all clean)
- [x] CI `test` is red only on `PluginService > installs the bundled Google Meet plugin and executes its isolated action`, which this branch touches no code for; `knip`, `coverage`, `duplication` and `sync` are green
- [x] Send a message in a chat the sidebar already lists: its title sweeps colour while the turn runs, then settles and renames itself without a reload


ahmadrosid:
this is cool

